PR #14, opened 2021-04-07
Add a Pyodide kernel

jtpio:
Add a basic Python kernel based on Pyodide using the web worker approach: https://pyodide.org/en/latest/usage/webworker.html

Some things are still a bit clunky and might require refactoring some of the kernel messaging a bit more (could also be done separately). We might also want to remove the iframe wrapper and communicate with the web worker directly.

Handy code snippet for testing:

```python
import pandas as pd
import numpy as np
from string import ascii_uppercase as letters

df = pd.DataFrame(np.random.randint(0, 100, size=(100, len(letters))), columns=list(letters))
df
```

https://user-images.githubusercontent.com/591645/113924197-cef0e500-97e9-11eb-91a3-acf1cafea83b.mp4

### Changes

- Basic Pyodide based Python kernel
- Basic support for kernel restarts
- Add the toc extension
- Start JupyterLab with the left area collapsed


jtpio:
Staying with Pyodide `0.15.0` for now (and python 3.7) as there seems to be some warnings with `0.16.0+` (with python 3.8)

image


Tracked in https://github.com/pyodide/pyodide/issues/877

jtpio:
Consoles also work and can be used side by side with a notebook:

image

jtpio:
Linking to the latest preview before making subsequent changes: https://jupyterlite-om2y06azw-jtpx.vercel.app/

jtpio:
I think this is working quite ok for now to merge it and iterate

PR #33, opened 2021-04-11
Add a JupyterLab Classic frontend

jtpio:
The idea will be to have both `JupyterLab` and `JupyterLab Classic` as frontends for `JupyterLite`.

For now we can keep them in the same repo here, but they could of course be split.

This will make it easier to deploy them together on Vercel to the main URL for demo purposes.


jtpio:
Some progress on making it work with JupyterLab Classic:

https://user-images.githubusercontent.com/591645/114442114-7bf1a600-9bcc-11eb-9ccf-9303e3851aa9.mp4

Looks good overall. Now let's see how both JupyterLab and JupyterLab Classic could be deployed on Vercel.

jtpio:
Looking smooth on the Vercel preview:

https://user-images.githubusercontent.com/591645/114452321-9d588f00-9bd8-11eb-8dd0-6cff1a96702d.mp4

jtpio:
For now we redirect `/tree` to `/classic/tree` so the "Launch Classic Notebook" menu entry opens JupyterLab Classic.

We can reconsider this later after it is moved to its own plugin: https://github.com/jupyterlab/jupyterlab/pull/10086

PR #38, opened 2021-04-13
Move web worker to a separate file

jtpio:
First steps towards fixing the `Failed to execute 'postMessage' on 'DedicatedWorkerGlobalScope': [object Object] could not be cloned` errors (for example when evaluating lists).


jtpio:
This should also fix evaluating basic results:

image

PR #39, opened 2021-04-13
Provide an alternative plugin to download files

jtpio:
To be able to download the files, instead of hitting the `/files` server endpoint.

This depends on the upstream 3.1 lab packages (alpha 5 or greater, to be published).

### DIFF
--- a/app/lab/index.js
+++ b/app/lab/index.js
@@ -76,7 +76,8 @@ window.addEventListener('load', async () => {
   const disabled = [
     '@jupyterlab/apputils-extension:themes',
     '@jupyterlab/apputils-extension:workspaces',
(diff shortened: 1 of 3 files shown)


bollwyvl:
As noted over on https://github.com/jtpio/jupyterlite/issues/40#issuecomment-821981458, i think maintaining files in some form would actually be quite helpful for users. A custom plugin will still be required, for sure, for localStorage-only files, but things like images, embedded, vidoes etc. really don't need to go into localstorage... until we have a video editor!

bollwyvl:
I assume this can be closed, given #89?

jtpio:
Yes, closing in favor of #89.

Issue #41, opened 2021-04-14
Integrate into existing Jupyter-based workflows


bollwyvl:
Awesome stuff here! Maybe these have been brought up elsewhere, if so, feel free to close!

## elevator pitch

Having an installable, python-based package with hooks for various other jupyter tools with a couple features would make it easy to adopt for workflows folk are already invested in.

## Use Cases

- as a Jupyter user, locally be able to interactively build up a custom jupyterlite site with contents, custom extensions, etc. and do whatever you need to with the files
- as an extension developer community, build a demo site in CI and deploy to somewhere (e.g. github pages)

### nbconvert

One of the jyve features I loved building is the [nbconvert exporter](https://github.com/deathbeds/jyve/blob/master/jyve/exporter.py). That code's pretty rough, but it _does the thing_ of marshalling all the static assets into a folder, ready to be committed to `gh-pages` (or dragged into SharePoint :scream:) for demos.

I'm imagining two, not necessarily mutually exclusive outputs of:
- a full editable lab
- a single Document that _looks_ like nbconvert, etc. but is built out of lab renderers
- a couple Documents, one per page

With those pieces, one could integrate nicely into other workflows like sphinx, jupyter-book, mkdocs, etc.

#### design ideas

- a python package 
  - an `entry_point` for `jupyterlite site`
  - this would make it show up in various CLI and API endpoints
  - all/some of the `jupyterlite` static assets as either...
    - `data_files` for "simplicity"
    - `package_data` it's gonna be _big_, duplication in wheels is no fun

As `nbconvert` likes to make _one_ file, the default output would probably most usefully be a `.zip` file... not so much fun for previewing, but quite handy for moving around (as again, it's gonna be big). But if it gets more complicated...

### labextension

There are bells and whistles that would need more tweaking than what the exposed nbconvert API offers. It would be lovely to have an in-juptyerlab(-classic) extension that built jupyterlite sites and then did... stuff with them.

#### design ideas

This part would _have_ to be in `data_files`, but wouldn't be very big.

- offer it as a new _File > export as... > custom jupyterlite_
  - open a new mainareawidget (or whatever makes sense in classic)
    - capture the advanced options as a JSON schema
      - make a JSON schema from ui
    - a drop area for adding additional contents paths

As to targets, I'm imagining other outputs/targets, such as:

- git bundle
- start hosting the site
  - just `/files`
  - serve on a federated content network, e.g. dat/ipfs
[...]

Issue #22, opened 2021-04-09
Disable URL routing


jtpio:
To make it easier to serve with a simple web server, we should check whether the URL routing (or at least the location rewrite) can be disabled.

![image](https://user-images.githubusercontent.com/591645/114220171-f1ab0700-996b-11eb-833f-79c1642689d8.png)

This will make it more convenient to refresh the page without a 404.

Alternatively we might want to define placeholders to handle that more gracefully.

jtpio:
This should be feasible by disabling the `@jupyterlab/application-extension:main` plugin and providing an alternative (simpler) on instead. And upstream commands could still be reused if they are moved to a separate plugin (https://github.com/jupyterlab/jupyterlab/pull/10073)

PR #48, opened 2021-04-16
Update developer experience

bollwyvl:
Hooray jupyterlite! I'm excited to start digging into #41 (and in turn getting others excited), but there's a fair amount of "read the source" right now, which is fine! While digging into it, I found a fair number of things that could be quickly updated to improve the developer "curb appeal", and pave the way for more contribution. 

Of course, if any of these things _aren't_ fun, we can take them out, split them up, whatever!

- [x] package.json
  - [x] rework yarn tasks
    - [x] `watch`
    - [x] split `build` into `build:lib` and `build:app`
    - [x] make `prettier` quieter
    - [x] increase the printWidth to 88 
    - [x] prettier more: markdowns, yamls, etc. 
    - [x] add `serve` (appears broken for me) and `serve:py` (used in proxy)
- [x] yarnrc
  - [x] add `.yarn-packages` for local mirroring
  - [x] set some defaults (timeouts, etc.) 
- [x] adds `metapackage`
- [x] docs
  - [x] flesh out CONTRIBUTING
  - [x] README
    - [x] remove dev stuff 
    - [x] add badges
      - [x] binder
      - [x] ci
- [x] adds a binder [this pr](https://mybinder.org/v2/gh/bollwyvl/jupyterlite/update-dx?urlpath=lab)
  - [x] conda env with lab, classic, etc. 
    - for offering _something_ actionable in CONTRIBUTING
      - and eventual stuff on #41
  - [x] jupyter-server-proxy, with a (strawman) lightbulb logo
    - [x] ~~always tries to redirect to `doc/workspaces/auto-X?reset`~~
      - [x] disabled `apputils-extension:resolver`, seems to do the thing, not sure what side-effects this may have, but workspaces deserve special consideration, anyway...
    - [x] ~~pyodide kernels die~~
      - [x] i'm giving up here, as this happens locally, too... but only 
        - if the browser console _isn't_ open, in classic heisenberg fashion
        - in chromium (FF LTS is fine, so Works For Me!)  
- [x] app
  - [x] replace all static bundle references and pageInfo with dynamic ones that sniff the `window.location`
    - [x] ~~the kernels all seem to fail, and thrash against the (non-existent) endpoint~~
      - [x] calculate the websocket url from the baseUrl
    - [x] favicons, too
- [x] .github
  - [x] add issue/pr templates (just copied the core ones, mostly)
  - [x] add `node_modules` and `.yarn-packages` caching to workflow


bollwyvl:
Ah looks like disabling IWindowResolver busts main, so there are a number
of things that will break that I just haven't found yet. Probably need to
overload... Which will have to do anyway once we can ship (a default)
workspace.

jtpio:
Thanks, lots of great stuff here!

> Ah looks like disabling IWindowResolver busts main

Yes, but with https://github.com/jupyterlab/jupyterlab/pull/10073 now in lab we should hopefully be able to write a minimal `main` plugin and disable the upstream one, after updating to the next 3.1 alpha packages.

> in chromium (FF LTS is fine, so Works For Me!)

I also noticed issues with Pyodide in Chromium making the tab crash. This might be fixed by https://github.com/pyodide/pyodide/pull/1474

app/tree/index.html
```diff
@@ -1,6 +1,14 @@
...
 <html>
-<head>
-  <meta http-equiv="refresh" content="0; url=/classic/tree" />
-</head>
+  <head>
+    <script>
+      // redirect to /lab by default
+      (function() {
+        window.location.href = window.location.href.replace(
+          /\/(index.html)?$/,
```
jtpio, line 8: We might also want to catch the case when coming from the "Launch Classic Notebook" menu entry in lab:

image

Otherwise it looks like it's going into an infinite redirect loop when hitting `/tree`.

docs/_static/icon.svg
```diff
@@ -0,0 +1,187 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
jtpio, line 1: Really nice icon!!

bollwyvl:
yeah, i'm working on the urls... there's a bunch of icky cases

packages/_metapackage/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@jupyterlite/metapackage",
+  "private": true,
+  "version": "0.1.7",
```
jtpio, line 4: Should we stick to `0.1.0` for now until we start publishing packages?

docs/_static/icon.svg
bollwyvl: :blush:

packages/_metapackage/package.json
bollwyvl: sure, sorry, went off `application`. it's private, so it doesn't matter... but think it has to have _something_ for... reasons.

bollwyvl: bde8c6d

app/tree/index.html
bollwyvl: bde8c6d

bollwyvl:
Yeah, just overloading IWindowResolver wasn't sufficient... i'd be inclined to keep what's in now (maybe add a comment/link to #22) and call it a win...

jtpio:
Maybe `@jupyterlab/application-extension:main` could also be disabled then. But I guess it's so also ok to have it fail hard so at least it's visible.

Because of this we lose some of the application commands such as "Close All Other Tabs". But it's fine and should be back when updating to the next 3.1 alpha packages.

.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
+<!--
+Thanks for contributing to JupyterLab!
+Please fill out the following items to submit a pull request.
+See the contributing guidelines for more information:
+https://github.com/jupyterlab/jupyterlab/blob/master/CONTRIBUTING.md
+-->
```
jtpio, line 6:
```suggestion
<!--
Thanks for contributing to JupyterLite!
Please fill out the following items to submit a pull request.
See the contributing guidelines for more information:
https://github.com/jtpio/jupyterlite/blob/main/CONTRIBUTING.md
-->
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+
+## User-facing changes
+
+<!-- Describe any visual or user interaction changes and how they address the issue. -->
+
+<!-- For visual changes, include before and after screenshots here. -->
+
+## Backwards-incompatible changes
+
+<!-- Describe any backwards-incompatible changes to JupyterLab public APIs. -->
```
jtpio, line 26:
```suggestion
<!-- Describe any backwards-incompatible changes to JupyterLite public APIs. -->
```

app/classic/edit/index.html
```diff
@@ -1,25 +1,43 @@
...
+          2
+        );
+        const bundle = document.createElement('script');
+        bundle.src = base + 'classic/build/bundle.js';
+        bundle.main = 'index';
+        document.body.appendChild(bundle);
+        const favicon = document.createElement('link');
+        favicon.rel = 'icon';
+        favicon.type = 'image/x-icon';
+        favicon.href = base + 'lab/favicon.ico';
```
jtpio, line 38: It would be nice to keep the classic favicon here (for files and notebooks):

image

app/classic/index.html
```diff
@@ -1,6 +1,14 @@
 <!DOCTYPE html>
 <html>
-<head>
-  <meta http-equiv="refresh" content="0; url=/classic/tree" />
-</head>
+  <head>
+    <script>
+      // redirect to /lab by default
```
jtpio, line 5:
```suggestion
```

packages/_metapackage/package.json
jtpio: :+1:

app/tree/index.html
jtpio: Thanks!

app/classic/edit/index.html
jtpio, line 38:
```suggestion
        favicon.href = base + 'classic/edit/favicon.ico';
```

app/classic/notebooks/index.html
```diff
@@ -1,25 +1,43 @@
...
+          2
+        );
+        const bundle = document.createElement('script');
+        bundle.src = base + 'classic/build/bundle.js';
+        bundle.main = 'index';
+        document.body.appendChild(bundle);
+        const favicon = document.createElement('link');
+        favicon.rel = 'icon';
+        favicon.type = 'image/x-icon';
+        favicon.href = base + 'lab/favicon.ico';
```
jtpio, line 38:
```suggestion
        favicon.href = base + 'classic/notebooks/favicon.ico';
```

app/classic/edit/index.html
bollwyvl: will revert, but maybe we need another treatment... i think it's important the app indicates that _this ain't your regular jupyter_... maybe a book/document with a lightswitch on it?

bollwyvl: c6cacef

app/classic/notebooks/index.html
bollwyvl: c6cacef

.github/workflows/build.yml
```diff
@@ -2,29 +2,51 @@ name: Build
...
+        id: cache-yarn-packages
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
+        with:
+          path: '.yarn-packages'
+          key: |
+            ${{ env.CACHE_EPOCH }}-yarn-packages-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            ${{ env.CACHE_EPOCH }}-yarn-packages-
+      - name: Install
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
```
bollwyvl, line 40: had forgotten this one to get the actual benefit of caching... knocks a minute off the runtime.

app/classic/edit/index.html
jtpio: Sounds good :+1:

jtpio:
Thanks a lot @bollwyvl, this is great!

PR #57, opened 2021-04-20
First steps towards an improved Pyodide kernel

jtpio:
## References

Improve the state of the Pyodide based kernel added in https://github.com/jtpio/jupyterlite/pull/14.

Move some of the logic to a `pyolite` Python package, which can then be loaded in Pyodide with:

```python
import micropip
await micropip.install('/url/to/the/wheel')
``` 

- [x] Rename to `pyolite` (Pyodide kernel for JupyterLite)
- [x] Copy the `pyolite` wheel to the `build` directory
- [x] Use the Pyodide Interactive console for code execution: https://pyodide.org/en/0.17.0a2/usage/python-api/pyodide.console.InteractiveConsole.html
- [x] Add a simple display module to display basic things like json, markdown and html

## Code changes

Things to improve later on:

- dependency on Python and its build system to produce the wheel? -> probably not too big of an issue when authoring a Python kernel
- remove the global webpack `asset/resource` rule after adding support for federated extensions: https://github.com/jtpio/jupyterlite/issues/55. The wheel should be a static asset just like the rest

## User-facing changes

displays


bollwyvl:
Super terrifying, yet super awesome.

Between this and federated extensions, I could see this evolving into some very effective stuff. It'll get a bit messy w/r/t emscripten when that comes about, but we're getting closer to have a single `mamba install` for the whole toolchain... and ways to run them on the server, which, nutty as it sounds, is actually quite attractive from a real compile-once-run-all-over perspective.

As these sizes increase, we may well have to think about service workers...

bollwyvl:
Also, if I may suggest: as we get to significant inter-language dependency (and eventual tests #6), I find having top-level tooling (with or without management of the outer environment) is really clutch. 

My go-to tool for this is `doit`, which is python-based and has effectively no dependencies, and is really good at kowning _what thing should I do next?_... sure, it's md5-based, but my build chain isn't trying to attack me. Even more crucially during a tight development loop, it knows _what things can I do at the same time?_ This is _very_ helpful when switching between PRs, as the repo becomes basically self-healing.

For my investigations on #41 (which will need a couple things ramped up before being in a pushable state), I have setup doit tooling in an outer repo that looks like this:

https://gist.github.com/bollwyvl/2278cddfb0af2caf77611588b58895a8

There is of course a certain art in identifying the _right_ entropy to track, as some files have lots of noisy timestamps, etc. but there's enough prior art out there at this point.

One thing I've played with is embedding some of the knowledge on the `package.json`, with a `doit` key with the same structure as `scripts`... and on the python side, I've been trying to get `doit` to speak `pyproject.toml`... but this has been a long slog.

jtpio:
Thanks Nick. Maybe we should indeed switch to something like `doit` at some point.

One thing that might start getting a bit annoying will be building and deploying to Vercel. Although they have basic support for installing Python dependencies it's a bit clunky and for now doesn't work for this PR.

jtpio:
> One thing that might start getting a bit annoying will be building and deploying to Vercel. Although they have basic support for installing Python dependencies it's a bit clunky and for now and it doesn't work for this PR (yet).

Maybe there is a way to build on GitHub Actions (so we could setup everything needed easily), and then upload the build directory to Vercel. Although not sure the integration would be the same as it is now (it sounds too manual). I really like the preview feature just for how easy it is to quickly try a change from a PR. Worst case, there is always Binder to keep some kind of preview.

Another option could be to deploy to GitHub Pages instead, using something like https://github.com/marketplace/actions/deploy-to-github-pages.

jtpio:
Adding some basic support for displaying html, svgs and latex:

image

jtpio:
Patching `matplotlib.pyplot.show` for now to be able to render plots as pngs:

image

Later on we might be better off using [`ipympl`](https://github.com/matplotlib/ipympl) after adding support for `ipywidgets`.

jtpio:
Another option which is becoming more tempting and might be the simplest way to get Vercel back, would be check the wheel in the repo to avoid triggering a Python build.

Not the best, but might be enough for now until we figure out a better workflow that would keep the preview on PRs.

jtpio:
Thinking about keeping it like this for now with the wheel checked in, so it works nicely on Vercel.

This should bring a couple of improvements to the Pyodide based kernel, which of course can be iterated on.

Then we can see how to fix the build system via https://github.com/jtpio/jupyterlite/issues/63

jtpio:
Until we have a way to easily add example notebooks (https://github.com/jtpio/jupyterlite/issues/30), there is this gist if folks want to try it out: https://gist.github.com/jtpio/3ba7feacc67431488e33a148429f96f1

Issue #55, opened 2021-04-18
Support federated extensions


bollwyvl:
### Problem

The webpack build(s) already take(s) a long time. As a site publisher adds additional extensions, this will take longer until it just doesn't work anymore: for example, while published on npm, ipydrawio doesn't reliably work as a source extension because... 5000 files. Further, and for the reason above, unless they _really_ need it, I don't think a lot of extension authors will even be _testing_ source installs.

### Proposed Solution

Add `federated_extensions` to `pageInfo` and the machinery to load them. 

Indeed, the jupyterlite kernels, etc. would likely be useful _outside_ of a jupyterlite install, e.g. when someone is building content, so they gain access to things like jupyter-lsp, etc. but that's probably for another day, and maybe counterfeited by a preview mode on #41, if the complexity is too high (because patching ServiceManager).

### Additional context

For pure js builds, provide some mechanism to get sdists from pypi.io by URL (or locally), unpack them, copy the files into `lab/labextensions/`.

For python-driven builds (#41), be able to copy things from `$PREFIX/share/jupyter/labextensions` to `lab/labextensions` (https://github.com/jtpio/jupyterlite/issues/16#issuecomment-821980005).

bollwyvl:
Had a look at this, appears to be a pretty deep well. I got up to where i was loading `federated_extensions`, but then fell over once they tried to resolve the "known-first-party" stuff. Definitely seems doable, though.

jtpio:
Thanks!

> Had a look at this, appears to be a pretty deep well. I got up to where i was loading `federated_extensions`, but then fell over once they tried to resolve the "known-first-party" stuff. Definitely seems doable, though.

Do you want to open a draft PR even if it's still a work in progress?

In JupyterLab Classic support for federated extensions is implemented in a very similar way as in the lab examples:

- https://github.com/jtpio/jupyterlab-classic/blob/main/app/index.js and https://github.com/jtpio/jupyterlab-classic/blob/main/app/webpack.config.js
- https://github.com/jupyterlab/jupyterlab/tree/master/examples/federated

> Add federated_extensions to pageInfo and the machinery to load them.

Don't we also need a way to create or maintain a list of the installed federated extensions? Since there is no server, not sure we can do something like in `jupyterlab_server`: https://github.com/jupyterlab/jupyterlab_server/blob/8bcdf52a06bd81e649825e153a3f93947064fe23/jupyterlab_server/config.py#L39

jtpio:
[...]

PR #58, opened 2021-04-20
federated extensions, webpack sharing, deploying/configuring docs

bollwyvl:
## References

- for #55 

## Code changes

- [x] Adds build from upstreams
- [x] better handle setting federated_extensions in the `index.html` files
- [x] try with a simple extension that has schemas, for example: https://github.com/jtpio/jupyterlab-cell-flash
  - for now schemas are all bundled into a `all.json` file on build:
- [x] be able to just copy the prebuilt extension assets to `lab/extensions`
  - not have to edit all.json when new federated extensions are installed?
- [x] settings service will have to be aware of federated extensions
  - [x] dynamically construct the list of all settings in this getAll() method.
- [ ] a `@jupyterlite/top` (or `site` or something) that carries and abstracts some of the boilerplate
  - [ ] should be able to dedupe the app webpack bundles

### Deployer
- Someone who wants to deploy a site would be able to
  - at the very leasst on this PR
    - drop some well-formed packages in `<app>/extensions` (as copied from `$share/labextensions`)
    - run a `.py` and/or `.js` script to update various `all.json` files

### End User

- n/a (though load times might be.... different, due to multiple packages)

?


jtpio:
Thanks for getting this started!

jtpio:
Trying this out locally:

- with https://github.com/jtpio/jupyterlab-python-file as the test extension
- manually copying the prebuilt assets to `/lab/extensions`

image

Also changed the structure and boilerplate code so it follows the upstream example a bit more closely: https://github.com/jupyterlab/jupyterlab/blob/master/examples/federated/core_package/webpack.config.js

jtpio:
Next steps:

- better handle setting `federated_extensions` in the `index.html` files
- try with a simple extension that has schemas, for example: https://github.com/jtpio/jupyterlab-cell-flash
- for now schemas are all bundled into a `all.json` file on build:

https://github.com/jtpio/jupyterlite/blob/534111c7c3b9f1dbe04c7b99c9eb7bff187e63a1/app/webpack.config.js#L37-L59

so the lab settings manager can fetch all settings at once:

https://github.com/jtpio/jupyterlite/blob/534111c7c3b9f1dbe04c7b99c9eb7bff187e63a1/packages/settings/src/settings.ts#L37

It would be nice to be able to just copy the prebuilt extension assets to `lab/extensions` and not have to edit `all.json` when new federated extensions are installed. Basically a copy should be enough. But then the jupyterlite settings service will have to be aware of federated extensions, and be able to dynamically construct the list of all settings in this `getAll()` method.

bollwyvl:
Thank you so much for looking into this. I'll check it out soon... the delta from my broken stuff should be enough for me to keep moving forward.

> settings service will have to be aware of federated extensions

Yeah, sounds like table stakes for getting this in.

Humans should not have to write anything other than the one JSON file that seeds pageinfo, etc. so this is just another part of that, i guess...

bollwyvl:
Nothing to concretely comment on, but the duplication between classic/lab definitely points to us perhaps wanting a `@jupyterlite/top` or something that carries and abstracts some of the boilerplate... will have to do some thinking.

jtpio:
> but the duplication between classic/lab definitely points to us perhaps wanting a `@jupyterlite/top` or something that carries and abstracts some of the boilerplate

Yeah that sounds good.

> the one JSON file that seeds pageinfo

And for now we could limit the scope to Python driven installs, which sounds the simplest so we could copy straight from `$PREFIX/share/jupyter/labextensions` to `lab/labextensions`.

bollwyvl:
Further, I think we have to plan for being able to deploy servlite
extensions _as_ federated extensions, and downselect them based on some
metadata at setup time... We can't maintain (and make people go use)
_another_ build chain, and then figure out where to upload it, etc.

It should be possible to work against a local one, however, without
packaging, ideally live deployed into the app with "labextension watch"...
We can monkey patch is we must, but I think all the switches are available
some way or another.

However, we don't need to solve that on this pr! Well already be able to do
lots of good demos with this, modulo the things we've called out (which
I'll hoist to the checklist)

bollwyvl:
I am making progress here, but not necessarily in the direction i was expecting:

- [x] move all `jupyter-config-data` into a `jupyter-lite.json`
  - [ ] add a schema for `jupyter-lite.json` (not using it at runtime, yet, because `ajv` is :elephant:)
  - [x] for now, this contains just `jupyter-config-data`, but I got some ideas...
    - [ ] I'm thinking we can abuse OpenAPI to normalize the `all.json`s
- [x] in each `index.html`, do a hard redirect, if needed, to `{pathname}/` so that relative URLs work predictably
- [x] make a `/config-utils.js` (could use a more lyrical name, perhaps) with no imports allowed which
  - [x] reads and mangles `jupyter-config-data` with `jupyter-lite.json` files up the chain to the `jupyter-lite-root`
  - [x] adds favicons (based on config)
  - [x] adds the bundle script tag
    - [x] specified with a `<script rel="preload" id="jupyter-lite-main" href="./build/bundle.js" as="script"/>`) 
      - _partially_ offsets the additional requests

This whole yak :barber:  is to demonstrate being able to have a single `_static/.jupyter-lite` deployed in the docs site which adds some `federated_extensions`... which works locally for some stuff:

Screenshot from 2021-05-02 17-33-18

So I reckon we add `tour` and a few other goodies to `docs`, and see how it goes...

bollwyvl:
So looking into it more: my go-to `.tar.gz` format is not fit for this purpose, as one would have to actually parse `setup.py` or something to get the location of the data files, whereas they are in an entirely-predictable place within the `.whl`.

On the python side:
- as I'd like to avoid a `jupyterlab` build dependency as long as possible:
  - `pip download --no-deps --only-binary :all: jupyterlab-tour` will grab the files to a predictable location without having to know too much information
  - then it's either `wheel unpack` (or `ZipFile`) to get them out

In a JS-only toolchain:
- one would _need_ to know _at least_ the actual version number... and potentially the architecture, etc. in order to get the correct `.whl`.
- then `zlib` can take care of the unpacking

While these don't sound huge, it is suggesting to me that landing the end user tool chains might need to go outside of this PR, but may be served by using a common config format, which may or may not be seeded by `jupyter-lite.json` (as it has to _update_ the file). We could put something lightweight into the current docs, however, without checking stuff in... it's really just a question of what are our top-flight selections. Of course, I'm partial to ipydrawio, but it is pretty heavy...

jtpio:
Thanks, this is looking great!

> However, we don't need to solve that on this pr! Well already be able to do
lots of good demos with this, modulo the things we've called out (which
I'll hoist to the checklist)

Yeah absolutely, need to give some more thoughts to this. Also in the case where third-party packages want to distribute both a `serverlite` and a "normal jupyterlab extension" together.

> So I reckon we add tour and a few other goodies to docs, and see how it goes...

> We could put something lightweight into the current docs, however, without checking stuff in... it's really just a question of what are our top-flight selections. Of course, I'm partial to ipydrawio, but it is pretty heavy...

Sounds good :+1: I think we could also have a list of curated extensions that work well, or just some that we like to use. 
Generally speaking, it would be nice to always have a functional demo on `main` that doesn't break too often. So as long as it doesn't disrupt the rest, adding new federated extensions to the default demo website on RTD should be fine.

app/config-utils.js
```diff
@@ -0,0 +1,211 @@
...
+  const bundle = document.createElement('script');
+  bundle.src = preloader.href;
+  bundle.main = preloader.main;
+  document.head.appendChild(bundle);
+}
+
+/**
+ * TODO: consider better pattern for invocation.
+ */
+await main();
```
jtpio, line 211: Looks like this might give a top level await error on some browsers.

docs/deploying.md
```diff
@@ -0,0 +1,115 @@
+# Deploying
+
+Deploying your JupyterLite requires:
+
+- an actual HTTP server (doesn't presently woth with `file://` URLs)
```
jtpio, line 5:
```suggestion
- an actual HTTP server (doesn't presently work with `file://` URLs)
```

app/config-utils.js
bollwyvl: yeah, i just wanted something that _didn't_ webpack for this stage. as mentioned, we likely want this for "in-app" stuff too, so it probably needs to become a package with an entry point in the top-level webpack.

We also need to do more work on cache-busting, getting that `[hash]` back into file names, as I've noted on RTD it's pretty easy to get in a bad cache state. there's probably some more stuff we can fiddle with to get more-reproducible chunks.

.eslintrc.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
       { avoidEscape: true, allowTemplateLiterals: false }
     ],
     curly: ['error', 'all'],
-    eqeqeq: 'error',
+    eqeqeq: ['error', 'always', { null: 'ignore' }],
```
jtpio, line 48: Do we still need the `null: 'ignore'` part? Or was is for testing?

It would be nice to stick to `===` checks in general, but maybe this was added for a specific reason?

bollwyvl: strict equality of `null` and `undefined` is not consistent, whereas if you let the browsers Do Their Casting Thing, it comes out correctly....

packages/settings/src/settings.ts
```diff
@@ -59,13 +50,77 @@ export class Settings {
...
+  const _federated = new Map<string, IPlugin | undefined>();
+
+  export const _storage = localforage.createInstance({
     name: STORAGE_NAME,
     description: 'Offline Storage for Settings',
     storeName: 'settings',
     version: 1
   });
+
+  export async function getAll(): Promise<{ settings: IPlugin[] }> {
```
jtpio, line 68: Looks like the settings are not updated on the RTD preview for this PR, for example when switching the theme or moving tabs to the left and right area: https://jupyterlite--58.org.readthedocs.build/en/58/_static/lab/index.html

Although refreshing the page does pick them up. So maybe it's related to the new `getAll` "merging" logic.

bollwyvl:
> list of curated extensions

That's a whacking good idea. quick sketch:
- as we don't yet know what we want it to look like "for real", add a dumb file of:
  - map of `npm-package-name` to 
    - full `.whl` URLs
    - whether we want them in the demo site
    - notes about their deployment
    - maybe a link to any open issues about supporting lite
- during the docs build, use that file to
  - download and unpack
  - write `jupyter-lite.json`
  - build up the docs page
    - read `whl` metadata to get homepage etc
    - look at the on-disk size to see how much heft we're talking

As for that dumb file:
- I don't see this becoming part of the end user tooling, _per se_
  - probably best to stick to stdlib dependencies when possible, which means `.json` or `.ini` :crying_cat_face: 
- but for docs purposes, maybe TOML for pr-ability?
  - we already have `flit` around, which depends on `toml`
    - even though we won't be able to ship _all_ the packages with it, because `data_files` :scream_cat:  

> always have a functional demo on `main`

No complaints there! Indeed, #6 should be testing _that_ (probably deployed locally with e.g. `tornado`), not just the empty-and-light-as-possible contents of `app` which might not be minified, etc.

> distribute both a serverlite and a "normal jupyterlab extension" together.

Well... on Jyve, i paid the :monkey: price to make stuff work in both full lab and the offline thing, and it was a huge hassle... but i'd _definitely_ want to be able to use things like `jupyterlab/pull-requests` with Notebooks for kernelites.

What we _can't_ do is introduce a new package build chain: it's gotta build with `labextension build`, and end up as another plugin.

I'd see hoisting `serverlite-extensions` up to `jupyter-lite.json` and making that something that the tooling can either detect at build time by introspecting a `jupyterlite` key in `package.json`, or that a deployer can manually specify.

packages/settings/src/settings.ts
bollwyvl: yeah, gross. needs test #6 lol

jtpio:
> we already have flit around, which depends on toml 

Using `flit` was quite arbitrary mostly to get the pyolite kernel up and running quickly (and it's simple to use), but we could switch to something else if needed.

> What we _can't_ do is introduce a new package build chain: it's gotta build with `labextension build`, and end up as another plugin.

That would be ideal yes. Just need to check whether we can keep that workflow given that the jupyterlite server app is (for now at least) a separate Lumino application.

bollwyvl:
> Using `flit` was quite arbitrary

I _strongly desire_ being able to ship jupyter tools with `flit`, if only for the [reproducible wheels](https://reproduciblewheels.com/) but we'll be stuck with `setup.cfg` and `setuptools` for actually shipping the built applications for the foreseeable... once we're worrying about cache busting and being able to confidently _update_ jupyterlite deploys, this gets pretty clutch...

.eslintrc.js
jtpio: So do you think we should keep that `ignore` clause, or leave it like it was before?

It might sound like a cosmetic thing mostly, but I was thinking we could try to follow the [JupyterLab TypeScript Style Guide](https://github.com/jupyterlab/jupyterlab/wiki/TypeScript-Style-Guide) to make it easier for folks to switch between code bases and contribute to these projects. It doesn't explicitly say anything about strict equality, but I guess it would try to follow the JupyterLab eslint / prettier config when that makes sense.

app/config-utils.js
jtpio: So top level await seems to be supported by Chromium 90 and Firefox 89.0b7 (developer edition). But not in Firefox 88.

docs/configuring.md
```diff
@@ -0,0 +1,117 @@
...
+inspect the `<script id="jupyter-config-data">` in the `<head>`. The entry you need will
+be contained there.
+
+Update your `/app/jupyter-lite.json` like so:
+
+```json
+{
+  "federated_extensions": [
+    {
+      "extension": "./extension",
```
jtpio, line 98: Maybe we'll also want to document the `"style": "./style"` key for loading style modules for the federated extensions that have one.

bollwyvl: that'd be about 10 lines below that... i don't think there's a lot to say, people shouldn't care, just keep kalm and copy-pasta... until they just need to know the _name_ of the extension, and can ignore all this stuff.

jtpio: Ah that hint was too big to be missed :laughing: 

image

But yeah this will be handled by some convenience tool anyway at some point.

packages/settings/src/settings.ts
jtpio: Yeah absolutely.

Just pushed c0bff5fe00cd05ad693f32d6b9c40d9503502ed9 in the meantime which should fix it.

It adds the network requests back to fetch `all.json` and the federated extension schemas too. But it would fine to take a look at it again in a separate PR if we want to optimize this.

jtpio: This should also handle editing settings for a federated extension via the settings editor.

packages/settings/src/settings.ts
```diff
@@ -59,13 +50,80 @@ export class Settings {
...
+  }
+
+  /**
+   * Get the settings for a federated extension
+   *
+   * @param id The id of the federated extension
+   */
+  export async function getFederated(id: string): Promise<IPlugin | undefined> {
+    const [ext, plugin] = id.split(':');
+    let federated: string[];
```
bollwyvl, line 97: oops, i think this comes back as a `{name, load, extension, style}`... will update...

packages/settings/src/settings.ts
```diff
@@ -18,54 +18,129 @@ export class Settings {
...
+    }
+
+    return found;
   }
 
   /**
    * Get all the settings
    */
   async getAll(): Promise<{ settings: IPlugin[] }> {
+    return Private.getAll();
```
bollwyvl, line 42: so... these will definitely _not_ pick up federated stuff at present...

packages/settings/src/settings.ts
bollwyvl: yeah, settings editing works now that the `isFederated` check is accurate

app/config-utils.js
jtpio: Actually do we need the await here? Or could this be changed to the following (at least for now):

```typescript
void main();
```

bollwyvl:
Yeah want to circle back on that... Ideally it would be in ts, as part of a
single build, with a cache-busting target, as if you get a bad cache of
that and things change, it's not going to go well.

Then each of the Index.html would also have to point at the cache-busted
script, which complicates matters, but it's worth it.

Also want to hoist the schema to ts types.

jtpio:
@bollwyvl I just pushed a couple of minor tweaks to:

- remove the top level await so it works in the latest stable Firefox. And we can probably add it back later once FF 88 is the new stable
- switch back to the previous eslint rule regarding the `null` checks (just to keep things consistent but could be added later again if needed)

Otherwise I think it looks really good for a first iteration on federated extensions! We can merge it as is and iterate in follow-up PRs and issues.

bollwyvl:
Sure, sounds good if you're feeling it. We should hoist the TBD from this
to new issues if not already done... Will look later today

jtpio:
This is really great stuff, thanks again Nick!

PR #62, opened 2021-04-22
add stopgap http server with mime types from python

### LINKED ISSUE #59: Serve wasm file with the `application/wasm` mime type
This doesn't seem to be the case at the moment when serving Pyodide from a local server started with `yarn serve` (which uses [`http-server`](https://www.npmjs.com/package/http-server)):

image

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlite/pyolite-serve-js-wasm?urlpath=lab)

## References

- fixes #59

## Code changes

- drops `http.server`
- add `scripts/serve.js`
  - not in-tree, because we can't ship this trash
  - mostly derived from, and may need better attribution to:
    - [MDN](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Node_server_without_framework)
    - [cpython](https://github.com/python/cpython/blob/3.9/Lib/mimetypes.py)

## User-facing changes

- n/a

- uses env vars instead of parsing argv :shrug:
- sets them both to bind to `127.0.0.1` because it's _literally_ the least we can do

### DIFF
--- a/.binder/jupyterlite-proxy.json
+++ b/.binder/jupyterlite-proxy.json
@@ -9,10 +9,13 @@
         }
       },
       "jupyterlite-js": {
-        "command": ["yarn", "serve", "--port", "{port}"],
+        "command": ["yarn", "serve"],
         "launcher_entry": {
           "icon_path": "docs/_static/icon.svg",
           "title": "JupyterLite (js)"
+        },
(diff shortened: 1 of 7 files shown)


jtpio:
Thanks!

PR #64, opened 2021-04-24
Add docs build for ReadTheDocs

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlite/add-docs-build?urlpath=lab)

## References

- starts #63

## Code changes

- [x] add `.readthedocs.yml`
  - [x] enabled  fail on warning, otherwise they stack up until it stops working
  - [x] sync the environment to binder 
- [x] Adds a sphinx site
  - [x] conf.py
  - [x] start some kernels stuff
    - [ ] couldn't figure out how to get e.g. `sphinx-autodoc` to work against pyolite... this can be a separate issue
- [x] adds a `dodo.py`
- [x] update contributing
- [x] remove checked-in `whl` 

## User-facing changes

> n/a

## Backwards-incompatible changes

- moved `pyolite` down one folder, since we might end up with `traitlites`, `litedgets`

PR #68, opened 2021-04-26
Add CI job to deploy to Vercel

jtpio:
## References

Complement https://github.com/jtpio/jupyterlite/issues/63 and #64.

This adds an extra job to the CI workflow to upload the tarball to Vercel, and is additive to deploying the app on RTD.

Should help with documenting deployments (https://github.com/jtpio/jupyterlite/issues/67)

We may or may not want to keep that step in the future if it's unnecessary. 

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,3 +68,36 @@ jobs:
       - name: Test
         run: |
           doit test
+
+  deploy:
+    needs: [build]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - uses: actions/download-artifact@v2
+        with:
+          name: jupyterlite dist ${{ github.run_number }}
+          path: ./dist
+      - name: Extract the app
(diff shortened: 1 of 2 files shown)


bollwyvl:
Yeah, that's cool: the more (working) examples we have the better. Github Pages as well... though it would be worth doing a fresh `--orphan` for `gh-pages` every time... we don't need the whole history of 75mb for every tag :scream:, and people shouldn't rely on being able to pull our `gh-pages` and expect patches to work indefinitely.

jtpio:
Looks like it worked.

jtpio:
I think we can indeed keep this extra deployment.

And make RTD the official source of truth and update remaining links as mentioned in https://github.com/jtpio/jupyterlite/issues/63#issuecomment-826856800.

jtpio:
Added some notes to https://github.com/jtpio/jupyterlite/issues/67#issuecomment-827091342 that we can use when adding documentation about deployments.

Issue #70, opened 2021-04-27
doit build failed on windows, The system cannot find the file specified


liudonghua123:
## Description

I followed https://github.com/jtpio/jupyterlite/blob/main/CONTRIBUTING.md, and when I executed `doit`, I got the following errors.

```shell
(common) D:\code\python\jupyterlite>doit list --all --status
R build                  build code and intermediate packages
R build:js:app:classic   build JupyterLite Classic with webpack
R build:js:app:lab       build JupyterLite Lab with webpack
R build:js:lib           build .ts files into .js files
R build:js:pack          build the JupyterLite distribution
U build:py:pyolite       build the pyolite python package for the brower with flit
R docs                   build documentation
R docs:sphinx            build the documentation site with sphinx
R env                    keep environments in sync
R env:binder
R lint                   format and ensure style of code, docs, etc.
R lint:black             format python files with black
R lint:eslint            format and verify .ts, .js files with eslint
R lint:prettier          format .ts, .md, .json, etc. files with prettier
R setup                  perform initial non-python setup
R setup:js               install node packages
R test                   test jupyterlite
R test:js                run the .js, .ts unit tests with jest
R watch                  watch sources and rebuild on change
R watch:docs             watch .md sources and rebuild the documentation
R watch:js               watch .ts, .js, and .css sources and rebuild packages and apps

(common) D:\code\python\jupyterlite>doit
.  lint:black
All done! ✨ 🍰 ✨
7 files would be left unchanged.
.  setup:js
Traceback (most recent call last):
  File "c:\users\liu.d.h\anaconda3\envs\common\lib\site-packages\doit\doit_cmd.py", line 190, in run
    return command.parse_execute(args)
  File "c:\users\liu.d.h\anaconda3\envs\common\lib\site-packages\doit\cmd_base.py", line 150, in parse_execute
    return self.execute(params, args)
  File "c:\users\liu.d.h\anaconda3\envs\common\lib\site-packages\doit\cmd_base.py", line 601, in execute
    return self._execute(**exec_params)
  File "c:\users\liu.d.h\anaconda3\envs\common\lib\site-packages\doit\cmd_run.py", line 264, in _execute
    return runner.run_all(self.control.task_dispatcher())
  File "c:\users\liu.d.h\anaconda3\envs\common\lib\site-packages\doit\runner.py", line 256, in run_all
    self.run_tasks(task_dispatcher)
[...]

PR #71, opened 2021-04-27
try to fix yarn/npm binary resolve issue on Windows

liudonghua123:
try to fix https://github.com/jtpio/jupyterlite/issues/70.

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -1,5 +1,6 @@
 import json
 import os
+import platform
 from pathlib import Path
 
 import doit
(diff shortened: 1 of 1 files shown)


dodo.py
```diff
@@ -258,7 +259,7 @@ class U:
     @staticmethod
     def do(*args, cwd=P.ROOT, **kwargs):
         """wrap a CmdAction for consistency"""
-        return doit.tools.CmdAction(list(args), shell=False, cwd=str(Path(cwd)))
+        return doit.tools.CmdAction(list(args), shell=platform.system() == 'Windows', cwd=str(Path(cwd)))
```
bollwyvl, line 262: Thanks for the PR!

See more thoughts on #70, but assuming all the shell tokens are escaped is usually bad for business once there's any kind of weirdness in strings (spaces, non-ascii, etc). Posted an alternate, painfully-won approach there...

I am not entirely sure if we can take on supporting windows as a fully-supported platform at this stage, and maybe ever if we get into bespoke emscripten stuff for pyolite, but if we do, it will have to involve a CI excursion.

bollwyvl (changes requested):

PR #77, opened 2021-04-28
fix windows build all.json broken

liudonghua123:
## References

try to fix https://github.com/jtpio/jupyterlite/issues/76.

### DIFF
--- a/app/webpack.config.js
+++ b/app/webpack.config.js
@@ -35,7 +35,8 @@ fs.copySync(
 );
 
 // ensure all schemas are statically compiled
-const schemaDir = path.resolve(buildDir, './schemas');
+// fix windows path issue, use / as path separator instead of \
+const schemaDir = path.resolve(buildDir, './schemas').replace(/\\/g, '/');
 const files = glob.sync(`${schemaDir}/**/*.json`, {
   ignore: [`${schemaDir}/all.json`]
 });
--- a/app/webpack.config.js
+++ b/app/webpack.config.js
@@ -36,7 +36,7 @@ fs.copySync(
 
 // ensure all schemas are statically compiled
 // fix windows path issue, use / as path separator instead of \
-const normalizePathSep = filePath => filePath.replaceAll(path.sep, path.posix.sep);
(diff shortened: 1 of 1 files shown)


app/webpack.config.js
```diff
@@ -35,7 +35,8 @@ fs.copySync(
 );
 
 // ensure all schemas are statically compiled
-const schemaDir = path.resolve(buildDir, './schemas');
+// fix windows path issue, use / as path separator instead of \
+const schemaDir = path.resolve(buildDir, './schemas').replace(/\\/g, '/');
```
bollwyvl, line 39: perhaps we use [`path.posix`](https://nodejs.org/api/path.html#path_path_posix) everywhere? There will be more of these as we add things like static files, federated extensions, etc.

liudonghua123: I tried [path.posix](https://nodejs.org/api/path.html#path_path_posix), the result of `path.resolve` use posix path separator indeed, however it reduced the driver letter. So [path.posix](https://nodejs.org/api/path.html#path_path_posix) will not fix this problem here.

```js
> buildDir = './build'
'./build'
> require('path').resolve(buildDir, './schemas')
'D:\\code\\python\\jupyterlite\\build\\schemas'
> require('path/posix').resolve(buildDir, './schemas')
'/code/python/jupyterlite/build/schemas'
>
```

bollwyvl:
Welp, any time we can lean on stdlib stuff rather than treating paths as
strings is a win... If we do have to, we should hoist it to a function e.g.
normalizePath...

liudonghua123:
> Welp, any time we can lean on stdlib stuff rather than treating paths as strings is a win... If we do have to, we should hoist it to a function e.g. normalizePath...

Wrap the path operation using a function like `normalizePath` is a good choice, it makes the code more readable. I will try to patch the pr.

bollwyvl:
This is now under test in [CI](https://github.com/jupyterlite/jupyterlite/actions/workflows/win-dev.yml), so presumably this is (and will stay) fixed.

Issue #83, opened 2021-05-04
Default list of federated extensions on the demo website


jtpio:
Follow-up to #58 

To showcase the use of existing federated extensions for JupyterLab (and JupyterLab Classic):

- `ipydrawio`
- `jupyterlab-cell-flash`
- `jupyterlab-tour`: #31 
- cool themes
- ?

This is to fix the following TODO:

https://github.com/jtpio/jupyterlite/blob/373b8aaeb4b92e9fc7697d2bedd51c5125c14457/docs/conf.py#L97-L99

bollwyvl:
Thanks for repping those first ones... Kinda torn as we want something in
there quick, but really want to start moving towards the "real deal"
tooling.

As mentioned, I move that we put darcula in that way instead of compiled in.

Some more wins for (maybe) docs:

- jupyterlab-videochat
  - needs some decoupling from serverextension, working on it
  - might need settings overrides
  - ain't gonna work in classic, yet, though
- jupyterlab-markup
  - though we kinda need the myst stuff to really hit it out of the park
- some of the jupyterlab-renderers
  - just because

Of course, the mime ones won't be fun until we can ship contents.

jtpio:
>   - ain't gonna work in classic, yet, though

I think it's totally fine if they don't all work in classic. It would already be really impressive to have it in lab.

> - some of the jupyterlab-renderers - just because

Those will be useful so we can double check loading mime extensions as federated works as expected.

bollwyvl:
Yep, perhaps for stuff we _know_ doesn't work with classic we can start it
just in /lab/jupyterlite.json... just makes the docs:extensions slightly
more complicated.

I think any optional goodies that don't make the site _too_ much slower to
load initially is good.

Part of this is on extension developers, for sure, but we should make sure
first party lite stuff is as lazy / non-blocking as possible.

For example, a servlitextension needs to register its routes, but doesn't
actually need to load its deps until a request actually comes in... Though
it could schedule a prefetch.

Similarly, the hard preload of matplotlib/numpy is kinda crushing to
time-to-execute-reply...

bollwyvl:
I think with `disabledExtensions` in, we are doing this pretty well.

jtpio:
Yes, and we can add more extensions to the demo site later if we want to.

Issue #81, opened 2021-05-03
Allow overriding plugin settings 


bollwyvl:
### Problem

When deploying a JupyterLite, some default settings might not be right for the use case of a particular deployment. It should be easy to finely-tune default settings, down the tool level (e.g. use one theme in `/lab`, and another in `/classic/notebooks` and _another_ in `/classic/editor`).

### Proposed Solution

Building on #58, add a new section to `jupyter-lite.json` that has the same structure as `overrides.json` (no JSON5, thanks!). So to disable the modal command palette:

```json
{
  "jupyter-lite-schema-version": 0,
  "jupyter-config-data": {},
  "settings-overrides": {
    "@jupyterlab/apputils-extension:palette": {"modal": false}
  }
}
```

Alternately, these could right into `jupyter-config-data`, but it wouldn't be "canonical"...

### Additional context

Wherever it lands, the Schema `definition` would be something like:

```json
"settings-overrides": {
  "description": "plugin-keyed settings that replace the defaults. See https://jupyterlab.readthedocs.io/en/stable/user/directories.html#overrides-json",
  "type": "object",
  "patternProperties": {
    "that-enormous-regex:.*": {
      "type": "object"
    }
  }
}
```

In `settings.ts`, these would be read either by `PageConfig` (simple) or we have equivalent `jupyter-lite.json` reading equipment on the "app" side (more flexible).

jtpio:
> add a new section to `jupyter-lite.json` that has the same structure as `overrides.json` (no JSON5, thanks!).

Sounds good :+1: If we can keep most of the configuration system similar to what JupyterLab already does, it should hopefully help keep things simple (and help documenting this).

bollwyvl:
Another approach might be to know about `overrides.json` in each folder, merging them upwards... it's more http requests that could 404, though...

bollwyvl:
Starting this now, going to keep in in `jupyter-lite.json` for the time being.

PR #84, opened 2021-05-04
Add list of federated extensions for the demo site

jtpio:
## References

Initial steps for #83 

## Code changes

Demo site changes only.

## User-facing changes

Users of the RTD demo will be able to use some federated extensions by default.
 

None

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -81,24 +81,32 @@ html_context = {
 }
 
 
-if RTD:
-    subprocess.check_call(
-        ["doit", "build", "docs:extensions", "docs:typedoc:mystify"], cwd=str(ROOT)
-    )
-
-
 def clean_schema(app: Sphinx, error):
-    if error:
-        return
+    """sphinx-jsonschema makes duplicate ids. clean them"""
+    print("jupyterlite: Cleaning generated ids in JSON schema html...")
     for schema_html in Path(app.builder.outdir).glob("schema-v*.html"):
         text = schema_html.read_text(encoding="utf-8")
         new_text = re.sub(r'<span id="([^"]*)"></span>', "", text)
         if text != new_text:
             schema_html.write_text(new_text, encoding="utf-8")
 
-    # if RTD:
-    #     subprocess.check_call(["doit", "docs:extensions"], cwd=str(ROOT))
+
+def before_rtd_build(app: Sphinx, error):
+    """this performs the full frontend build, and ensures the typedoc"""
+    print("jupyterlite: Ensuring built application...")
+    subprocess.check_call(
+        ["doit", "-n4", "build", "docs:typedoc:mystify"], cwd=str(ROOT)
+    )
+
+
+def after_rtd_build(app: Sphinx, error):
+    """copy the local labextensions and patch `_build_static/jupyter-lite.json`"""
+    print("jupyterlite: Copying installed labextensions from environment...")
+    subprocess.check_call(["doit", "-s", "docs:extensions"], cwd=str(ROOT))
 
 
 def setup(app):
     app.connect("build-finished", clean_schema)
+    if RTD:
+        app.connect("config-inited", before_rtd_build)
+        app.connect("build-finished", after_rtd_build)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -83,8 +83,10 @@ html_context = {
 
 def clean_schema(app: Sphinx, error):
     """sphinx-jsonschema makes duplicate ids. clean them"""
-    print("jupyterlite: Cleaning generated ids in JSON schema html...")
-    for schema_html in Path(app.builder.outdir).glob("schema-v*.html"):
+    print("jupyterlite: Cleaning generated ids in JSON schema html...", flush=True)
+    outdir = Path(app.builder.outdir)
+    for schema_html in outdir.glob("schema-v*.html"):
+        print(f"... fixing: {schema_html.relative_to(outdir)}")
         text = schema_html.read_text(encoding="utf-8")
         new_text = re.sub(r'<span id="([^"]*)"></span>', "", text)
         if text != new_text:
@@ -93,15 +95,19 @@ def clean_schema(app: Sphinx, error):
 
 def before_rtd_build(app: Sphinx, error):
     """this performs the full frontend build, and ensures the typedoc"""
-    print("jupyterlite: Ensuring built application...")
+    print("jupyterlite: Ensuring built application...", flush=True)
     subprocess.check_call(
         ["doit", "-n4", "build", "docs:typedoc:mystify"], cwd=str(ROOT)
     )
 
 
 def after_rtd_build(app: Sphinx, error):
-    """copy the local labextensions and patch `_build_static/jupyter-lite.json`"""
-    print("jupyterlite: Copying installed labextensions from environment...")
+    """copy the local labextensions and patch `_build/_static/jupyter-lite.json`"""
+    print(
+        "jupyterlite: Copying installed labextensions from environment...", flush=True
+    )
+    env = dict(**os.environ)
+    env["JLITE_DOCS_OUT"] = str(app.builder.outdir)
     subprocess.check_call(["doit", "-s", "docs:extensions"], cwd=str(ROOT))
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -82,7 +82,9 @@ html_context = {
 
 
 if RTD:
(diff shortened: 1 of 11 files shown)


jtpio:
Looks like the `docs:extensions` task makes the RTD build time out.

It probably triggers a new full build for some reason, although this doesn't seem to be the case locally.

bollwyvl:
ah.

So `docs:extensions` expect to run _after_ the docs are built, as it _overwrites_ a file copied in by normal sphinx machinery. It was probably try to restart itself over and over again.

I've split up the "app" stuff in `conf.py` to make it a little more explicit, and add the `-s` to try to make _extra_ sure it doesn't try to recurse.

bollwyvl:
Hm, not sure what's happening here... i think we need to make it work the way you had it, and generate an extra static path someplace, relying on sphinx to do the layering properly.

docs/conf.py
```diff
@@ -46,7 +46,15 @@
...
 html_favicon = "../app/lab/favicon.ico"
-html_static_path = ["_static", "../app"]
+# rely on the order of these to patch json, labextensions correctly
+html_static_path = [
+    # docs stuff
+    "_static",
+    # the as-built application
+    "../app",
+    # extensions and patched jupyter-lite.json
+    "../build/env-extensions",
```
bollwyvl, line 56: this is now a bit more "normal means," and we let the path order take care of it

bollwyvl:
it built, but the paths aren't quite right yet... looking...

bollwyvl:
Oh, weird. If you change the settings, it warns:
```
Stylesheet failed to load: /en/84/_static/lab/build/themes/@telamonian/theme-darcula/index.css
```
... but then if you refresh the page, it correctly loads the extension.

so we'll probably need a servlite route to offer the themes stuff in the expected place, as they can be derived from metadata.

bollwyvl:
gah, it doesn't `fetch` them, it builds a `link`. So we unpack them into an overlay location... on it...

jtpio:
Really cool, thanks Nick for picking it up!

For theme yes, until now we used to statically generate the assets during the build:

https://github.com/jtpio/jupyterlite/blob/373b8aaeb4b92e9fc7697d2bedd51c5125c14457/app/webpack.config.js#L116-L121

Probably the custom theme manager should handle federated themes just like for the settings.

bollwyvl:
k, gotta get on some other stuff, but this worked locally...

bollwyvl:
welp, raise up the issue for theme manager, i guess, but things that we can handle by "copying files to a place" don't feel too bad at this stage.

jtpio:
Or yes copying to the right location like in https://github.com/jtpio/jupyterlite/pull/84/commits/931075559b64b4e072e1552eedf5f9ed4b142fb1 is even simpler.

dodo.py
```diff
@@ -557,28 +590,46 @@ def validate(schema_path, instance_path=None):
...
 
         extensions = []
         all_package_json = [
-            *B.DOCS_LAB_EXTENSIONS.glob("*/package.json"),
-            *B.DOCS_LAB_EXTENSIONS.glob("@*/*/package.json"),
+            *B.CACHED_LAB_EXTENSIONS.glob("*/package.json"),
+            *B.CACHED_LAB_EXTENSIONS.glob("@*/*/package.json"),
         ]
+        # we might find themes
+        themes = B.PATCHED_STATIC / "lab/build/themes"
```
jtpio, line 609: We could also add them to classic.

bollwyvl:
Well, copies are copies, they go out of sync. Also probably doesn't work in
classic. I think we need our files-on-disk to match upstream as closely as
possible... And get that multi-app dedupe. But probably not on this pr. I
think we're getting closer to where we can be working more things in
parallel, though.

bollwyvl:
I guess if you see a way to fix it once in the theme manager, go for that
and revert the copy craft!

jtpio:
Like you said for now it's fine to have it like this.

btw, feel free to add more extensions if you want! Probably we can add `jupyterlab-tour` without having to create a tour for now, since it ships with the default lab tour anyway.

jtpio:
Interesting, looks like we get `nbdime` from somewhere?

image

bollwyvl:
Yeah, tour ftw.

I think if we can get tour to [respect settings](https://github.com/jupyterlab-contrib/jupyterlab-tour/issues/15), #81 will give this (and all the other things) to us as a matter of course, plus this would be a good demonstration for someone that _doesn't_ want to do a custom labextension to add that one bit of state.

jtpio:
That would be really convenient :+1:

jtpio:
> Interesting, looks like we get `nbdime` from somewhere?

So `nbdime` seems be coming from `myst-nb` -> `jupyter-cache` -> `nbdime`.

bollwyvl:
b15603f copies themes to classic, and updates the "Hard Way" docs with the knowledge we've gathered... built locally, :crossed_fingers: !

bollwyvl:
Added the generated `jupyter-lite.json` to the configuring docs page, with some notes about the known unknowns in federated-extension land...

docs/configuring.md
```diff
@@ -83,7 +92,25 @@ determined by looking in `package.json` for the extension, specifically
 `#/jupyterlab/sharedPackages`.
 ```
 
-### Fill Out `federated_extensions`
+#### Handle theme assets
+
+The Theme Manager expect to be able to load theme CSS/font assets from
```
jtpio, line 97:
```suggestion
The Theme Manager expects to be able to load theme CSS/font assets from
```

docs/configuring.md
```diff
@@ -107,11 +134,43 @@ Update your `/app/jupyter-lite.json` like so:
...
 federated extensions will only be avaialable for pages within that file tree.
 
 ## Customizing Settings
 
 > _TBD_
+
+## About the Demo
+
+This documentation site contains the JupyterLite Demo (the **Try** buttons on the top of
+the screen) and use a number of techiniques described on this page.
```
jtpio, line 149:
```suggestion
the screen) and use a number of techniques described on this page.
```

docs/deploying.md
```diff
@@ -16,14 +16,82 @@ You can get an empty JupyterLite by:
...
+beyond it. It is well adapted to building sites of any size, and tools like [myst-nb]
+enable make it very palletable to include executable, and even interactive, content.
+
+JupyterLite assets can be copied to the default static directory in `conf.py`, e.g.
+`docs/_static` with [`html_static_path`](#html_static_path), or replace the entire site
+with [`html_extra_path`](#html_extra_path)
+
+### `html_static_path`
+
+This searach path can be merge several layers deep, such that your theme assets, the
```
jtpio, line 51:
```suggestion
This search path can be merged several layers deep, such that your theme assets, the
```

docs/deploying.md
```diff
@@ -16,14 +16,82 @@ You can get an empty JupyterLite by:
...
+    "_static",
+    "../upstream-jupyterlite",
+    "../my-jupyterlite"        # <- these "win"
+]
+```
+
+The composite directory will end up in `docs/_build/_static`.
+
+```{hint}
+See the JupyterLite [conf.py] for an example approach, though it likely a good
```
jtpio, line 65:
```suggestion
See the JupyterLite [conf.py] for an example approach, though it's likely a good
```

docs/deploying.md
```diff
@@ -47,22 +115,32 @@ local purposes.
 python -m http.server -b 127.0.0.1
 ```
 
+If you are using a recently-released Python 3.7+, this will correctly server
```
jtpio, line 118:
```suggestion
If you are using a recently-released Python 3.7+, this will correctly serve
```

docs/deploying.md
```diff
@@ -86,18 +164,29 @@ however, that `http-server` does not support the `application/wasm` MIME type.
...
-In a sphinx-based build, include the JupyterLite assets in `_static`, or configure with
-`html_static_path`.
+The [Sphinx](#sphinx) deployment approach will work almost transparently with
+[ReadTheDocs](https://readthedocs.org), for the small price of a `.readthedocs.yml` file
+in the root of your repository.
 
-```python
-html_static_path = ["_static", "../my-jupyterlite"]
+```{hint}
+See the JuptyerLite [.readthedocs.yml] for an example.
```
jtpio, line 185:
```suggestion
See the JupyterLite [.readthedocs.yml] for an example.
```

jtpio:
Thanks!

Issue #88, opened 2021-05-14
Real Time Collaboration


jtpio:
With JupyterLab `3.1.0a8` shipping with the initial RTC work, we might want to start considering what enabling real time collaboration would look like with JupyterLite.

### Problem

Enable real time collaboration with JupyterLite. Questions:

- should that apply to anyone using the same JupyterLite deployment (that might be too disruptive as a default)
  - this will already be useful to users opening multiple tabs with the same notebook
- should we enable it on the demo website by default?
  -  maybe we'll need some kind of token sharing mechanism (implemented as a separate plugin) to let people copy and send links to other folks

### Proposed Solution

Incremental steps we can take:

- start using the `3.1.0-alpha.x` `@jupyterlab` packages
- enable the `collaborative` flag in PageConfig
- use a different provider than the default websocket provider (will require a change upstream to allow for more flexibility)

bollwyvl:
Very pro this!

Semi-related: I am pursuing a monkeypatch-tastic approach on
jupyterlab-video chat, which I would combine with making the labextension
work without the server, which acts as a room metadata storage, but isn't
strictly required.

I think Jitsi has Good Enough information theory components to work as an
on-by-default bus on rtfd.io.
 the lab integration has an opt-in setting to allow for creating
unobfuscated room names, including a router URL, and then they take care of
all the trust.

jtpio:
For reference #89 added a plugin to replace the default provider:

https://github.com/jtpio/jupyterlite/blob/695373748db6c5aa015224e94d4aa8679eb34ef2/packages/application-extension/src/index.ts#L51-L64

We can maybe switch to using the `WebrtcProvider` instead, and add a way to share tokens between users via query string parameters so they can connect to the same rtc room.

jtpio:
Closing as fixed by #109.

Follow-up items are in https://github.com/jupyterlite/jupyterlite/issues/241

PR #89, opened 2021-05-14
Update to `3.1.0-alpha.9` lab packages and retrolab

### LINKED ISSUE #22: Disable URL routing

### LINKED ISSUE #61: Update to the 3.1 `@jupyterlab` alpha packages
A couple of improvements landed upstream to make the plugins easier to reuse and replace.

So we should start thinking about updating to the new packages at some point to pull them in.

jtpio:
## References

Initial step for #88 

Fixes #22 
Fixes #61 
Closes #39 (replaces)

This will also make it possible to grab some fixes to upstream plugins so they can more easily be reused.

## Code changes

- Update lab dependencies to `3.1.0-alpha.9` packages
- Update classic to retro: https://github.com/jtpio/retrolab
- Add a small script to make sure to keep the `resolutions` in sync (light integrity check).

## User-facing changes

Some upstream changes will now also be visible in JupyterLite. For example there are now commands to create new python files directly:

image

Update from JuptyerLab Classic to [RetroLab](https://github.com/jupyterlab/retrolab):

image


jtpio:
Looks like it would be fine to start depending on upstream alpha packages since `jupyterlite` itself is still pretty alpha anyway.

bollwyvl:
I see no downside to helping alpha test this particular release.

We have the package available from `channels:
[conda-forge/label/jupyterlab_alpha, conda-forge]`, which might be a source
of some of the problems on CI... On phone, so can't see logs...

jtpio:
Nice thing is we automatically get the application commands back:

image

jtpio:
Updated to alpha 9 (which includes https://github.com/jupyterlab/jupyterlab/pull/10256).

With a simple plugin and using the `WebSocketProvider` with `wss://demos.yjs.dev`, we get some basic support RTC:

https://user-images.githubusercontent.com/591645/119104897-168ba300-ba1d-11eb-8b43-0b69ab2a0d48.mp4

Need to check what's going on with the caret CSS, and new cells being inserted automatically. Maybe we'll need to properly implement the (`IDocumentProvider`](https://github.com/jupyterlab/jupyterlab/blob/abcb9242af1a97fa9e897d34f0556e07bb913fa3/packages/docprovider/src/tokens.ts#L15) interface.

packages/application-extension/src/index.ts
```diff
@@ -3,13 +3,131 @@
...
+
+import { IMainMenu } from '@jupyterlab/mainmenu';
+
 import { ITranslator, TranslationManager } from '@jupyterlab/translation';
 
+const YJS_WEBSOCKET_URL = 'wss://demos.yjs.dev';
+
+class WebSocketProvider extends WebsocketProvider implements IDocumentProvider {
+  constructor(options: IDocumentProviderFactory.IOptions) {
+    super(YJS_WEBSOCKET_URL, options.guid, options.ymodel.ydoc);
```
jtpio, line 26: For now this is passed the path to the file:

https://github.com/jupyterlab/jupyterlab/blob/abcb9242af1a97fa9e897d34f0556e07bb913fa3/packages/docregistry/src/context.ts#L86

Maybe we can add another plugin that would create a uuid, that could then be shared with other folks via a query string parameter, and be concatenated to the `guid`.

jtpio: Or use the jupyterlab-video chat approach: https://github.com/jtpio/jupyterlite/issues/88#issuecomment-841518843

jtpio:
cc @bollwyvl if you want to check it out on RTD (since this is updating quite a bunch of stuff)

I'm seeing some weird behavior in the file browser after updating to alpha 10, which brings this new dialog:

image

Which was added in: https://github.com/jupyterlab/jupyterlab/pull/10043

Might be an upstream issue, but we can look into it separately.

jtpio:
This PR also leaves RTC disabled for now. We can tackle it as part of #88 probably by implementing a token-based sharing approach using the `WebrtcProvider`.

bollwyvl:
> This PR also leaves RTC disabled for now. We can tackle it as part of #88
> <https://github.com/jtpio/jupyterlite/issues/88> probably by implementing
> a token-based sharing approach using the WebrtcProvider.
>
Right. Until the UI is a lot clearer _when_ you're going"on the air",
off-by-default is good.

>

bollwyvl:
I checked it out on RTD... everything seems on the up and up, and it doesn't seem like much can happen until we land this... still getting some SEGSEV in chrome on that latter cells of the pyodide example, but I think that's just... a thing.

jtpio:
Thanks for checking :+1: 

So yeah let's get this in so it's easier to iterate. We can follow the lab 3.1 release cycle and update to newer packages when available.

Issue #86, opened 2021-05-07
Logo


jtpio:
The current logo is really cool: https://github.com/jtpio/jupyterlite/blob/main/docs/_static/icon.svg

And it would be great if we could use it for the GitHub organization when we start moving things there, or even in the README.md.

However it's not clear whether it complies to the design guidelines mentioned in https://github.com/jupyter/design/blob/master/brandguide/brand_guide.pdf

There is also a discussion about this in https://github.com/jupyter/design/issues/65.

As an alternative, we could also replace the semicircles by a plain circle (or something else), even though it would be nice to keep the semicircles:

![jupyterlite-logo-test](https://user-images.githubusercontent.com/591645/117472267-01147480-af59-11eb-9153-ba149fbc9c36.png)

bollwyvl:
![Screenshot from 2021-05-15 09-06-45](https://user-images.githubusercontent.com/45380/118362276-e5aef800-b55c-11eb-8036-34bd4954f7bb.png)

Close up, and with the potential "off" variant:
![Screenshot from 2021-05-15 09-08-26](https://user-images.githubusercontent.com/45380/118362338-260e7600-b55d-11eb-9ff6-b54455cea141.png)

bollwyvl:
it's unreadable in the favicon (no surprise) but it I do like it.

bollwyvl:
So some other things we could try:
- get rid of (or offer a variant without)  the "lite" word
- do a monochrome one
- grab a (or some) different colors from the stock ones
  - it's hard to pair with the orange, though maybe the purple
- deploy it as a LabIcon in the application layer
  - make that the canonical source for the icon, vs docs

bollwyvl:
Here are some wordmark explorations from #92.

![Screenshot from 2021-05-15 09-50-16](https://user-images.githubusercontent.com/45380/118375720-7a334d80-b591-11eb-91a2-f62de22dcc84.png)


![Screenshot from 2021-05-15 15-09-35](https://user-images.githubusercontent.com/45380/118375714-743d6c80-b591-11eb-80f0-8dafe286cc31.png)

jtpio:
Really like the one with the canary / yellow color too!

Wondering whether the black part would render well with the dark theme though?

bollwyvl:
When used in-theme, we can make it work with the semantic classnames...

bollwyvl:
Been a spell here, but how do we feel about making a badge folk can use to link to their demos?

[...]

PR #93, opened 2021-05-15
add wordmark

bollwyvl:
> of course open to **any** feedback!

Screenshot from 2021-05-15 16-17-31

## References

- continues #86

## Code changes

- the `wordmark.svg` and `logo.svg` both contain some more content in inkscape layers

## User-facing changes

- in the docs, there would be a word mark

- n/a

## Alternatives

> these are stored within `wordmark.svg`, which is probably not permanent.


jtpio:
Looks great, thanks!

Maybe we can go for the orange version first, and then check whether we would like to switch to another one? The yellow / black one also looks good.

Should we add it to the README as well? This could even replace the "JupyterLite" title and be placed above the badges.

bollwyvl:
>  go for the orange version first

Sounds good. I'll split out/optimize just ol' orange, and leave the working stuff in another file.

> Should we add it to the README as well?

Ehm, I reckon we should wait until it actually points at something on `main`.

jtpio (approved):
Thanks!

bollwyvl:
Oh, whoops, didn't optimize that... it's ok, we can do more with it in post.

Issue #85, opened 2021-05-07
Render Matplotlib Figure Example


owenlamont:
### Problem

I couldn’t figure out how to render a matplotlib figure in a notebook and don’t know if this is supported yet.

### Proposed Solution

If figure rendering is supported can an example be shown in these documentation - if unsupported can this also be noted in the documentation.

jtpio:
Thanks @owenlamont for the interest!

There will soon be example notebooks available on the [demo website](https://jupyterlite.rtfd.io/en/latest/try/lab) (#30).

In the meantime there is this gist with some info: https://gist.github.com/jtpio/3ba7feacc67431488e33a148429f96f1

You can download the notebook and then upload it via the JupyterLab interface if you would like to try it out:

![image](https://user-images.githubusercontent.com/591645/117410271-7d379980-af12-11eb-8661-c719336dc33f.png)


For now `matplotlib` in JupyterLite is pretty basic but should at least work with simple plots as static pngs. The Pyolite kernel currently runs in a web worker so interactivity is not supported yet. A sample code snippet would like this:

```python
import numpy as np
import matplotlib.pyplot as plt

x = np.linspace(0, 10, 1000)
plt.plot(x, np.sin(x));

plt.show()
```

There is ongoing work to make this better:

- add support for ipywidgets for JupyterLite: #18 
- canvas based renderer for matplotlib in Pyodide: https://github.com/pyodide/pyodide/pull/1579

owenlamont:
Cheers! Thanks for the examples. 

I got a little caught out expecting the current Jupyter lab behavior of auto rendering Matplotlib figures - I tried the old Matplotlib inline magic then realised magics don't seem to be parsed.

I was trying the Pandas DataFrame plot method and was wondering why I couldn't see anything.

jtpio:
The default Python kernel is indeed pretty minimal at the moment. So quite a few common things are still missing.

Ideally, it will be possible to reuse some of the features from existing kernels and `IPython`.

bollwyvl:
@owenlamont as of earlier today the `examples/` folder on the docs site (give it a good hard refresh!) should now show a minimal example of using matplotlib in pyolite. magics will come soon enough!

owenlamont:
Fantastic, thank you very much.

bollwyvl:
Going to close this, please reopen if you have any other concerns!

Issue #40, opened 2021-04-13
Provide a way for kernels and server extensions to expose static assets


jtpio:
In some way this sounds similar to having static assets distributed via Python packages.

It would be particularly useful for the pyodide kernel, so it could provide additional Python modules as-is and load them on the fly from a static url.

One way could be to leverage the webpack asset modules:

https://webpack.js.org/guides/asset-modules/

Although it would also be nice to support loading kernels and extensions from other remotes and CDN.

jtpio:
This would also be useful when providing kernel resources such as logos, which are for now manually placed in https://github.com/jtpio/jupyterlite/tree/main/app/kernelspecs

bollwyvl:
Been thinking about this more while fiddling with #48: the [approach in jyve](https://github.com/deathbeds/jyve/blob/master/packages/jyve-kyrnel-p5-unsafe/package.json#L21) was to get the stuff into `package.json` and having a custom key, and I feel this approach is worth considering. 

The `package.json` will always be part of an npm package, and is still consumable by other languages (a la #41). A critical thing to add would be a schema to keep multiple implementations honest... it seems like mirroring the api routes as closely as possible probably makes the most sense... but some special notation will be required as we're talking about:
- in-line JSON fragments
- individual files
- lists of files

For things like pyodide wasm, it's a little murkier: in a traditional server-backed setup, these would just be `!url-loader` or `!file-loader`d in, and then served from the `share/labextensions` path. 

Similarly, being able to specify `files` (a la #30) would also have enough information to drive contents without storing the base64-encoded JSON...

so, perhaps:

```yaml
"jupyterlite": {
  "routes": {
    "/api/kernelspecs#/kernelspecs/pyodide": {
      # the actual kernelspec
    },
    "/api/kernelspecs/pyodide/logo-32x32.png":  "style/logo-32x32.png", # an implicit path
    "/static/labextensions/@jupyterlite/pyodide-kernel/": [{ # a "mount point"
        "path": "pyodide-dist",
        "files": ["pandas.data", "pandas.wasm"] # etc. globs are... hard
    }],
    "/files/examples/": [{
      "path": "docs",
      "files": ["pyodide.ipynb"]
    }]
  }
}
```

in webpack land, these could be interepreted, potentially with that asset plugin, but there's also enough data here to do good things without it.

[...]

Issue #30, opened 2021-04-10
Create a set of example notebooks


jtpio:
These example notebooks could be added to the local storage by default (so available in the file browser on the first startup), and be used to quickly showcase a specific feature of a kernel or the language.

jtpio:
For now maybe the simplest would be to have:

- a new `examples-extension` package, with a plugin that depends on the [IContents](https://github.com/jtpio/jupyterlite/blob/1b402f869ac6d72507b12c0fa05075f72f4fd7b6/packages/contents/src/tokens.ts#L15) service
- a set of notebooks in `packages/examples-extension/notebooks`, as normal `.ipynb` files with the correct metadata already populated (so it starts the right kernel)
- on build, notebooks are placed under `app/examples/notebooks` as-is
- on plugin activation, `@jupyterlite/examples` fetch the examples and adds them to the local storage with [contents.save()](https://github.com/jtpio/jupyterlite/blob/1b402f869ac6d72507b12c0fa05075f72f4fd7b6/packages/contents/src/contents.ts#L290)
- to be less disruptive, they could even be placed under an `examples/`directory:

![image](https://user-images.githubusercontent.com/591645/115048431-d98e3700-9ed9-11eb-9ab1-2b3d67c593d5.png)

bollwyvl:
This is the part where folk will want to be able to put their stuff in and
all but _instantly_ get a working site, with notebooks, file assets, data
sets, that work with markdown, image viewers, etc. I see this as realizing
the /api/contents route, even if the _actual_ contents stay in their native
format.

As hinted at on #41, I think the part where one adds contents should
definitely not require webpack, nor should examples be shipped as part of
the distributed base assets. Really just should need to build up the JSON
tree, which should be doable in node or python (or whatever).

PR #94, opened 2021-05-16
Add user content

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlite/add-user-content?urlpath=lab)

## References

- semi-related to #40, #85, #30

## Code changes

- [x] some notional docs
- [x] adds a server-backed overlay backed by `files` and `api/contents`
  - [x] notebooks
  - [x] subdirectories
  - [x] text
  - [x] base64 
- [x] during the docs build
  - [x] copies some files
    - [x] `README.md`
    - [x] `docs/_static/logo.svg`
    - [x] a notebook
    - [ ] some other JSON (e.g. the site's `jupyter-lite.json`>)
    - [x] a PNG (or some other base64 thing) 
  - [x] builds `all.json`
    - i kinda think we need a less-collision-tastic name   

## Not On This PR
- [x] adds a `swagger.json` (not used yet for anything, but thinking about it)
  - [ ] use for validating fake static server API responses?
  - [ ] pulling this for now because https://github.com/jupyter-server/jupyter_server/issues/518

## User-facing changes

- on the docs site, `README.md`, and `docs/_static/logo.svg` should be visible
  - if someone has already made a `README.md`, they'll see that!

- :laughing: 

## Alternatives

While _put files in `files/`_ feels good, having to generate the `api/contents` response feels... less good.

It seems like we could have a base set of files we look for, and then allow subsequent files to be "automagically" found when they are linked. Some precedents:
- `files/index.html` would render a (trusted?) HTML view?
- `files/README.md` would render as a (trusted?) Markdown preview?
- `files/README.ipynb` same, _maybe_ restart and run all?

I feel like the supporting mechanism of "best effort contents" is doable here, but actually _picking_ the default files might need to be deferred, probably moved to `jupyter-config.json` schema... or replaced entirely with `.jupyter-workspace` files.


bollwyvl:
I think there's a latent refactor here, of some kind of ordered `ContentsLayer`.

The `LocalContentsLayer` should work by itself.

This new thing is like a `ReadonlyServerContentsLayer`.

A number of more advanced drivers are possible, e.g. `webdav`, `ipfs`, `dat`, I don't know how to think about it yet, but maybe they need some kind of "mount points", similar to [hybridcontents](https://pypi.org/project/hybridcontents).

bollwyvl:
... making progress with notebooks and images...

bollwyvl:
Editing notebooks, svg seem to be working... png loading works.

jtpio:
Thanks!

> but actually _picking_ the default files might need to be deferred

These defaults could indeed be quite arbitrary, and we can definitely think about it separately.

docs/configuring.md
```diff
@@ -41,17 +41,62 @@ schema-v0
...
+- Paste this JSON in `$YOUR_JUPYTERLITE/api/contents/all.json`
+- Copy your files in `$YOUR_JUPYTERLITE/files`
+- Repeat this for every subfolder `:(`
 
-- _Copy your files in `$YOUR_JUPYTERLITE/files`._
-- _Do something to generate valid Jupyter Server API responses_
-- _Put them someplace_
+Now, when the app reloads, these files will appear in the File Browser _if_ there isn't
+an existing file of that name in browser storage. If a user _has_ created such a file,
+and is deleted, the original server-backed file will become visible.
```
jtpio, line 99: This layering mechanism feels really nice :+1:

bollwyvl: Yeah... it would be nice to be able to reflect the state of the source of the content, but we can't do much while still fulfilling the contract everywhere, which seems like the most important property to maintain. The whole `IDrive` thing would be confusing, even though we _will_ need it eventually.

bollwyvl:
I dropped pulling in the `swagger.json` from `jupyter_server`, as it has false positives for our content.

JSON is working fine, but always loses its formatting... I don't think this is overly troublesome for demo purposes, and can be fixed in post, but eventually, we'd want to ensure that stuff stays consistent so that lite-edited files can be part of a sane DVCS-backed experience. For notebooks, this means getting the canonical concrete format, for all its warts. For everything else... need to just do what the user says when working with it as a string.

jtpio:
Thanks again Nick for this!

Checking it out on RTD, it looks like creating a new notebook throws an error:

image

bollwyvl:
Yeah that's a problem! 😊 Should be able to track it down quickly...

bollwyvl:
Yep, `fetch` doesn't throw, just needed to check `response.ok`.

Also: did verify that "hidden" files (e.g. exists in `/files`, but doesn't have listings in `api/contents`) already somehow just work. So if one did have, for example, a `README.md` that linked to a bunch of things, one could "slow reveal" the content stored in the site without having to mess with pre-indexing. They just kinda spring into existince, and don't appear in the tree until you actually press save... but I'll take it!

bollwyvl:
Yerp, confirmed it worked on rtd.

jtpio (approved):
Thanks!!

Issue #31, opened 2021-04-10
Add a Tour of JupyterLite


jtpio:
Using https://github.com/jupyterlab-contrib/jupyterlab-tour

This would help guide new users of the JupyterLite interface, and explain that kernels run the browser.

bollwyvl:
Tours are a _whacking_ good idea, and best of all, simple ones can be encoded in JSON (https://github.com/jupyterlab-contrib/jupyterlab-tour/issues/2).

Going beyond the docs for _this_ repo, I'd definitely see making custom tours something that can be added without re-building the app, as could be enabled/built in #41.

bollwyvl:
started https://github.com/jupyterlab-contrib/jupyterlab-tour/pull/17

bollwyvl:
That got merged, not sure release timeline, but would give us something concrete to develop #81 around.

In doing docs for that, turns out lab core has a `pageConfig.json`... perhaps it's worth looking into that, and divesting `jupyter-config-data` from `jupyter-lite.json`. I'm thinking any time there is a canonical file, we should use it instead of making up new things... in this case, it doesn't get loaded at runtime, but still...

bollwyvl:
tour 3.0.1 landed with the settings stuff, so i'll get #94 buttoned up (just validating JSON at this point, I think) and start into #81...

jtpio:
Thanks!

Pasting the ideas mentioned in https://github.com/jtpio/jupyterlite/pull/98#issuecomment-843168648 here for visibility

---

> Also what do you think the JupyterLite tour should cover?

Ain't that the way! Some ideas, out of order:

-  highlight the file browser icon
   - _Where are my files stored?_
   - _JupyterLite starts with files hosted along with the application_
   - _You can upload new files, make copies, and change existing tiles. These are stored in your browser's localStorage or IndexDB_
   - _If you delete a browser file, or work in Private browsing and close your browser, it's GONE FOR REAL_
      - > maybe we need a trashlite?
- hilight the file menu
  - _If you've used JupyterLite before, there (should) be some things missing here_
  -  _like Download As_
    - _We hope to support downloading your current JupyterLite site as an archive, though_
  
- For each kernel, as they grow in complexity, we may need their own tour of their notebook
  - how to import libraries  
    - or in the case of pyolite, _install_ libraries
    - and why they might not install/import properly
  - specific limitations
 
[...]

PR #98, opened 2021-05-17
Support settings overrides, add basic tour

### LINKED ISSUE #81: Allow overriding plugin settings

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlite/gh-81-support-overrides?urlpath=lab)

Screenshot from 2021-05-17 11-30-36

## References

- fixes #81
- enables #31 (i mean, i am adding a tour, but it's boring!)

## Code changes

- [x] update server settings route regex to handle `-` characters
- [x] add `pluginOverrides` to `jupyter-lite.json`
- [x] handle overriding defaults the same way as upstream (top-level keys)
- [x] add a very simple tour
- [x] validate settings against schema (if available)  in `doit check`
- [ ] validate nesting (e.g. use `darcula` in `classic`)

## User-facing changes

- deployers can now specify an `settingsOverrides` in their `jupyter-lite.json`
- users will see these new defaults, unles they have set their own

- n/a

## Alternatives
- as a real, well-known, file, `overrides.json` _could_ be used at-rest, which would make it easier to reuse with e.g. binder
  - however, it's not available transparently to the frontend at runtime, and only supports a single instance... also, it will be easier to build interactive tooling around it, if it's not a separate file.


jtpio:
Nice, thanks!

Looks like there is a small conflict to fix now that #94 is merged. 

Also what do you think the JupyterLite tour should cover?

bollwyvl:
> Also what do you think the JupyterLite tour should cover?

Ain't that the way! Some ideas, out of order:

-  highlight the file browser icon
   - _Where are my files stored?_
   - _JupyterLite starts with files hosted along with the application_
   - _You can upload new files, make copies, and change existing tiles. These are stored in your browser's localStorage or IndexDB_
   - _If you delete a browser file, or work in Private browsing and close your browser, it's GONE FOR REAL_
      - > maybe we need a trashlite?
- hilight the file menu
  - _If you've used JupyterLite before, there (should) be some things missing here_
  -  _like Download As_
    - _We hope to support downloading your current JupyterLite site as an archive, though_
  
- For each kernel, as they grow in complexity, we may need their own tour of their notebook
  - how to import libraries  
    - or in the case of pyolite, _install_ libraries
    - and why they might not install/import properly
  - specific limitations
 
Will do some more thinking... also, big dummy me for not doing it when i did the settings PR, but we should be able to put tours in notebook's metadata.

jtpio (approved):
Nice, thanks!

PR #100, opened 2021-05-18
Add more examples

jtpio:
## References

Follow-up to #94

Fixes #30.

Adding a couple of examples that were until now stored in the browser local storage for demo purposes.

## Code changes

New examples.

## User-facing changes

New examples on the demo website.

## Backwards-incompatible changes

None

PR #102, opened 2021-05-18
Support disabledExtensions

### LINKED ISSUE #101: Full docs site is noisy with errors
## Description

On the docs, all the `nbdime` stuff shows up in the UI, but very little works, making lots of noise.

## Reproduce

- open the docs
- open the developer tools
- see lots of errors in console
- open a notebook
- click nbdime buttons
- see ugly "cannot fetch" errors

## Expected behavior

- now
  - see no errors when the console opens
    -  test this #6
- eventually
  - nbdime works

## Context

- RTD

## Alternatives

- specifically don't copy `nbdime` in the `doit docs:extend`
  - pro
    - low-cost
    - doesn't ship the artifacts in the tarball
  - con
    - doesn't document how to disable extensions
    - other extensions might need it from `sharedPackages` and not bundle
- add a disabled plugins key to PageConfig
  - pro
    - good documentation
  - con
    - maybe not canonical?
    - ships the assets in the tarball

bollwyvl:
## References

- fixes #101
- includes #100 (whoops, but fine, really)

## Code changes

- adds `disabledExtensions` to schema
  - didn't run it down too much, but `jupyter labextension disable` Did Some Things to `jupyter-config-data`, so I'm assuming it's right
- adds reading of these from `PageConfig` to `config-utils.js`
- updates some docs
- disables `nbdime` and `jupyter-server-proxy` in docs

## User-facing changes

- broken `nbdime` buttons will not show up
- console will be much cleaner (just our known, wontfix `404`s)

- n/a

## Alternatives

As mentioned on #101, we could also just _not copy_ `nbdime` and `@jupyterlab/server-proxy` into the bundle, but kicking the tires on disabling some stuff dynamically seems like a good exercise.


jtpio (approved):
Thanks!

PR #109, opened 2021-05-28
Initial support for real time collaboration

jtpio:
## References

#88 

**To try it out**

https://jupyterlite--109.org.readthedocs.build/en/109/_static/lab/index.html?room=the-coolest-room

https://user-images.githubusercontent.com/591645/120011349-787c7780-bfde-11eb-94a5-f8ae0808128e.mp4

## Code changes

- [x] Switch to `WebrtcProvider`
- [x] Set `collaborative` to `true` on the demo site
- [x] Basic link sharing to group users in rooms
- [x] ~Extra URL parameter to enable RTC? For example `?rtc=1` to enable, disabled otherwise~ -> not needed
- [x] Update the docs to show how to use RTC features

## User-facing changes

Add options to enable RTC to end users.

### DIFF
--- a/docs/rtc/index.md
+++ b/docs/rtc/index.md
@@ -23,5 +23,25 @@ The main differences compared to JupyterLab are:
 
 ## Enabling RTC in JupyterLite
 
+To enable RTC in JupyterLite, you need to set the `collaborative` flag in the
+`jupyter-config-data`. This follows the same behavior as in JupyterLab 3.1 where real
+time collaboration is an opt-in feature.
+
+In addition to the `collaborative` flag, end users must specify the `room` query
+parameter in the URL. An example of such URL is as follows:
+
+[https://jupyterlite.readthedocs.io/en/latest/\_static/lab/index.html?room=my-custom-room](https://jupyterlite.readthedocs.io/en/latest/_static/lab/index.html?room=my-custom-room)
+
+Users are grouped together in rooms using a combination of:
+
+- the host, for example `jupyterlite.example.com` or `myserver:5000`
+- the name of the room parsed from the query string parameter, for example
+  `my-custom-room`
+
+Currently both `collaborative` and `room` must exist for RTC to be enabled. When this is
+the case, users can collaborate on documents together:
+
+TODO: add screencast
+
 See the [configuring](../configuring.md) section of the docs for more details on how to
 configure `jupyter-lite.json`.
--- a/docs/rtc/index.md
+++ b/docs/rtc/index.md
@@ -41,7 +41,7 @@ Users are grouped together in rooms using a combination of:
 Currently both `collaborative` and `room` must exist for RTC to be enabled. When this is
 the case, users can collaborate on documents together:
 
-TODO: add screencast
+![rtc-rtd](https://user-images.githubusercontent.com/591645/125045331-2af33000-e09d-11eb-909b-a7d1d46910f6.gif)
 
 See the [configuring](../configuring.md) section of the docs for more details on how to
 configure `jupyter-lite.json`.
--- a/docs/rtc/index.md
+++ b/docs/rtc/index.md
@@ -1,0 +1,27 @@
+# Real Time Collaboration
+
+JupyterLite enables basic support for Real Time Collaboration over WebRTC.
+
+## What is Real Time Collaboration
+
+Real Time Collaboration (RTC for short) makes it possible for users to collaborate on
+the same document at the same time.
+
+Here is an example screencast of what it looks like in JupyterLab:
+
+![rtc-demo](https://user-images.githubusercontent.com/591645/117701750-e6940280-b1c7-11eb-92e6-2ce0331febeb.gif)
+
+Since JupyterLite reuses most of the upstream JupyterLab components and plugins, it also
+gets support for RTC almost automatically.
+
+The main differences compared to JupyterLab are:
+
+- the user content is decentralized: users store notebooks and files in their own
+  browser local storage
+- the communication with peers and the initial syncing mechanism happens over WebRTC,
+  instead of the WebSocket Jupyter Server handler in JupyterLab
+
+## Enabling RTC in JupyterLite
+
+See the [configuring](../configuring.md) section of the docs for more details on how to
+configure `jupyter-lite.json`.
--- a/docs/rtc/index.md
+++ b/docs/rtc/index.md
@@ -29,8 +29,7 @@ time collaboration is an opt-in feature.
 
 In addition to the `collaborative` flag, end users must specify the `room` query
 parameter in the URL. An example of such URL is as follows:
-
-[https://jupyterlite.readthedocs.io/en/latest/\_static/lab/index.html?room=my-custom-room](https://jupyterlite.readthedocs.io/en/latest/_static/lab/index.html?room=my-custom-room)
+`you-lite-deployment.example.com/lab/index.html?room=4869637241`
 
 Users are grouped together in rooms using a combination of:
 
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -6,6 +6,7 @@
 :maxdepth: 2
 
 kernels/index
+rtc/index
 ```
 
 ## Frequently-Asked Questions
--- a/examples/jupyter-lite.json
+++ b/examples/jupyter-lite.json
@@ -1,6 +1,7 @@
 {
   "jupyter-lite-schema-version": 0,
   "jupyter-config-data": {
+    "collaborative": true,
     "disabledExtensions": [
       "@jupyterlab/server-proxy",
       "jupyterlab-server-proxy",
--- a/packages/application-extension/package.json
+++ b/packages/application-extension/package.json
@@ -49,6 +49,7 @@
     "@jupyterlab/ui-components": "^3.1.0-alpha.11",
     "@jupyterlite/ui-components": "^0.1.0-alpha.4",
     "@lumino/algorithm": "^1.6.0",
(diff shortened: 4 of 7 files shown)


packages/application-extension/src/index.ts
```diff
@@ -15,20 +15,19 @@ import {
...
 
-const YJS_WEBSOCKET_URL = 'wss://demos.yjs.dev';
+import { WebrtcProvider } from 'y-webrtc';
 
-class WebSocketProvider extends WebsocketProvider implements IDocumentProvider {
+class WebRtcProvider extends WebrtcProvider implements IDocumentProvider {
   constructor(options: IDocumentProviderFactory.IOptions) {
-    super(YJS_WEBSOCKET_URL, options.guid, options.ymodel.ydoc);
+    super(options.guid, options.ymodel.ydoc);
+    this.awareness = options.ymodel.awareness;
```
jtpio, line 27: We could also make the signaling server(s) configurable via the settings / overrides.

jtpio:
A simple "room" support a la Whereby. So folks can share links that look like the following:

https://jupyterlite.readthedocs.io/en/latest/_static/lab/?room=the-coolest-room

image

jtpio:
RTC between lab and retro:

https://user-images.githubusercontent.com/591645/120047618-5520ef00-c015-11eb-9c4d-7dc57f3ecc00.mp4

bollwyvl:
As mentioned on #207: it looks like there is some new behavior on the yjs connector that needs to be accounted for.

> - implement setPath in packages/application-extension/src/index.tsx
>   - didn't have time to go run this one down, yet
>   - consider deferring to #109?

Not sure which of these we'd be most likely to land first...

jtpio:
> Not sure which of these we'd be most likely to land first...

This PR still need a couple of fixes (sync when opening the documents), so probably we can get #207 in first and then rebase here.

jtpio:
From https://readthedocs.org/projects/jupyterlite/builds/14201215/:

> Command killed due to excessive memory consumption

Maybe a sign that we are starting to reach the limits? Will restart the build to double check.

jtpio:
Also adding some docs about RTC: https://jupyterlite--109.org.readthedocs.build/en/109/rtc/index.html

jtpio:
Pushed a couple of improvements to fix the issue with duplicated content when a document is opened.

Now hitting that one when creating a new notebook that is already opened and edited by a peer (they need to choose Overwrite):

image

But it looks like it's fine when opening a existing example notebook, so maybe something to check in the jupyterlite contents manager.

jtpio:
> so probably we can get #207 in first and then rebase here.

Since RTC is opt-in and also requires setting the `room` query string parameter, maybe we can actually merge this one first.

jtpio:
@bollwyvl would you like to give it a try on the RTD demo? 

https://jupyterlite--109.org.readthedocs.build/en/109/_static/lab/index.html?room=the-coolest-room

Thinking we could ship as is if it works ok, and we can address the contents manager issue in follow-up PRs.

jtpio:
> we can address the contents manager issue

For that we might want to start using `y-indexeddb` as an alternative (configurable) backend for the contents manager: https://github.com/yjs/y-indexeddb

bollwyvl:
> Command killed due to excessive memory consumption

Yeah, `mamba` let us sneak by over the initial hump, but webpack will increasingly destroy us.

Some band-aids until we really go after #118 (_almost_ started trying to make some part of core federated, but worked on #238, because users can use it)

- set the `--max-old-space-size` intentionally low, turn on all the caching, and _hope_ that it can re-use cache results from failed runs to get there... eventually
- split up some more parts of each app's build... perhaps the `server` and `client` could be separate builds... and heck, maybe `server` should be running in a webworker _anyway_...

> using y-indexeddb as an alternative (configurable) backend for the contents 

sounds like another issue... i don't understand fully the ramifications, but localForage is just _nice_, not _criticial_, as if a user's browser doesn't have indexeddb, they're likely going to have... other problems with this site.

docs/rtc/index.md
```diff
@@ -0,0 +1,47 @@
...
+## Enabling RTC in JupyterLite
+
+To enable RTC in JupyterLite, you need to set the `collaborative` flag in the
+`jupyter-config-data`. This follows the same behavior as in JupyterLab 3.1 where real
+time collaboration is an opt-in feature.
+
+In addition to the `collaborative` flag, end users must specify the `room` query
+parameter in the URL. An example of such URL is as follows:
+
+[https://jupyterlite.readthedocs.io/en/latest/\_static/lab/index.html?room=my-custom-room](https://jupyterlite.readthedocs.io/en/latest/_static/lab/index.html?room=my-custom-room)
```
bollwyvl, line 33: i wouldn't even make this a clickable link... and frankly, we shouldn't even suggest a room name, as we basically _don't_ want people going to The One Room because it will get filled up with spam.

A nice approach might be to offer an RTC generator with an in-page widget... though `StringConcatenate` isn't a thing, and `wxyz` doesn't work with `requirejs`

docs/rtc/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+Users are grouped together in rooms using a combination of:
+
+- the host, for example `jupyterlite.example.com` or `myserver:5000`
+- the name of the room parsed from the query string parameter, for example
+  `my-custom-room`
```
bollwyvl, line 39: I wonder if the _actual_ key that gets used in the code on the public server could at least be obfuscated, e.g. by taking the hash of the `appName`, `appVersion` and the user-provided key. My confidence is _very low_ in any of these things working between different deployments... and see spam.

packages/application-extension/src/index.ts
bollwyvl: Did this happen? I think we should add this configuration sprocket before merging.

bollwyvl:
It indeed works (between browsers) on RTD, quite magical :sparkles: 

I feel like the presence indicator on the cursors `b1` feels _waaay_ better... and we'll want to make the list of "anonymous" names configurable, as people get _real_ attached to those.

Overall (and i've raised this in core, too, somewhere) I feel like we need _something_ that says, _Hey buddy, you're broadcasting to everyone with this same room, and there's 800 people listening to your `Untitled.ipynb`, so don't put anything stupid in there_. 

Some ideas (not mutually exclusive):
- overlays
  - i despise the dialogs (here, lemme take away your agency), but feel like having an opt-out "THIS IS SHARED" dialog might be appropriate here may be appropriate
  - we could use toast
- main menu
  - top-right, with a distinctive icon, and a peer count if we can get it
- file manager
  - change the file manager icon to be "shared folder" or "drive upload folder", make it orange
  - somehow indicate that _every damn file_ is shared, so can be overwritten at any moment
- launcher
  - add a note to top of launcher, e.g. "Shared with <link>"
  - add "Shared" to all of the launcher category titles
- dock panel
  - each shared tab should get something, but this is crazy lumino vdom
- status bar
  - add a positive (and negative) indicator that you are sharing (with a link to the docs section if not sharing)

So, _technically_ I see no issues to merging, but there are a lot of _social_ and _perceptual_ issues that make this complicated.

bollwyvl:
In punching it a few different ways, i was able to get to a state where _one_ of the windows (the second launched) couldn't make a new file. After the first started a new file, then the second window could make a new file.

bollwyvl:
editing `bar.vl.json` and seeing it update the plot on the other screen is very satisfying.

docs/rtc/index.md
jtpio: Good point. We can still show to specify the `room` parameter, but it could be something else or just a random string.

> A nice approach might be to offer an RTC link generator with an in-page widget

Yes that would make it much more convenient to share a link.

docs/rtc/index.md
jtpio: Yes there is still the risk of folks "seeing" each other. So having some kind of seeds per deployment would greatly reduce the collisions.

jtpio:
> sounds like another issue... i don't understand fully the ramifications, but localForage is just _nice_, not _criticial_, as if a user's browser doesn't have indexeddb, they're likely going to have... other problems with this site.

Definitely something else for later. `localForage` is really nice indeed, but if we start seeing more issues with decentralized content, if one day we enable rtc by default, then maybe indexeddb backed by Yjs could be interesting.

docs/rtc/index.md
bollwyvl: An in-app one would be clutch, too, but could be a federated extension (let's not make webpack any worse until we make it better).

In this _specific_ instance, I mean a link generator _on the docs page_, akin to the binder one, where we can hoist all of the bits and bobs.

jtpio:
> Some ideas (not mutually exclusive):

Nice set of ideas :+1: We can definitely track them in separate issues, and many of these might actually be about reusing existing plugins or extensions made for "normal" RTC.

> In punching it a few different ways, i was able to get to a state where one of the windows (the second launched) couldn't make a new file. After the first started a new file, then the second window could make a new file.

Right I've also seen these. Not sure how frequent they happen but probably something to do the with the `requestInitialContent` logic.

bollwyvl:
> enable rtc by default

let's not, like ever, until the sharing feature upstream isn't _entirely_ opaque to the user. Like, a different IDrive. with lots of indicators of who can see it. And identity.

jtpio:
> let's not, like ever, until the sharing feature upstream isn't _entirely_ opaque to the user. Like, a different IDrive. with lots of indicators of who can see it. And identity.

Yeah this will also be definitely relevant in lab and hub.

bollwyvl (approved):
LGTM, we can figure other stuff out in follow-ons.

Issue #110, opened 2021-05-29
Support creating interactive viz and data apps


MarcSkovMadsen:
I see jupyterlite as a game changer. Thanks so much for taking it so far.

My interest is in interactive data and ML apps. See awesome-panel.org and awesome-streamlit.org. I would like to be able to do interactive, exploratory data analysis and create data apps in the browser for the browser using jupyterlite.

One thing I would need is 1) the possibility to export to a stripped out version of the notebook just like Voila does it 2) some kind of widgets that can communicate and update plots and similar objects. Maybe its the [comms](https://github.com/jtpio/jupyterlite/issues/18) and ipywidgets that are needed. Maybe not - we are already in the browser. So maybe something else would be easy to do.

If creating interactive viz and data apps in any is possible or could be supported I would appreciate anything that can move this forward. Like ideas or examples.

lrowe:
To the same end, it would be nice to support [altair](https://altair-viz.github.io). I'm able to install altair with the following, but it does not display the chart:

```python
import micropip
await micropip.install('Jinja2')
micropip.PACKAGE_MANAGER.builtin_packages['jinja2'] = micropip.PACKAGE_MANAGER.builtin_packages['Jinja2']
await micropip.install("https://files.pythonhosted.org/packages/77/de/47e35a97b2b05c2fadbec67d44cfcdcd09b8086951b331d82de90d2912da/jsonschema-2.6.0-py2.py3-none-any.whl")
await micropip.install("altair")
import sys
sys.setrecursionlimit(255)

import altair
import pandas
source = pandas.DataFrame({
    'a': ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I'],
    'b': [28, 55, 43, 91, 81, 53, 19, 87, 52]
})
altair.Chart(source).mark_bar().encode(
    x='a',
    y='b'
)
```

I tried installing [ipyvega](https://github.com/vega/ipyvega) with `await micropip.install('vega')` but it didn't help. Perhaps the javascript [@jupyterlab/vega5-extension](https://www.npmjs.com/package/@jupyterlab/vega5-extension) needs to be installed?

jtpio:
Thanks @MarcSkovMadsen and @lrowe.

Support for `ipywidgets` (https://github.com/jtpio/jupyterlite/issues/18) should already help getting more interactive visualization libraries to work in JupyterLite. In theory most should (almost) work out of the box when the Comms are implemented.

[...]

PR #113, opened 2021-05-30
Add vega extension, basic support for `altair`

jtpio:
## References

Incremental improvement for #110 

## Code changes

Add the `vega5-extension` to the main application bundle [like in JupyterLab](https://github.com/jupyterlab/jupyterlab/blob/a230896fc5efcb916b4542b3457701ade7c5d805/dev_mode/package.json#L251)

### TODO

- [x] Add `@jupyterlab-vega5-extension`
- [x] Add example vega file to the demo site examples
- [x] Add to retro
- [x] Add `altair` example notebook

## User-facing changes

image

Basic support for `altair`:

https://user-images.githubusercontent.com/591645/120116837-545d9980-c18a-11eb-9097-b3eac5303fb9.mp4

None


bollwyvl:
Awesome, vega is a great first step for viz stuff.

While writing vega by hand isn't super fun, would it be worth having the magic MIME string(s) encoded within `Vega` and/or `VegaLite` display types in pyolite?

jtpio:
> While writing vega by hand isn't super fun, would it be worth having the magic MIME string(s) encoded within Vega and/or VegaLite display types in pyolite?

I think so yes. Ideally we can get Altair to work without too many patches.

jtpio:
Example using @lrowe code snippet from: https://github.com/jtpio/jupyterlite/issues/110#issuecomment-850916083:

https://user-images.githubusercontent.com/591645/120113427-ba422500-c17a-11eb-9e5a-8ca928f301b8.mp4

jtpio:
Users still need the bootstrap code mentioned in https://github.com/jtpio/jupyterlite/issues/110#issuecomment-850916083:

```python
# code snippet below from @lrowe: https://github.com/jtpio/jupyterlite/issues/110#issuecomment-850916083
import micropip
# Work around https://github.com/pyodide/pyodide/issues/1614 which is now fixed in pyodide
await micropip.install('Jinja2')
micropip.PACKAGE_MANAGER.builtin_packages['jinja2'] = micropip.PACKAGE_MANAGER.builtin_packages['Jinja2']
# Last version of jsonschema before it added the pyrsistent dependency (native code, no wheel)
await micropip.install("https://files.pythonhosted.org/packages/77/de/47e35a97b2b05c2fadbec67d44cfcdcd09b8086951b331d82de90d2912da/jsonschema-2.6.0-py2.py3-none-any.whl")
await micropip.install("altair")
import sys
sys.setrecursionlimit(255)
```

Maybe that's something that could be added as a "helper patch" in pyolite to make it more convenient for now.

lrowe:
> Maybe that's something that could be added as a "helper patch" in pyolite to make it more convenient for now.

Breaking down the steps of my bootstrap code:

> ```python
> # Work around https://github.com/pyodide/pyodide/issues/1614 which is now fixed in pyodide
> await micropip.install('Jinja2')
> micropip.PACKAGE_MANAGER.builtin_packages['jinja2'] = micropip.PACKAGE_MANAGER.builtin_packages['Jinja2']
> ```

Fix is being actively worked on here: https://github.com/pyodide/pyodide/pull/1615

> ```python
> # Last version of jsonschema before it added the pyrsistent dependency (native code, no wheel)
> await micropip.install("https://files.pythonhosted.org/packages/77/de/47e35a97b2b05c2fadbec67d44cfcdcd09b8086951b331d82de90d2912da/jsonschema-2.6.0-py2.py3-none-any.whl")
> ```

I believe the right way to fix this is to build pyrsistent as a pyodide package: https://pyodide.org/en/stable/development/new-packages.html

Was going to look at this but haven't managed to get pyodide to build on my ARM Mac.

> ```python
> import sys
> sys.setrecursionlimit(255)
> ```

This is to avoid running into the recursion limit while rendering with altair. `sys.getrecursionlimit()` returns 127 in jupyterlite on Chrome for me which seems very low. At https://pyodide.org/en/stable/console.html it returns 220 on Chrome and 912 on Safari (jupyterlite doesn't seem to work properly on Safari for me) and 1000 on my regular python 3.9.5 interpreter.

Is the lower recursion limit set by jupyterlite or calculated somehow by pyodide? Seems reasonable to increase it a bit.

bollwyvl:
> Is the lower recursion limit set by jupyterlite or calculated somehow by pyodide?

The recursion limit stuff is derived from limitations of the WASM stack, as pyodide frames are _waaaay_ heavier than JS (much less WASM) ones... it's a bit of a heuristic, derived from how the upstream works.

> Last version of jsonschema before it added the pyrsistent dependency (native code, no wheel)

Yeah, this is going to be a killer in a bunch of places.

jtpio:
Thanks @lrowe for the extra context :+1: 

> Is the lower recursion limit set by jupyterlite or calculated somehow by pyodide? Seems reasonable to increase it a bit.

Looks like `pyodide` has some logic to choose a value for the recursion limit here: https://github.com/pyodide/pyodide/blob/2e0475227255166269d910f260d138274c02d5fc/src/js/pyodide.js#L118-L143

JupyterLite doesn't explicitly set the limit. Although we could probably place the `sys.setrecursionlimit(255)` in https://github.com/jtpio/jupyterlite/blob/main/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py.

lrowe:
I just plucked the 255 figure out of the air. It seems that 170 is sufficient for the altair chart I tested. Anything much above that and I get a pyodide fatal error in the JS console (`RangeError: Maximum call stack size exceeded`) with the following code:

```python
import sys; sys.setrecursionlimit(170)
def test(i=0):
    print(i)
    test(i+1)
test()
```

It's uncomfortably close to the limit either side though, 160 and altair cannot render, 180 and I see pyodide fatal error. Guess the following might be okay for now:

```python
import sys
sys.setrecursionlimit(max(170, sys.getrecursionlimit())
```

jtpio:
> Guess the following might be okay for now

Added as default in https://github.com/jtpio/jupyterlite/pull/113/commits/6bdf4f184de3754ff1392e1861fecabdb0ba9094 for now.

jtpio:
Let's get this in then, we can always improve and iterate in follow-up PRs.

Thanks all!

PR #117, opened 2021-06-01
Add CI job to create GitHub releases and upload assets

### LINKED ISSUE #16: Publish release assets via GitHub releases
During the release process, the static assets can be published as release assets so it's easy for anyone to download them and serve the content of the tarball right away with a simple web server.

jtpio:
## References

Fixes #16 

Add a CI job to automatically create a GitHub release, and upload the content of the tarball as release assets.

This is a first step to make it easier for folks who want to deploy their own JupyterLite.

### TODO:

- [x] Add the `release` job to the existing `build` workflow
- [x] Trigger on `v*` tag
- [x] Create the release via the action with https://github.com/softprops/action-gh-release (https://github.com/actions/upload-release-asset is deprecated)

## Code changes

CI

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,3 +86,19 @@ jobs:
           if-no-files-found: ignore
           path: |
             /tmp/sphinx-*.log
+
+  release:
+    needs: [build]
+    # if: startsWith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: jupyterlite dist ${{ github.run_number }}
+          path: ./dist
+      - name: Extract the app
+        run: |
+          tar -xvf ./dist/*.tgz
+      - name: tmp - list dir
+        run: |
+          tree .
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,8 @@ name: Build
 on:
   push:
     branches: [main]
+    tags:
+      - 'v*.*.*'
   pull_request:
     branches: '*'
 
@@ -89,16 +91,19 @@ jobs:
 
   release:
     needs: [build]
-    # if: startsWith(github.ref, 'refs/tags/v')
+    if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/download-artifact@v2
         with:
           name: jupyterlite dist ${{ github.run_number }}
           path: ./dist
-      - name: Extract the app
-        run: |
-          tar -xvf ./dist/*.tgz
-      - name: tmp - list dir
-        run: |
-          tree .
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          prerelease: true
+          files: |
+            dist/jupyterlite-*.tgz
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,6 +104,6 @@ jobs:
           draft: true
           prerelease: true
           files: |
-            dist/jupyterlite-*.tgz
+            dist/*
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


jtpio:
We could then extend this workflow to publish to PyPI.

jtpio:
Testing:

image

Will update the workflow for real use, and we should be good to go for now.

.github/workflows/build.yml
```diff
@@ -86,3 +88,22 @@ jobs:
...
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: jupyterlite dist ${{ github.run_number }}
+          path: ./dist
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          prerelease: true
```
jtpio, line 105: Keeping `prerelease` for now.

jtpio:
Follow-up:

- Set versions to `0.1.0a0` and `^0.1.0-alpha.0`
- Scripts to bump versions
- Document the process to cut releases

.github/workflows/build.yml
```diff
@@ -3,6 +3,8 @@ name: Build
 on:
   push:
     branches: [main]
+    tags:
+      - 'v*.*.*'
```
bollwyvl, line 7: i don't recall if that's a regex, but the `.` might need escaping, and potentially a `^` up-front. but since they're tags...

.github/workflows/build.yml
bollwyvl: it's a bit of filagree, but CHANGELOG notes are nice... have done this regex-fu vs `---` ... though not sure if we _have_ one...

bollwyvl:
Yeah, i'm generally dubious about third-party actions for pushing Real Stuff... on some other stuff, i've used the `release` action (triggered by the GUI) which then can use normal `gh` CLI stuff (already on the workers) to do GH artifact uploads with the built-in secret. Doesn't solve the pypi thing, but at least that is an "official" action.

But this is all fun and games for a while, and :ship: good.

.github/workflows/build.yml
jtpio: This one comes from https://github.com/softprops/action-gh-release#-limit-releases-to-pushes-to-tags, which hopefully has the correct syntax.

jtpio:
> I've used the `release` action (triggered by the GUI) which then can use normal `gh` CLI stuff (already on the workers) to do GH artifact uploads with the built-in secret

Nice, we could definitely switch to this. Do you have a link to where this is used? (so it can serve as an example).

bollwyvl:
Well, this is not exactly a _nice_ example, but it's how we're doing it for some _truly_ awful builds that are larger than the 2gb artifact limit. Not proud:

https://github.com/gt-coar/gt-coar-lab/blob/master/.github/workflows/ci.yml#L882
https://github.com/gt-coar/gt-coar-lab/blob/master/_scripts/release.py
https://github.com/gt-coar/gt-coar-lab/blob/master/_scripts/upload-github-release-asset.sh

Again, it's a lot of work for probably minor piece of mind, but the huge files seem to be a relatively unique (ab)use of the GHA system. I'm frankly quite disappointed GH divested themselves of responsibility for official actions, which at least would have the advantage of many eyeballs. Guess you (don't) get what you (don't) pay for!

jtpio:
Right, that looks a bit more complicated.

But we could consider it later (and track it in an issue). It would indeed be better to not depend on third-party actions for things that consume tokens.

.github/workflows/build.yml
```diff
@@ -86,3 +88,22 @@ jobs:
...
+        with:
+          name: jupyterlite dist ${{ github.run_number }}
+          path: ./dist
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          prerelease: true
+          files: |
+            dist/jupyterlite-*.tgz
```
bollwyvl, line 107: any particular reason to not ship _evrything_, including the SHA256?

jtpio: No real reason . Let's change it to `dist/*` :+1:

jtpio:
```suggestion
            dist/*
```

jtpio:
First GitHub Release: https://github.com/jtpio/jupyterlite/releases/tag/v0.1.0a0

Looks like it's missing the sha file though.

.github/workflows/build.yml
bollwyvl: hm... i don't see why it would get missed... no big, but worth exploring. might be because it doesn't have an extension? but its definitely the convention. at worst, we name it explicitly, as releases have no folders: 
```
dist/SHA256SUMS
dist/*
```

jtpio:
> Looks like it's missing the sha file though.

Ah but we don't create it in the workflow, that could explain why: https://github.com/jtpio/jupyterlite/blob/main/.github/workflows/build.yml

jtpio:
Opened: https://github.com/jtpio/jupyterlite/pull/126

Issue #60, opened 2021-04-22
Custom About Dialog


jtpio:
### Problem

As more people get to try JupyterLite by going to the demo link directly: https://jupyterlite.vercel.app/

We should provide a way for them to reach the repo, for example by swapping the about dialog for a custom one:

![image](https://user-images.githubusercontent.com/591645/115743642-24082b80-a392-11eb-82d1-edacc9e241e5.png)

### Proposed Solution

The about plugin was made more modular upstream in https://github.com/jupyterlab/jupyterlab/pull/10089.

So this will be easier to implement after updating to the 3.1 alpha packages.

jtpio:
The about dialog could also be the same for both lab and retro.

PR #130, opened 2021-06-03
Add ui-components package

### LINKED ISSUE #124: Expose the logo and wordmark
### Problem

The lab and retro interfaces use the default icons and wordmark for upstream applications for now:

image

image

We might want to start using the lite assets instead.

Also third-party extensions might want to be able to reuse the logo.

### Proposed Solution

- Create a `@jupyterlite/ui-components` package
- Expose `jupyterLiteLogo` and `jupyterLiteWordmark` as `LabIcon`
- Use `jupyterLiteLogo` in the top left corner 

image

### Additional context

Logo and wordmark discussed in #86

jtpio:
## References

Fixes #124 

## Code changes

Add a new `ui-components` to expose the icon and wordmark.

### TODO

- [x] Bootstrap the package
- [x] Replace the JupyterLab main logo plugin
- [x] Replace the RetroLab main logo plugin
- [x] Add `doit` task to copy the icons from the source in docs

## User-facing changes

image

image

None


jtpio:
> Add a new `ui-components` to expose the icon and wordmark.

They could then be used in #60 to implement the about dialog.

bollwyvl:
Very cool!

So it's kind of a gross dependency, but perhaps instead of copying them, we should `svgo` the assets as they are shipped? Keeping them in full editable form in inkscape in the docs feels good, but as release assets, they are pretty bloated.

jtpio:
> So it's kind of a gross dependency, but perhaps instead of copying them, we should svgo the assets as they are shipped?

Sounds like it would be fine to optimize them with `svgo` yes. Also to automatically remove leading `<?xml>` tag which might cause some issues when using it with react in the about dialog.

bollwyvl:
> using it with react in the about dialog.

Gross. Sounds like a react problem.

bollwyvl:
Another worry is all that other junk i left in the wordmark... if nobody's sad about the colors, etc. we can just stick with what we got, and dump all the other stuff. Though I did like the canary yellow.

bollwyvl:
Though the canary doesn't have a symbolic name in the `jp-icon-colorN` names...

bollwyvl:
Sorry, shoulda said i'll do some tweaks on here. While the icon is good (4kb) the wordmark is still 140kb... might need some more finesse to get it down. I'll go ahead an update with the yellow...

bollwyvl:
```
wordmark.svg:
Done in 102 ms!
35.574 KiB - 60.9% = 13.912 KiB
```

bollwyvl:
Screenshot from 2021-06-03 20-41-54

bollwyvl:
i don't like the outline on dark, trying the semantic themes...

bollwyvl:
Much better with `jp-icon3` (lower is still too shouty):

Screenshot from 2021-06-03 21-57-49

bollwyvl:
No further changes planned, but happy for review!

jtpio:
Thanks!

Issue #118, opened 2021-06-01
Deduplicate the lab and retro assets


jtpio:
### Problem

Tracking the following comment from https://github.com/jtpio/jupyterlite/issues/16#issuecomment-821980005 in its own issue:

> Further while talking about release assets... Right now, the release is
> about 14mb, 64 uncompressed. A huge proportion of that is the duplication
> of large assets in lab/build and classic/build. It's probably worth
> exploring how to get as much as possible to end up in ../static, relative
> to each of them.

### Proposed Solution

- Put all assets in a common `/static` folder
- Rename `index.js`, `style.js` and `bundle.js` to `index-lab.js`, `index-retro.js` ... so they don't collide
- Update the references to `./build`, `bundle.js` and others

bollwyvl:
Thanks for starting this. Some additional wrinkles on this: 

- might block MVP
  - when this gets tackled, we gotta get cache-busting in there
    - the top-level can be handled by `index.js?v=<date>`
    - but subsequent requests below it need to "just work" on redeployment update, or there's gonna be a lot of _really_ confused users
- should not block MVP:
  - re #41, it would be ideal if there was a solid way to extract _which_ `static` are needed for which app, such that `jupyter lite build --app @jupyterlite/retro-lite ./my-site` only ships the assets needed, while adding another `--app` would only increase the size by the delta of missing packages, _without_ having to know `index-this` from `index-that`
  - have _a_ path (if "fork it and..." in the docs, initially, so be it) to shipping (only) that novel _third_ application (e.g. `CustomLabLite`). This is a case where, unless we unbundle _everything_ into federated modules, someone will still need _all_ the nodejs machinery, no real way around it at present

bollwyvl:
So i got a bit hung up on how we would refactor the various scripts to not make use of `data`... and the interplay between `bootstrap.js` and the sharing piece, as they will both actually be on the page at the same time, while multiple _apps_ will _not_ be on the page at the same time... but might share worker code, etc. So I'll probably take a look at some other stuff before re-engaging on this, in case someone else has a more concrete plan forward :pray: !

As we think about #104, I guess there's a latent refactor in the baseline of `@jupyterlab/builder` (though maybe we land it here first)... when it gets to the [good stuff](https://github.com/jupyterlab/jupyterlab/blob/master/examples/federated/core_package/index.template.js#L105), might we have a pattern that supported 
```
serverliteExtensions -> package.json#/jupyterlite/???
```
> or, more generally 
[...]

PR #137, opened 2021-06-04
Add offline support

dmonad:
This PR adds a service worker to the lab folder. It will intercept all requests to the server and store the results in a cache. The next time you load the website, the content will be loaded from the cache instead from the server (while updating the cache in the background). If there is no internet access, the `/lab` address should still work.

Note that this might make it harder to debug because it will serve stale content by default. When debugging the application (e.g. in dev_mode on localhost) you should enable "bypass for network" in the debugging application panel. 

Maybe we could disable service worker for localhost alltogether.


bollwyvl:
:heart_eyes_cat:  This is a _huge_ step forward, and the reload times are _fantastic_ with e.g. pyodide. Works as advertised when turning off wifi! We'd probably want to enable even more goodies, e.g. have a pyodide kernel shared between multiple tabs.

But indeed, the various deployment gotchas are very real, so it needs to be easy to turn off at build time for someone who knows they will be deploying someplace not so fun, e.g. `jupyter-lite.json#/jupyter-config-data/disableServiceWorker`.

Also, I think before landing this, we'd also want to land #118 to get deduplicated, cache-busting assets, so at least first-party stuff doesn't have surprising stale experiences on the docs site, especially since we're tracking upstream alphas now.

As we want this to work on every page, and likely know _whether_ we are in a service worker, perhaps we move all this to the `config-utils.js`, so it gets used on both the full lab and the various retro pages, and set `useServiceWorker` dynamically, if not explicitly opted-out... if we _can_ use it, we probably always _should_!

dmonad:
Yep, this all makes sense. Let's wait a bit until we merge this. I'll try to keep this up to date. I think it would be nice to also add a Manifest so that users can install this as a web-app. Maybe the service worker should only be active when this app is served as a web-app. This would be another method to circumvent some of the issues.

Btw, the documentation generated a working example of this:

https://jupyterlite--137.org.readthedocs.build/en/137/_static/lab/index.html

bollwyvl:
> documentation generated a working example

Oh yeah, tried it out immediately, did the ceremonial network-turn-off and everything! This _really_ makes slightly-larger-than-trivial compute reasonable for a documentation site.

dmonad:
I added a manifest so you can install this as a web app now :)

Screenshot from 2021-06-04 16-22-33

bollwyvl:
Very cool. We'll definitely want to hoist the icons, label, theme colors, etc. to something a user "build" (e.g. copy and add <5 json files or a notebook) can configure in #41.

bollwyvl:
So with #173, we've got some more structure in place for configuring how a site builds, etc. I left a placeholder for the serviceworker stuff, but wasn't sure how to proceed. It would be interesting to get `main` merged into this so we can start looking at options for moving forward.

bollwyvl:
Another angle: the [webpack docs](https://webpack.js.org/guides/progressive-web-application/) point out [workbox](https://github.com/GoogleChrome/workbox) which seems to make some of the stuff a little more manageable over time. Not sure how this would play with our desire to be able to tweak things _after_ the webpack build, but might still be interesting.

martinRenou:
I'd like to take over that PR and rebase it if it's fine with everyone.

I've been looking for the past working days at service workers and how to make use of their advantages from the Python kernel.

I'm not only interested in the caching logic they bring (for offline support etc), I'm also interested in the Python kernel being able to synchronously access data from the virtual file system (local storage). The Python kernel running in a web worker, it can make blocking HTTP requests that are intercepted by the service worker, and the service worker can answer with whatever data it finds. We could monkey patch the `open` global function with something that does exactly that.

joemarshall:
@martinRenou  I was just thinking about this. I think the right approach for flexibility would be to register a very simple service worker which just allows the registration of URL handlers. So one could make a cache handler extension, a sleep extension, import extension.

It would need some way to identify which main browser thread owned webworker clients, I'm not sure if that can be done through fetch request detection in the service worker or if it needs a wrapper for webworker.

joemarshall:
I just checked, and it is trivial to associate client ids of web-workers to the clientid of the main window - 
1) In chrome - all fetches seem to come with the main window clientid
2) In firefox, when you first fetch the webworker, the fetch has 'resultingClientId for the worker, which is a new client id. worker requests come with that client ID and the same url. So if you're in a web-worker, the client id of the main window is trivial to get.

So a basic URL handler API for pyodide or similar stuff where you essentially want synchronous calls to async javascript things (files, sleep etc.) would look like this:
1) Serviceworker gets a POST request to a special URL with some content (as JSON or something)
2) It makes this into a promise and postmessages it to the correct mainwindow (and returns a new unresolved promise to the fetch request)
3) The main app converts that to a jupyter message and sends it wherever it needs to go. 
4) The handler extension makes a response and sends that back to the main app, which posts it back to the serviceworker, which then resolves the promise made in step 2.
5) Tada, XML httprequest in webworker finishes, we're all good.

I think caching should maybe be handled in the main serviceworker for performance reasons. But the rest of the serviceworker should stay absolutely barebones - it just takes requests and turns them into jupyter messages if they are a special PUT request.

martinRenou:
Thanks for your comments @joemarshall !

My comments about monkey-patching the `open` global function in Python are invalidated now, so please discard them.

I'm exploring implementing a custom FileSystem (in the emscripten sense) that we mount for Python to use, and exposes the files of the current JupyterLab drive. The work is done in https://github.com/jupyterlite/jupyterlite/pull/655 (not all my code is pushed yet).
The problem is emscripten file systems must be synchronous (there is currently some work in emscripten to make those APIs async but it's not finished/released yet, see discussion in the PR mentioned above).

So I think using a service worker the way you're describing above is needed, in order to turn async file/directory fetches into synchronous tasks.

> Serviceworker gets a POST request to a special URL with some content (as JSON or something)
It makes this into a promise and postmessages it to the correct mainwindow (and returns a new unresolved promise to the fetch request)
The main app converts that to a jupyter message and sends it wherever it needs to go.
The handler extension makes a response and sends that back to the main app, which posts it back to the serviceworker, which then resolves the promise made in step 2.
Tada, XML httprequest in webworker finishes, we're all good.

This makes perfect sense. I was reading this morning about [`BroadcastChannel`](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel) which I think will be perfect for that:
- it's bidirectional
- you can have multiple channels (one for `input`, one for `sleep`, one for the file system etc)

joemarshall:
I think normal postmessage makes more sense, because e.g. if you call input, you only want to read from things in the same window the kernel was launched from.

Instead of channels the messages could just have a URL included, so that they can be routed to server or client plugins by JupyterLiteServer same way other messages are routed (I guess messages start off at the front-end because presumably that is what has access to the window object) I think that makes more sense than adding another form of communication to the whole thing.
________________________________

jtpio:
Thanks @dmonad for initially starting this PR!

The commits have been included in https://github.com/jupyterlite/jupyterlite/pull/686.

Issue #104, opened 2021-05-22
Support federated serverlite extensions


bollwyvl:
### Problem

At present, a JupyterLite pretty much must include _all_ of the kernels, exactly the contents/settings extensions provided etc. It's also basically not possible to add new serverlite extensions without forking or going through the PR process here.

### Proposed Solution

At extension build time adopt either:
- _exactly_ the `jupyter labextension build` toolchain, 
  - we might be able to get by with "requiring" (as in, might build, but not run without) `package.json#/jupyterlab/sharedPackage` pointers to `jupyterlite/server`
- an extension-by-subclass of that machinery for `jupyter serverlitextension build` (that is very long)

At configuration time, either:
- use exactly `federated_extensions`
- use a very similar `federated_serverlite_extensions`

At runtime:
- use the config
- actually read `package.json`

### Additional context

As a stopgap, it was proposed elsewhere to offer a way to specify a mock server response via OpenAPI to provide enough server information to trick an existing, useful labextension into working, e.g. `jupyterlab-videochat`.

bollwyvl:
> from #348: We should probably start looking into #104.

Agreed, (unfortunately, as this looks very cool). 

More broadly, bespoke kernels will end up being the "long pole" for size, build time, and complexity: for example, pyodide is already quite complex, and will get _more_ complex on #310.

For users, this would play out like:
- each kernel build/config chain would end up being a separately-installable `pip` package which can bring its "core" distribution, extra `build` steps and dependencies, etc.
  - each (e.g , `jupyterlite-lualite`) would depend on a new `jupyterlite-core` package
- the `jupyterlite` metapackage would then depend on the "grandparented" packages (`jupyterlite-jslite`, `jupyterlite-ipyolite`)

PR #138, opened 2021-06-06
add chunkHashname with contenthash in webpack

bollwyvl:
## References

- a small part of #118
- probably needed for #137 to not be too crazy

## Code changes

- tweaks some webpack settings 

## User-facing changes

- should result in more reliable outputs resulting in fewer _odd_ cache situations

- n/a

## future work

- the `bundle.js` is probably still a problem?
- it'll still take some doing to handle the multiple entrypoints correctly
 - when we get it squared away, we can also explore the `runtimeChunk` options to get better cache behavior
- we might need some more movement towards #104
- the _build-time_ `cache`, which I _haven't_ turned on is _pretty awesome_

```
webpack 5.37.1 compiled successfully in 1223 ms
lerna success run Ran npm script 'build:prod' in 1 package in 58.5s:
```
And then, subsequently:
```
webpack 5.37.1 compiled successfully in 423 ms
lerna success run Ran npm script 'build:prod' in 1 package in 6.5s:
```

### DIFF
--- a/app/webpack.config.js
+++ b/app/webpack.config.js
@@ -225,6 +225,9 @@ module.exports = [
         }
       ]
     },
+    optimization: {
+      moduleIds: 'deterministic'
+    },
     plugins: [
       new webpack.DefinePlugin({
         // Needed for Blueprint. See https://github.com/palantir/blueprint/issues/4393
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

Issue #116, opened 2021-05-31
Rename the Pyolite kernel


jtpio:
Looks like `pyolite` is already a thing on PyPI: https://pypi.org/project/pyolite/

To avoid confusion we can pick a different name for the default Python kernel we ship in: https://github.com/jtpio/jupyterlite/tree/main/packages/pyolite-kernel

bollwyvl:
> @jpio on #129 
>> So now might be a good time to start having a minimal IPython (mostly IPython.display) that would run in jupyterlite.

Indeed! But answering the API mail _exactly_ on #18 will need this anyway for e.g. [`bokeh`](https://github.com/bokeh/bokeh/blob/232c5bef5b4433b1b0ed3c23ca1ca2fd6deedfff/bokeh/io/notebook.py#L337) or freaky-deaky [wrapper kernels](https://github.com/robots-from-jupyter/robotkernel/blob/a2567489fe7d766d72a0982fa398743d89905ad4/src/robotkernel/display.py#L3) that subclass `KernelBase`.

So, offspring-of-pyolite probably needs to not only offer an importable `IPython` and `ipykernel`, but also _distributions_ named `ipython` and `ipykernel` that have version-equivalence with the "real ones" on PyPI.

Further, as the pyodide one is, for the time being, _by far_ the heaviest kernel, it ideally needs to be distributed as its _own_ package(s), that _is/are_ on PyPI, a la #104. So the notional _real_ `{sys.prefix}/share/jupyter` of the building environment after installing these package(s) might be something like:

```
share/jupyter/lite/pyodide/
  ipython-7.24.0-py3-none-any.whl
  ipykernel-5.5.5-py3-none-any.whl
```

...and build up a compatible index similar to how we're doing contents for now.

Starting to get pretty fractal at this point, as these would _also_ be `.whl`s that someone could `micropip.install()`.

cc @carreau

Carreau:
Happy to do changes needed in IPython and reduce the size. I doubt I'll have time myself.

bollwyvl:
@careau No worries! I don't think we'll need any changes upstream, just alerting you that there might be another `whl`-in-a-`whl` of IPython on the horizon that only works in the browser. all the zmq stuff, etc. is basically a non-starter, and we're going to have to fake it. But luckily no windows problems!

jtpio:
As a start we probably don't need the whole `ipython`, but mostly its display the module? And being able to configure or override the `publish_display_data`.

[...]

Issue #18, opened 2021-04-08
Comms support


martinRenou:
How easy do you think it would be to add comms support?

martinRenou:
If we can implement comms, ipywidgets **might** work for free?

jtpio:
I think @ibdafna was looking into it at some point and it should be feasible in theory.

It's possible that the main issue for now is the dependency on `ipykernel`: https://github.com/jupyter-widgets/ipywidgets/blob/e0d41f6f02324596a282bc9e4650fd7ba63c0004/setup.cfg#L38

So having a lighter ipywidgets which could be a fork, or maybe by splitting parts of ipywidgets into several packages might work.

martinRenou:
Indeed the ipykernel dependency might be annoying. The ipykernel dependency is because it provides the Python Comm implementation.

We had this issue in xeus-python as well, and we had to monkey patch ipykernel to inject our own Comm class. This might be needed in this pyodide case as well, although it is not a nice solution. I started a discussion here: https://github.com/jupyter-xeus/xeus-python/issues/342 to hopefully find a solution to have a cleaner way for the "current" kernel to expose the comms implementation. Maybe using entry-points might be the key.

bollwyvl:
One thing to consider: while we don't want to bite off re-implementing the entire widget stack, comms will be useful in many kernels. For example, we've considered replacing/augmenting bespoke websockets with comms for delivering LSP https://github.com/krassowski/jupyterlab-lsp/pull/278.

So: having this at the base js kernel level, and offering something that spans the inevitable webworker divide, with an initial in-worker python implementation, would be an interesting example of getting more features into more kernels.

It seems like basing it on the [ipykernel CommManager][https://github.com/ipython/ipykernel/blob/master/ipykernel/comm/manager.py#L17) isn't the worst idea.

psychemedia:
Re: ipywidgets, I just started trying to go through the dependencies required to install it and get stucj with a missing core package (`termios`):

```
await micropip.install('https://files.pythonhosted.org/packages/f6/7d/3ecb0ebd0ce8dcdfa7bd47ab85c1d4a521e6770ef283d0824f5804994dfe/traitlets-5.0.5-py3-none-any.whl')
await micropip.install('https://files.pythonhosted.org/packages/c0/34/abf879c9c8a282b2d06c880f3791ce91b25de1912bdcfa4a209afebe12f3/ipython-7.24.1-py3-none-any.whl')
[...]

PR #141, opened 2021-06-09
[do not merge] Naive JS Comms

bollwyvl:
So this... kinda re-implements ipykernel's CommManager, and slaps it on all the kernels, plus adds some widgets.

I am still getting my feet under me as to the local approaches vs previous endeavors, and will admit there was _even more_ `as any` going on, but it's kinda into shape, from a soundness perspective :blush:.

## References

- [x] a potential approach to (part of) #18

## Code changes

- [x] adds a `DefaultCommManager` to all kernels
  - [ ] can be overridden by subclasses
- [x] adds a barebones JS widget implementation
  - this approach is not going to scale, but need to show a few things!
    - [x] `HasTraits`
    - [x] `Widget`
    - [x] `FloatSlider`
    - [x] `widget.observe`
    - [ ] `link`
    - [ ] `dlink`
    - [ ] `jslink`
    - [ ] `jsdlink`
    - [ ] `Textarea` 
    - [ ] `Select` 
    - [ ] `Upload` (for buffers?) 
    - [ ] `interact`?
- [x] in the JS Kernel, adds `self.kernel`, and `widgets`, pretty much defeating the iframe isolation :scream: 

## User-facing changes

Users of the docs site would be able to use at least a couple `FloatSlider`s in JS :wink: :

Screenshot from 2021-06-09 12-35-00

I did originally dogfood the whole thing _in_ a JupyterLite JS notebook... it was kind of hair-raising, and we gotta work on our persistence game. 

- Haven't thought about it much... if someone already wrote a comm implementation and is holding it back, it would no doubt be incompatible :blush: 

## Future Work

- do a data-driven lift of the upstream as a file (e.g. JSON Schema) and generate classes
  - types are fun and all, and probably table stakes for testing, 
    - but we want concrete (generated) js implementations that will work with (eventual) completion in bare JS
- proxy this api _again_ back into `daughter-of-pyolite` #116 across the `WebWorker` divide
  - create `ipykernel`, `ipython`, `ipywidget` shims
- make the CommManager more pluggable 
  - alternate offer contracts than OneCommManagerPerKernel
    - `localForage` 
    - Yjs-backed


bollwyvl:
Gross:
```
/home/docs/checkouts/readthedocs.org/user_builds/jupyterlite/checkouts/141/docs/api/ts/interfaces/kernel_src.icommmanager.itarget.md:10: WARNING: Non-consecutive header level increase; 2 to 4 [myst.header]
```

jtpio:
Really neat, thanks for starting this!

jtpio:
Playing around on the RTD preview:

https://user-images.githubusercontent.com/591645/121414506-fe95a800-c966-11eb-9095-463504ee1e92.mp4

martinRenou:
You're amazing!

martinRenou:
@jtpio can you share a link to the RTD preview?

jtpio:
> can you share a link to the RTD preview?

It's in the checks below: 

image

app/lab/package.json
```diff
@@ -54,6 +54,7 @@
     "@jupyterlab/mathjax2-extension": "^3.1.0-alpha.11",
     "@jupyterlab/notebook": "^3.1.0-alpha.11",
     "@jupyterlab/notebook-extension": "^3.1.0-alpha.11",
+    "@jupyterlab/outputarea": "^3.1.0-alpha.11",
```
jtpio, line 57: This one should be fixed on `main` now.

martinRenou:
We should find a way to plug your comm implementation to ipywidgets. Do you think monkey patching ipykernel could be enough? something like `import sys; sys.modules['ipykernel.comm'] = your_comm_module` should do the trick, this [is what we do in xeus-python and it works like a charm](https://github.com/jupyter-xeus/xeus-python/blob/master/src/xinterpreter.cpp#L75). I would agree monkey patching is an ugly thing to do. But unless ipywidgets is changed, there was no way to do otherwise.

The only problem is ipywidgets bringing ipykernel, we should install ipywidgets ignoring dependencies. I don't know if conda can do that.

jtpio:
> Do you think monkey patching ipykernel could be enough?

Maybe a shim will be enough? That would mean creating a pure python package for `ipykernel` that would be installed in the pyodide kernel by default. And probably the same for `ipython`. We could also add first-party support for `ipywidgets` so users don't need to `micropip.install('ipywidgets')`.

bollwyvl:
Yeah, I dunno just yet about how to make this quack right in pyolite with actual packages... it'll be interesting, of course, but maybe still outside the scope of this PR.

packages/kernel/src/kernel.ts
```diff
@@ -102,6 +135,15 @@ export abstract class BaseKernel implements IKernel {
       case 'history_request':
         await this._historyRequest(msg);
         break;
+      case 'comm_open':
+        await this._comm_manager.comm_open(msg);
```
jtpio, line 139: Wondering whether we should adopt the same pattern as for the other messages, by having methods on the `IKernel` interface? 

https://github.com/jtpio/jupyterlite/blob/97bbe1a0e8ec31108d905c3b7752666a01634851/packages/kernel/src/tokens.ts#L45

For example with `commOpen`, `commMsg` and `commClose`.

This could allow other kernels to completely change the implementation of the comm manager if they want, or just forward messages in the case of the Python kernel reusing the existing comm implementation.

bollwyvl: Yeah, as i mentioned, i'm still figuring out the architecture. the joke was, of course, first try to sneak into how one _might_ get this into pyolite, realizing that was way crazy, and then unwrapping the actual flow of messages through the system to get to _something_ that works.

This would of course need to move into its own package, but no reason not to have the API in better shape...

jtpio: > Now I am wondering if it's feasible and if it makes sense to reuse @bollwyvl 's JS implementation of the comms instead, we would need to expose those comms to the "Python process".

In theory probably yes (or hopefully yes), if the JS implementations follow what ipykernel expects. This could be passed as a custom Python package from JavaScript: https://pyodide.org/en/stable/usage/faq.html#how-do-i-create-custom-python-packages-from-javascript

bollwyvl: Right, I think we win if these two PRs can be harmonized, but prettied up, moved to some new packages....

```
packages/
  kernel/
    kernel.ts        # gets the comm(Open|Msg|Close)Msg handlers
  js-kernel/
    kernel.ts        # has new CommManager({kernel: this})
  py-kernel/src/
    worker.ts        # also needs CommManager in here...
    kernel.ts        # ... injects it here e.g. `import js; (await js.import(${COMM_URL})).CommManager`
    py/.../
      patches.py     # sys.modules hacks, as on #145
                     # ...and the shim to do-async everything from #141 js
  comms/src/
    comm.ts
    comm-manager.ts
    providers/
      memory.ts
      # TODO: localForage.ts
      # TODO: yjs.ts
```

In `patches.py`, there are probably some more pointy bits... At the `pyolite`/`import js` boundary we need it to pretty much be dumb JSON types and `Array`s, but it should be feasible.

bollwyvl: I'll hoist this back to #18

PR #145, opened 2021-06-15
Comms pyolite

martinRenou:
Related to https://github.com/jtpio/jupyterlite/issues/18 Similar to https://github.com/jtpio/jupyterlite/pull/141 but for pyolite and ipywidgets.

This is not yet working.


jtpio:
Thanks, exciting stuff!

jtpio:
To add more example widgets to the demo site on read the docs, we should be able to just add the new dependencies to this file:

https://github.com/jtpio/jupyterlite/blob/e188555bc69a7cbf31fef7e4fd237eec5d55da56/docs/environment.yml#L27

packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
```diff
@@ -15,7 +17,26 @@
...
+ip_mock = IPMock(kernel_instance)
+
+ip = types.ModuleType('IPython')
+ip.get_ipython = lambda: ip_mock
+ip.InteractiveShell = InteractiveShellMock
+
+sys.modules["IPython.display"] = pyolite.display
+sys.modules["IPython"] = ip
+sys.modules["IPython.core.getipython"] = ip
+sys.modules["IPython.core.interactiveshell"] = ip
```
martinRenou, line 42: This needs to be removed. We should make a "real" IPython so that ipython does not get installed anymore

packages/pyolite-kernel/py/pyolite/pyolite/display.py
```diff
@@ -43,6 +46,11 @@ def format_result(result, raw=False):
     return bundle
 
 
+# TODO Implement the clear_output
+def clear_output(*args, **kwargs):
+    pass
```
martinRenou, line 51: If you don't mind, I am leaving this for later. It's needed so that ipywidgets can import it. But might not be straightforward to implement.

jtpio:
We could also add a new example notebook in: https://github.com/jtpio/jupyterlite/tree/main/examples

For example the one from the ipywidgets repo: https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/source/examples/Widget%20List.ipynb

packages/pyolite-kernel/py/pyolite/pyolite/display.py
jtpio: Sounds good :+1:

packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
jtpio: Right. It's fine to leave it like this for now and we can track that separately.

martinRenou:
Arf. I cannot run the linter locally entirely... I am running out of memory.

Also, the readthedocs deployment does not seem to display the widgets.

jtpio:
> Arf. I cannot run the linter locally entirely... I am running out of memory.

Is it because of ESLint? Maybe we're missing some ignore files in the config.

On CI it looks like it's related to black:

image

So maybe a `black .` locally will do it.

martinRenou:
> Is it because of ESLint? 

Yeah, doing `yarn run lint` even after a fresh restart leads to `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`

martinRenou:
What remains:

- Wiring comm_info_request
- Fix bqplot example (issue related to binary buffers)

jtpio:
> Yeah, doing `yarn run lint` even after a fresh restart leads to `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`

This one happens in the jupyterlab repo too sometimes, but the code base there is bigger.

bollwyvl:
There are some magic incantations to get eslint to log every file it is
looking at... Likely we've just missed something glaring in .eslintignore,
and were sneaking by before...

jtpio:
Kicking the RTD build

jtpio:
> Also, the readthedocs deployment does not seem to display the widgets.

Looking good on RTD :sparkles: 

https://user-images.githubusercontent.com/591645/122111256-cd621f80-ce1f-11eb-8935-5b09bc35d40b.mp4

examples/pyolite - ipywidgets.ipynb
```diff
@@ -0,0 +1,144 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "23b32857-2958-4083-b16a-ac26cd2408d4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from bqplot import *\n",
```
jtpio, line 70: We'll also need the `micropip.install('bqplot')` for `bqplot` to be available in the kernel.

martinRenou: Yes. I was thinking of running `micropip.install('bqplot')` directly in the Notebook, what do you think?

Although we might not want to convey the wrong idea that any widget can be installed in the example notebook...

martinRenou: But still, it's better than having pyolite install bqplot automatically.

jtpio: Yes for the example notebook I think it's completely fine.

We could also add a comment saying that bqplot works, but other widgets might not.

packages/pyolite-kernel/package.json
```diff
@@ -64,9 +64,12 @@
...
   },
   "pyolite": {
-    "packages": [
-      "py/pyolite"
-    ]
+    "packages": {
+      "py/pyolite": "0.1.0",
+      "py/ipykernel": "5.5.5",
+      "py/nbformat": "4.2.0",
+      "py/widgetsnbextension": "3.5.0"
```
jtpio, line 71: Maybe as a follow-up we could find a way to keep the versions in sync with the `__version__` defined in the packages.

martinRenou:
Added ipydatagrid and ipygany in the example notebook. Let's see!

martinRenou:
Mmmh. ipydatagrid's conditional formatting seems to have issues on pyolite (inspect module not able to inspect functions source). And the ipygany example fails with some binary buffers issue. 

I guess we could add those examples in a separate PRs.

bollwyvl:
Very exciting stuff!

Looks like `py2vega` isn't quite working, but getting to `bqplot` is _very_ impressive. 

`pythreejs` would round it out very nicely (and bring along `ipydatawidgets`)... and of course am itching to get `wxyz` in that house...

Stylistically: I think the "flat stack of notebooks" is fine for us, given the limitations of both `tree` file browsers (though having a `data` directory might be good). As this list gets longer, though, we might want to
- tighten up the naming for sorting/filtering and screen real estate, e.g. `py - widgets - core.ipynb`, `js - widgets - core.ipynb` 
- add a `examples/README.md` which we can launch by default... ideally that had a nice table of labeled thumbnails (this is would fall out of, and indeed could _drive_, #6).

martinRenou:
> Looks like py2vega isn't quite working, but getting to bqplot is very impressive.

Indeed. It seems to be related to the inspect Python module behaving weirdly (not working for inspecting functions).

packages/pyolite-kernel/package.json
bollwyvl: Yeah, the #147 piece will grow the ability mess about with `.whl` files... we can figure out some way to sync the versions (maybe a _dumb_ `requirements.txt`, (no `-r` or `git` shenanigans).

packages/pyolite-kernel/py/pyolite/pyolite/display.py
bollwyvl: `update_display` will be _really_ fun, too

martinRenou:
Yep, more examples and clearer notebook layout would be good!

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -8,6 +8,9 @@ import { PromiseDelegate } from '@lumino/coreutils';
 
 import worker from './worker?raw';
 
+import widgetsnbextension from '../py/widgetsnbextension/dist/widgetsnbextension-3.5.0-py3-none-any.whl';
+import nbformat from '../py/nbformat/dist/nbformat-4.2.0-py3-none-any.whl';
+import ipykernel from '../py/ipykernel/dist/ipykernel-5.5.5-py3-none-any.whl';
```
bollwyvl, line 13: We're going to need a way to automate the _hell_ out of this, and make it build-time serviceable. I did a similar thing on [ipydrawio](https://github.com/deathbeds/ipydrawio/blob/master/packages/ipydrawio-webpack/lib/_static.js), but we'll actually be building no-foolin packages, so worth a thought.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -17,11 +17,22 @@ async function loadPyodideAndPackages() {
   await pyodide.loadPackage(['matplotlib']);
   await pyodide.runPythonAsync(`
     import micropip
-    await micropip.install('${_pyoliteWheelUrl}')
+    # Doing it in two times is important, otherwise ipywidgets doesn't pick our mocks
```
bollwyvl, line 20: i kinda think we have to get these out of the critical path... loading `matplotlib` every time isn't fun, and this will start to get out of hand.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -171,6 +281,18 @@ self.onmessage = async (event: MessageEvent): Promise<void> => {
...
+    case 'comm-open':
+      return commOpen(messageContent);
+
+    case 'comm-msg':
+      return commMsg(messageContent);
+
+    case 'comm-close':
+      return commClose(messageContent);
+
     default:
```
bollwyvl, line 296: Maybe we can get clever here in type-land, and use the [`switch ... never` pattern](https://javascript.plainenglish.io/never-miss-a-switch-case-with-typescript-684bf5d0e1d1?gi=6215774607fe). Indeed, it might be that it's actually `IKernel<T, U>` where `T` and `U` are the `messageTypes` it accepts/creates respectively.

packages/pyolite-kernel/src/worker.ts
martinRenou: Yeah, this doesn't scale. Let's remove it and install it manually in the Notebook example.

We might want to find a way for jupyterlite to have a plug-in system allowing users to define which packages/wheels they want to load automatically at the start. The same way ipython allows executing code when it starts.

packages/pyolite-kernel/src/worker.ts
martinRenou: I actually failed using more types in this file, loading type definitions from jupyter lab made webpack add a `export = {} `, at the end of the file, turning the worker blob unusable.

I didn't look deeply into the issue, but yeah I totally agree

packages/pyolite-kernel/src/worker.ts
jtpio: Right and in retrospect we don't ship `@jupyter-widgets/jupyterlab-manager` by default. So having `ipywidgets` only might be confusing with widgets not displaying if the prebuilt extension is not installed and served.

bollwyvl: yeah, and it's important folk are able to _own_ the versions of those... anything we hard-code in will be no fun. Now, we _could_ ship it as an optional "preferred" `federated_extensions` (either in the `jupyterlite` wheel, or fetched from the `jupyterlab_widgets-*.whl`), and ship a known-good tarball...

bollwyvl: also, regarding the imports: perhaps adopting a best-effort "runs-everywhere" pattern in the notebooks like:

```py
try:
    import pyolite
    await pyolite.patches.matplotlib()
except ImportError:
    pass
import matplotlib
```

with the _piece de resistance_ being if `importnb` works...

packages/pyolite-kernel/src/worker.ts
bollwyvl: well, if we explore `comlink`, some of these problems _should_ go away (though i _have_ had some issues with webpack5 vs comlink vs wasm and don't have a resolution yet...)

jtpio:
Looks like `ipydatagrid` views can't be displayed on RTD:

image

martinRenou:
Thanks for trying it! I actually cannot get to the error message point without my page to completely crash... Let's remove the ipydatagrid example for now then, and investigate later

jtpio:
Right it looks like it is hitting the ipydatagrid lab extension code so at least the extension should be loaded. So maybe something doesn't make it through.

martinRenou:
Amazing! ipytone is a brilliant name BTW ahah

jtpio:
In general yes it's possible to access the DOM and other JS objects from Pyodide, which means we could for example access a global Tone.js object: https://pyodide.org/en/stable/usage/quickstart.html?highlight=dom#accessing-javascript-scope-from-python

However with JupyterLite, the Pyolite kernel runs in a Web Worker so there is an additional layer of isolation involved, which might make it more difficult to accomplish. There is some related discussion in https://github.com/pyodide/pyodide/issues/1518

jtpio:
> with JupyterLite, the Pyolite kernel runs in a Web Worker

Something that was discussed elsewhere, we could allow starting the same kernel either in the main window or in a web worker like it is now. Or have that new kernel that runs in the same thread be a separate (third-party) `serverlite` extension based on the core pyolite kernel.

bollwyvl:
> One challenge in ipytone is that Tone.js music event scheduling is based on JS callbacks, which is tricky to replicate on the Python side.

If you're already doing this over comms, your widget `.ts` will be running in the same JS thread as `Tone.js`, and I'd think everything is going to just work! Even if you're having to do nasty `application/javascript` stuff, that will also be in the display system in the main thread.

> we could allow starting the same kernel either in the main window or in a web worker like it is now.

While the developer convenience of this is decent... the user experience will be... questionable. pyodide takes up a _lot_ of CPU time and pulls a _lot_ of data. The OG iodide experience was... not super responsive because of running in-thread. If #137 lands, it would at least be _better_.  Once again,

jtpio (approved):
Thanks!!

jtpio:
We can track some of the outstanding comments in separate issues.

martinRenou:
Thanks! I will open those issues, I took note of all of them

jtpio:
Nice. I opened https://github.com/jtpio/jupyterlite/issues/151 and https://github.com/jtpio/jupyterlite/issues/152, but feel free to open more if needed.

benbovy:
Thanks for your answers @jtpio, @bollwyvl, @martinRenou. And nice to see this merged as well as this project evolving rapidly!

> If you're already doing this over comms, your widget .ts will be running in the same JS thread as Tone.js, and I'd think everything is going to just work!

Indeed, I'm using comms and it works pretty well, but it has some limitations, e.g.,

```python
import random
import ipytone

osc = ipytone.Oscillator().to_destination()

def clb(time):
    osc.start(time).stop(time + random.uniform(1, 2))

# this will call `clb` once with a specific `time` argument so that an equivalent
# callback can be re-constructed on the JS side before passing it to Tone.js.
# downside: `random.uniform()` above is evaluated only once too.
ipytone.transport.schedule_repeat(clb, "8n")
``` 

Using [Python callbacks in javascript with pyodide](https://pyodide.org/en/stable/usage/type-conversions.html#using-a-python-callback-with-an-existing-javascript-function) would be more flexible, but if I understand well this is not possible currently.

Sorry this is quite off-topic, I'll open a new discussion if that's worth it.

PR #146, opened 2021-06-15
Bootstrap the `jupyterlite` command line tool

jtpio:
## References

Addresses some parts of #41

## Code changes

In scope for this PR:

- [x] Bootstrap the `jupyterlite` CLI tool with `flit`
- [ ] Copy existing lab extensions from `$PREFIX/share/jupyter/labextensions` to the application folder
- [ ] Update `jupyter-lite.json` with the `federated_extensions` data
- [ ] Replace some of the logic used for RTD demo to use `jupyterlite`
- [ ] Update contributing guide
- [ ] Update CI

These are the first steps to at least make it more convenient to work with third-party federated extensions locally.

## User-facing changes

None

## Backwards-incompatible changes

None

PR #147, opened 2021-06-16
Start python CLI package (alternate)

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlite/gh-41-start-cli?urlpath=lab)

## References

- starts #41
- alternative to #146

## Code changes

- could be on another PR
  - [x] add `mimeExtension` to schema, make `extension` not required
    - we may need to further break up that definition
  - [x] fix discovery of `index.ts?` for typedoc building
  - [x] adds a skeleton labextension package built with `setuptools`
    - [x] #165 very wip, could pull from this PR 
- adds a fairly fully-fleshed-out CLI package built with `pip`
  - [x] extensible via traitlets/entry_points
    - [x] add `--disabled-addons`
    - [ ] allow for replacing "core" `addons`
  - [ ] mostly a `traitlets` wrapper around chained `doit` tasks
    - [x] creates a unique `.jupyterlite.doit.db` (needs docs) 
    - [x] `status` overview of what it _might_ do
    - [x] `list` list of everything you _could_ do
    - [x] `init` just copy in the gold master
    - [x] `build` build the site with user choices! (after `init`)
      - [x] files -> contents 
      - [x] `overrides.json` -> `jupyter-lite.json` 
      - [x] federated_extensions -> `/lab/extensions` -> `{app}/build/themes` -> `jupyter-lite.json`
      - [x] handle `jupyter-lite.ipynb` in more places
      - [x] handle merging `federated_extensions` better
        - [x] everything fails if there's duplicates, probably just need last-in-wins by `(mimeE|e)xtension`
      - [x] #151 handle resolving/fetching/extracting `federated_extensions` from `whl`s
        - this may also become important for `pyolite` patches
    - [x] `check` use all those schema to check the built site (after `build`)
    - [ ] **future** `publish` put it... somewhere (after `build`... or `check`?)
    - [x] `serve` with stdlib (5s time-to-launcher) or tornado (2s)  (after `build`)
    - [x] `archive` build an archive like the `npm` tarball which can be used as a baseline for other sites  (after `build`)
    - [x] add a passthrough to the full `doit` CLI (such as it is), e.g. `jupyter lite doit` 
    - [x] enable access to the tasks below from an external `dodo.py`
      - `from jupyterlite import LiteManager; m = LiteManager(...); globals.update(m.tasks)` 

  - [ ] docs (some started)
    - [ ] narrative
      - [x] step-by-step walkthrough (started)
      - [x] configuring (started)
      - [x] deploying (started)
      - [ ] **future** extending (started)
    - [x] API
      - [x] improve types, use `sphinx-autodoc-typehints`
  - [ ] tests
    - [x] simple command line tests with e.g. `pytest-console-scripts` with a _simple_ fixture
      - [x] at least execute all subcommands
      - [x] check one `--files`
      - [x] check nested directories of `--files` 
      - [x] check `--overrides`
      - [ ] check `--disable-addons`
      - [x] check idempotence of `--output-archive` to `--app-archive` 
        - [ ] maybe rename? `Seed`? `Bulb`? 
        - > I'd strongly like to see a `lightbulb | flower` if someone's feeling artisitic
    - [ ] *maybe future*: expand fixture to exhaustive `hypothesis-(jsonschema)` (maybe a separate PR, but will probably be more efficient up-front)
    - [x] *future*: include in #6
- start moving dodo stuff into the package, then still calling it (and new stuff) from the outer dodo
  - [x] isort
  - [x] pyflakes
  - [x] `jupyterlite check` _before_ the docs build
  - [x] `jupyterlite archive` _after_ the docs build
  - [x] `pytest-cov`
  - [x] `pytest-html`
  - [x] create full `dist` with `SHA256SUMS` of everything
  - [x] #163 check reproducibility/idempotence of `--output-archive` across platforms/pythons

## User-facing changes

| note | screenshot |
|------|-----|
| for the docs, this treats `examples/` as the `--lite-site` and deploys to `/build/docs-app`, which then gets copied into `docs/_build/_static` by sphinx. |  Screenshot from 2021-06-17 08-20-24

- n/a


jtpio:
Wow thanks for pushing this!

This is way more advanced than https://github.com/jtpio/jupyterlite/pull/146, so we should continue here instead :+1:

py/jupyterlite/src/jupyterlite/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""a command line tool for building JupyterLite sites"""
+__version__ = "0.1.0"
```
jtpio, line 2:
```suggestion
__version__ = "0.1.0a0"
```

bollwyvl: right, we need to single-source that from... someting, or at least put it under integrity control

bollwyvl: maybe the version of the tarball it ships

bollwyvl: normalized this and pyolite (which is also hard-coded). Need a plan... if only an integrity regex check...

bollwyvl:
The binder (added linked at the top) now has the CLI ready for demo:

Screenshot from 2021-06-17 12-10-02

- Starting `jupyter lite --serve`
- Use the `jupyter-server-proxy` URL:

```
https://hub.gke2.mybinder.org/user/<user-abcd123>/proxy/8000
```

I _don't_ think it's worth doing a `jsp`-specific thing, as we'll _eventually_ have the fully-fledged UI available from within lab (at least, if not lite itself :scream_cat:).

bollwyvl:
Welp, I didn't quite make it to the wheel stuff (turns out reproducible builds are hard) but I'm pretty okay with everything else in here. As we have nothing that does anything like that right now, it's just _not a feature_ rather than _broken_.

Aside from the  `jupyter lite doit` backdoor, there shouldn't be _too_ much stuff that leaks out from the chosen task engine, so we can change/simplify it in the future if we don't need the robustness.  However, especially in light of the wheels, we can't be downloading (and not caching) potentially huge `whl` every time, so I'm kinda okay with `doit`, for now.

I promise not to push anything else for the next 8 hours or so (ha) so this is as good a time as any for review. I think breaking this one up into smaller pieces would be rather hard, except for _perhaps_ the reproducible build stuff, but it turns out that has impacts across the build lifecycle, so I'd really rather not...

jtpio:
Thanks a lot @bollwyvl for this, it's looking great!

jtpio:
Quick smoke test on the RTD: it looks like the federated extensions are not copied and served to the frontend.

bollwyvl:
Sure enough, not in the [SHA256SUMS](https://jupyterlite--147.org.readthedocs.build/en/147/_static/SHA256SUMS)!

bollwyvl:
Yeah, as recently as https://readthedocs.org/projects/jupyterlite/builds/14058569/ it worked, so shouldn't be too hard. Guess need more tests, too!

bollwyvl:
Ok, looks like everything is good on RTD... testing out binder, too...

bollwyvl:
...binder looks solid

bollwyvl:
So I was thinking some about the implications of reusing the build components interactively/as part of a serverextension: basically, the `LiteBuildConfig` needs to be fully separated from the manager config... and a lot of references need to change from `self.manager.whatever` to `build.whatever`, and won't have much impact on the configuration/documentation, so is probably safe to hold off on for now...

py/jupyterlite/src/jupyterlite/manager.py
```diff
@@ -0,0 +1,191 @@
+"""Manager for JupyterLite
+"""
+
+import doit
```
jtpio, line 4: Should this be added to the install requires in `pyproject.toml`?

jtpio:
> Aside from the jupyter lite doit backdoor, there shouldn't be too much stuff that leaks out from the chosen task engine

Do we need to expose it, or is it mostly for convenience?

bollwyvl:
> Do we need to expose it, or is it mostly for convenience? 

Just for testing. I think if someone was going to be using the underlying `doit` stuff (read: i would) they would do so programatically.

py/jupyterlite/src/jupyterlite/manager.py
bollwyvl: haha! a527085

bollwyvl:
> do so programatically

Added a notebook with an example of just this...

jtpio:
Thanks for all the work here, it looks great and solid!

I think we can ahead and merge this one, which should make it easier to iterate on.

One potential concern would have been related to having the tasks backed by `doit`, which could look slightly complicated to folks discovering the codebase since it's another tool and they might not be familiar with it. However as mentioned above it's an implementation detail, and would not be visible to consumers of the `jupyterlite` CLI.

bollwyvl:
Right. As the "Hard Way" docs reflect, this is not a site structure one can just knock together and maintain with a pile of scripts, so I feel we need _something_ to organize it. Over the years, I've tried all the things, `invoke`, `fabric`, `pants`, `grunt`, `brocoli`, `gulp`... and of course `make`... I sure as shooting wouldn't want to rewrite content-based cache invalidation for something like this. `doit` fills the bill, and indeed I would love to see it more widely adopted in jupyter tech stacks, as "run one command to know things are okay" is a pattern that _few_ of our projects have... but yeah, an implementation detail for now!

jtpio:
> I would love to see it more widely adopted in jupyter tech stacks, as "run one command to know things are okay" is a pattern that _few_ of our projects have

Yeah it would be interesting to see if that could become more used indeed.

For now let's see how it goes!

jtpio (approved):
Thanks!!

PR #156, opened 2021-06-18
Try ipyleaflet

martinRenou:
cc. @giswqs

### DIFF
--- a/examples/ipyleaflet.ipynb
+++ b/examples/ipyleaflet.ipynb
@@ -1,0 +1,259 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import micropip\n",
+    "\n",
+    "await micropip.install(['bqplot', 'ipyleaflet'])"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "from urllib.request import urlopen\n",
+    "import json\n",
+    "from datetime import datetime\n",
+    "\n",
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "\n",
+    "from js import fetch\n",
+    "\n",
+    "from ipywidgets import Dropdown\n",
+    "\n",
+    "from bqplot import Lines, Figure, LinearScale, DateScale, Axis\n",
+    "\n",
+    "from ipyleaflet import Map, GeoJSON, WidgetControl"
+   ]
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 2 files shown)


examples/ipyleaflet.ipynb
```diff
@@ -0,0 +1,248 @@
...
+    "from ipyleaflet import Map, GeoJSON, WidgetControl"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data = pd.read_json('https://raw.githubusercontent.com/jupyter-widgets/ipyleaflet/master/examples/nations.json')"
```
jtpio, line 41: We probably still need to use the `from js import fetch` trick to fetch extra data for now:

https://github.com/jtpio/jupyterlite/issues/119#issuecomment-854495013

jtpio:
Nice, thanks!

Looks like the latest `ipyleaflet` removes the hard dependency on `shapely`: https://github.com/jupyter-widgets/ipyleaflet/blob/master/CHANGELOG.md#v0140

And now shows up nicely on the RTD preview :tada: 

image

jtpio:
Maybe we can move this `console.log`: 

https://github.com/jtpio/jupyterlite/blob/5f38f5cfbf373ad2c8a6c4bfd4f2d54124c36525/packages/pyolite-kernel/src/worker.ts#L270

Down to the execute request statement:

https://github.com/jtpio/jupyterlite/blob/5f38f5cfbf373ad2c8a6c4bfd4f2d54124c36525/packages/pyolite-kernel/src/worker.ts#L276

Since it seems to be flooding the dev tools with comm messages (for example when moving the mouse):

image

martinRenou:
Cool! 

Now might be a good time to reorder notebooks in the examples directory

giswqs:
How to access the JupyterLite RTD preview?

jtpio:
@giswqs there is a link below in the checks:

image

jtpio:
@martinRenou is this ready for review and merge?

Looks like it's working well on the RTD preview, and it would be a great example to add to the demo site.

martinRenou:
Yep! Should we change the directory layout of the examples though? Would having a "Interactive Widgets" directory be a good idea?

Also maybe moving all pyolite examples in a separate dir?

jtpio:
Sounds good.

It would be fine to do the example restructuring either here or in a separate PR.

martinRenou:
Ok :) Let's do it in another PR then. And get the ipycanvas example in as well before the restructuring?

jtpio:
> And get the ipycanvas example in as well before the restructuring?

Sounds good :+1:

jtpio (approved):
Thanks!

martinRenou:
Thank you! :)

PR #157, opened 2021-06-18
Move console.log inside worker

martinRenou:
Originally suggest by @jtpio https://github.com/jtpio/jupyterlite/pull/156#issuecomment-863860067


jtpio (approved):
Thanks!

Issue #121, opened 2021-06-01
Release scripts


jtpio:
At the very least, we need some easy way to:

- bump version numbers
- update dependencies

Later we'll also want to publish packages to `npm` as tarballs.

Lab has some tooling in the `buildutils`: https://github.com/jupyterlab/jupyterlab/tree/master/buildutils
Retro also has some in https://github.com/jupyterlab/retrolab/tree/main/buildutils

We could do something similar.

Or have that in Python in `dodo.py`.

jtpio:
There is also https://github.com/jupyter-server/jupyter_releaser

jtpio:
The version should also go to the page config so it shows up in the UI:

![image](https://user-images.githubusercontent.com/591645/120514996-22d51000-c3ce-11eb-8cee-cc5b5f6a34dc.png)

jtpio:
Trying out the Jupyter Releaser on this repo to create the changelog for `v0.1.0a1`.

Looks like the workflow does not work just yet: https://github.com/jtpio/jupyter_releaser/runs/2931307989

bollwyvl:
Hm... i think we'd want the version to live in the "last resort" `jupyter-config-data` in the `/app/index.html` which might be more annoying to update than `jupyter-lite.json`, as we want that to be a deployer-owned file (even if we ship a minimum viable one). If nothing else, I'd see us wanting to put that under an integrity check against "the version" in `package.json`.

jtpio:
In terms of flow, we could probably stick to one version for all JS and Python packages to keep things simple. Instead of versioning the JS packages independently. 

There could then be a `doit` task taking care of the following:

- bump the js packages, for example using the `update-dependency` script from `@jupyterlab/buildutils`
- bump each of the Python packages (`jupyterlite` and `pyolite` for now) with `tbump`
- translate the Python specifier to npm, for example `0.1.0a3` -> `0.1.0-alpha.3`
- ensure integrity and resolutions

PR #169, opened 2021-06-22
Add changelog for `v0.1.0a0`

jtpio:
## References

Backfill the `v0.1.0a0` changelog using `github-activity`: https://github.com/executablebooks/github-activity

This is in preparation for having automated releases (#121, probably using [Jupyter Releaser](https://github.com/jupyter-server/jupyter_releaser)), so we can have a consistent changelog file and release content.

## Code changes

None

## User-facing changes

None

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## v0.1.0a1 (Unreleased)
+## 0.1.0a1 (Unreleased)
 
 ### `jupyter lite` CLI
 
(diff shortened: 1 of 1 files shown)


jtpio:
Ah looks like something is off with the RTD build.

Just rebased https://github.com/jtpio/jupyterlite/pull/109 and it's also happening there: https://readthedocs.org/projects/jupyterlite/builds/14071447/

bollwyvl:
looks like `flit` ain't happy for the py-in-js packages... not sure what to make of that... will take a look locally

PR #170, opened 2021-06-22
Investigate RTD build issues

bollwyvl:
## References

- checking out RTD issues on #169

## Code changes

- use `flit build --debug`
- commit some changed files
- fix some dangling chad from #147 (finding settings in `all.json`)

## User-facing changes

- n/a

- n/a


bollwyvl:
hm, at a bit of a loss. the debugging stuff indicates it thinks a _lot_ of files have changed, which seems really odd.

a couple options:

- start checking in the `js:py` `whl` files
  - if we use the reproducible build stuff, it might not be so bad
- clobber/add to .gitignore some stuff on rtd
  - e.g. we could copy the contents into a `/tmp` directory and build from there without there being any vcs stuff involved

jtpio:
Thanks for investigating :+1: 

This looks odd, since the latest `main` build passed just fine on commit e45ed32ab5b8b1a0e1acf70f2801e61eee1874c4 (which includes the bigger change from #147):

image

bollwyvl:
Welp, it's not reporting back success, but did work ([log](https://readthedocs.org/projects/jupyterlite/builds/14078561/), [site](https://jupyterlite--170.org.readthedocs.build/en/170/_static/lab/index.html)). Provided it _does_ eventually report back, this is probably ready for review.

I still don't understand the issue with `flit`, as this most recent build didn't even fail and do the tmpdir thing... but at least we're moving forward.

bollwyvl:
Ha, now i'm entirely confused, because it's failing again...

bollwyvl:
Alright, everything green now. I don't know if the issue lies with `flit` or `rtd`, but neither is giving super helpful responses... and it works now. I feel like anything that isn't _The CLI_ running in RTD can be made of duct tape.

jtpio:
Wow, thanks for investigating this.

> I don't know if the issue lies with `flit` or `rtd`

It looks like it might be related to some changes on the rtd runners? (based on the observation in https://github.com/jtpio/jupyterlite/pull/170#issuecomment-866631641) 

Gonna kick the build one more time to double check.

jtpio (approved):
Thanks!

PR #171, opened 2021-06-24
Use IPython in Pyolite kernel

madhur-tandon:
Using `await micropip.install('ipython')` and then `import IPython` works in the notebook.
However, we want to be able to `import IPython` by default, so it makes sense to run `await micropip.install('ipython')` in the `worker.ts` script.

However, it doesn't work and we get `ModuleNotFoundError: No module named 'IPython'` on using `import IPython` in the notebook.


jtpio:
Thanks @madhur-tandon for opening this!

Quickly testing on the [RTD preview](https://jupyterlite--171.org.readthedocs.build/en/171/) and it's looking promising :sparkles: 

image

jtpio:
So if `IPython.display` works fine, we might be able to remove the custom `display` in https://github.com/jtpio/jupyterlite/blob/main/packages/pyolite-kernel/py/pyolite/pyolite/display.py?

packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
```diff
@@ -14,38 +14,16 @@
...
-
-core = types.ModuleType("core")
-core.pylabtools = pt
+fcntl_mock = types.ModuleType("fcntl")
+resource_mock = types.ModuleType("resource")
 
-sys.modules["IPython.display"] = pyolite.display
-sys.modules["IPython"] = ip
-sys.modules["IPython.core"] = core
-sys.modules["IPython.core.getipython"] = ip
```
jtpio, line 49: Checking on the RTD preview, it looks like `get_ipython()` would now return `None`:

image

Related to: https://github.com/ipython/ipython/blob/abee6378682179bbd3a57bd11eda65a5ad9cbaae/IPython/core/getipython.py#L17-L24

So we might still need that mock for now?

martinRenou: @madhur-tandon Either we continue to mock this one for now and merge the PR as is (so that all examples are working), or we create an InteractiveShell and have get_ipython return it? I guess doing it in two PRs is good 👍🏽 the second pass of implementing the interactiveshell might be a big PR

jtpio:
So what we can do for now to be able to merge this PR:

- add the `get_ipython` mock back: https://github.com/jtpio/jupyterlite/pull/171#pullrequestreview-691624933
- remove some of the primitive from the `pyolite.display` module if they can be used from IPython directly: 

https://github.com/jtpio/jupyterlite/blob/3930210c422f52de4597a5d44013fdf7d01c3bd7/packages/pyolite-kernel/py/pyolite/pyolite/display.py#L53-L86

madhur-tandon:
Summary:

1) We can remove `HTML`, `Markdown`, `Latex` and `JSON` from `display.py` but we do need `Image` class for `matplotlib` to function correctly (using that patch). `Image` imported from `IPython.display` doesn't work. This step is fine.
2) For the stuff regarding adding the `mock` stuff back, we need the following statements:

```
class IPMock:
    def __init__(self, kernel):
        self.kernel = kernel


class InteractiveShellMock:
    pass


ip = types.ModuleType("IPython_mock")
ip_mock = IPMock(kernel_instance)
ip.get_ipython = lambda: ip_mock
ip.InteractiveShell = InteractiveShellMock

sys.modules["IPython.core.getipython"] = ip
```

This however breaks `ipywidgets` demo on Jupyterlite. On writing `slider`, we don't get the widget, the cell executes, but there is no widget present.

So @martinRenou suggested that we also add `sys.modules["IPython.core.interactiveshell"] = ip` line.

Note that we will not be using `sys.modules["IPython"] = ip` at all because this will have an effect of replacing the original `IPython` with our fake `IPython`.

Now after adding the 2 lines i.e. `sys.modules["IPython.core.getipython"] = ip` and `sys.modules["IPython.core.interactiveshell"] = ip`, we get the error:

```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/lib/python3.8/site-packages/ipywidgets/__init__.py", line 23, in <module>
    from IPython import get_ipython
  File "/lib/python3.8/site-packages/IPython/__init__.py", line 56, in <module>
    from .terminal.embed import embed
  File "/lib/python3.8/site-packages/IPython/terminal/embed.py", line 15, in <module>
    from IPython.core.interactiveshell import DummyMod, InteractiveShell
ImportError: cannot import name 'DummyMod' from 'IPython_mock' (unknown location)
```

This is because, since we did not use `sys.modules["IPython"] = ip` i.e. we ARE using the real `IPython`, it's trying to import it's own `interactiveshell` (which is kinda patched). So we are in a mixed state where we don't patch the whole `IPython`, but only a part of it. This leads to the `cannot import name 'DummyMod'` error.

But since all this stuff is going to be removed eventually, so we should begin with the real work i.e. using IPython's own interactive shell rather than trying it in 2 PRs. This is the conclusion. cc: @martinRenou

jtpio:
Thanks @madhur-tandon for the detailed write-up :+1: 

> This is because, since we did not use `sys.modules["IPython"] = ip` i.e. we ARE using the real `IPython`, it's trying to import it's own `interactiveshell` (which is kinda patched). So we are in a mixed state where we don't patch the whole `IPython`, but only a part of it. This leads to the `cannot import name 'DummyMod'` error.

Right, if we add partial support for IPython we might get into a state where half of the things might or might not work, which can indeed be confusing.

packages/pyolite-kernel/py/pyolite/pyolite/interpreter.py
```diff
@@ -19,9 +29,8 @@ def display(self, result):
...
-            self.load_packages_and_run(self.run_complete, code)
-        )
-        result = await self.run_complete
-        if result is not None:
-            display(result)
+        exec_code = self.transform_cell(code)
+        await _load_packages_from_imports(exec_code)
+        self.result = self.run_cell(code)
+        if self.result is not None:
+            display(self.result)
```
martinRenou, line 36: We'll have to remove this `display(self.result)` and use a custom displayhook class instead

See
https://github.com/ipython/ipython/blob/e37bc7afddaf74327eabf661b4ebd4e1047299b5/IPython/core/interactiveshell.py#L444

jtpio:
FYI @madhur-tandon you might want to rebase on the latest `main` to grab the fix for the windows check on CI.

madhur-tandon:
1) re-implement `stdout` callbacks since `interpreter.stdout_callback` was from pyodide's console and is not available in `IPython`
2) Implement `display publisher` the correct way i.e. by inheriting from `IPython`
3) Make sure `get_ipython()` doesn't return None so that `get_ipython().kernel` can be accessed in `comm.py`

martinRenou:
4. Use InteractiveShell's logic for the completion
5. Bring back Error handling

packages/kernel/src/kernel.ts
```diff
@@ -241,6 +241,42 @@ export abstract class BaseKernel implements IKernel {
...
+    this._sendMessage(message);
+  }
+
+  /**
+   * Send a `clear_output` message to the client.
+   *
+   * @param content The clear_output content.
+   */
+  protected clearOutput(content: KernelMessage.IClearOutputMsg['content']): void {
+    const message = KernelMessage.createMessage<KernelMessage.IClearOutputMsg>({
```
martinRenou, line 269: 🚀

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
```diff
@@ -1,14 +1,17 @@
...
 
-from .display import display_publisher
+from .custom_display import XDisplayHook, XDisplayPublisher
 from .interpreter import Interpreter
-from .custom_display import XDisplayHook
+
 
 class Pyolite:
     def __init__(self):
-        self.interpreter = Interpreter(kernel=self, displayhook_class=XDisplayHook)
```
martinRenou: You need to pass the `display_pub_class` to the Shell constructor

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
```diff
@@ -1,14 +1,17 @@
...
-        self.display_publisher = display_publisher
```
martinRenou, line 11: And `self.display_publisher` would be `self.interpreter.display_pub`

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
martinRenou: Ah. Commenting old code, sorry.

madhur-tandon: Yes I did, this is an outdated file

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
madhur-tandon: we won't need to use it anymore, because we are not doing `kernel.display_publisher.display_callback = displayCallback;`

we are directly doing

```
  interpreter.display_pub.clear_output_callback = clearOutputCallback;
  interpreter.display_pub.display_data_callback = displayDataCallback;
  interpreter.display_pub.update_display_data_callback = updateDisplayDataCallback;
```

madhur-tandon: Marking as resolved.

packages/pyolite-kernel/py/pyolite/pyolite/interpreter.py
madhur-tandon: aye aye captain 🚀

packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
madhur-tandon: This is fixed now.

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -91,25 +91,9 @@ export class PyoliteKernel extends BaseKernel implements IKernel {
...
-      case 'stderr': {
-        const { message } = msg.stderr;
-        const content = {
-          event: 'stream',
-          name: 'stderr',
-          parentHeader,
-          text: message ?? msg.stderr
-        } as KernelMessage.IStreamMsg['content'];
-        this.stream(content);
+      case 'stream': {
```
martinRenou, line 94: 🚀 

This is closer to the official protocol indeed

madhur-tandon: 💪🏻

jtpio: Nice!

packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
```diff
@@ -5,47 +5,35 @@
...
+from .patches import ensure_matplotlib_patch
 
-ip_mock = IPMock(kernel_instance)
+# apply patches for available modules
+ensure_matplotlib_patch()
 
-ip = types.ModuleType("IPython")
-ip.get_ipython = lambda: ip_mock
-ip.InteractiveShell = InteractiveShellMock
+from .display import XStream
```
jtpio, line 27: Small cosmetic thing: the `X` prefix probably comes from the `xeus-python` inspiration?

It would be fine to have it here too. Otherwise an alternative could be `Lite`?

madhur-tandon: Yes, We will change this. The first step is to make all things work though.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -183,39 +198,36 @@ async function execute(content: any) {
...
   interpreter.payload_manager.clear_payload();
 
   if (typeof interpreter._last_traceback === 'undefined') {
-    reply_content['status'] = 'ok';
-    // TODO: set reply_content['user_expressions']
+    results['status'] = 'ok';
+    // TODO: set results['user_expressions']
   } else {
-    console.log(interpreter._last_traceback);
     const last_traceback = formatResult(interpreter._last_traceback);
```
martinRenou, line 211: It seems you will need to call `formatResult` on each element individually (`formatResult` does not seem to work recursively on your `_last_traceback` object).

packages/pyolite-kernel/src/worker.ts
```diff
@@ -183,39 +198,36 @@ async function execute(content: any) {
...
-    console.log(last_traceback);
-    reply_content['status'] = 'error';
-    reply_content['ename'] = last_traceback['ename'];
-    reply_content['evalue'] = last_traceback['evalue'];
-    reply_content['traceback'] = last_traceback['traceback'];
-  }
+    results['status'] = 'error';
+    results['ename'] = last_traceback['ename'];
```
martinRenou, line 213: `results['ename'] = formatResult(last_traceback['ename'])`

etc

packages/pyolite-kernel/src/worker.ts
```diff
@@ -123,6 +123,21 @@ async function execute(content: any) {
     });
   };
 
+  const publishExecutionError = (ename: any, evalue: any, traceback: any): void => {
+    const bundle = {
+      ename: ename,
```
martinRenou, line 128: `ename: formatResult(ename),`

etc

martinRenou: But this does not seem to work for some reason (leading to `Uncaught TypeError: obj.forEach is not a function`, because `ename.toJs()` does not work as intended)

packages/pyolite-kernel/py/pyolite/pyolite/interpreter.py
```diff
@@ -32,21 +32,8 @@ def enable_gui(self, gui=None):
...
-    ):
-        try:
-            etype, value, tb = self._get_exc_info(exc_tuple)
-        except ValueError:
-            print("No traceback available to show.", file=sys.stderr)
-            return
-
-        self._last_traceback = {"ename": etype, "evalue": value, "traceback": tb}
+    def _showtraceback(self, etype, evalue, stb):
+        self._last_traceback = {"ename": etype, "evalue": evalue, "traceback": stb}
```
martinRenou, line 36: Maybe do `self._last_traceback = {"ename": str(etype), "evalue": str(evalue), "traceback": stb}`

martinRenou:
This is looking gooooooood!

packages/kernel/src/kernel.ts
```diff
@@ -241,6 +241,23 @@ export abstract class BaseKernel implements IKernel {
     this._sendMessage(message);
   }
 
+  /**
+   * Send a `execute_result` message to the client.
+   *
+   * @param content The execute_result content.
+   */
+  protected executeResult(content: KernelMessage.IExecuteResultMsg['content']): void {
```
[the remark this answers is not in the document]
madhur-tandon: This is done now.

madhur-tandon:
Ready for review. cc: @jtpio @martinRenou

martinRenou (approved):
This is rad! It looks good to me :) If all Notebooks are still working as expected it's 👍🏽 from me

martinRenou:
Actually, there is an issue in the execution count for the execute_result, it's always `[1]`:

exec_count

bollwyvl:
Not a blocker at all on this, which is of course fabulous work :cake: :tada: , but has me thinking: for #6, we could split up (or tag) the example notebooks by those that are reproducible by a non jupyterlite kernel. Heck, maybe we can hook into the `nbval` toolchain. Having an oracle is nice for catching false positives.

jtpio:
Looking great, thanks Madhur!

> If all Notebooks are still working as expected it's 👍🏽 from me

Yes, we can do a quick smoke test on the ReadTheDocs demo just to make sure most notebooks work. Hopefully soon we'll have #6 in place to help us with this.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -98,87 +106,136 @@ async function sendComm(
...
  */
 async function execute(content: any) {
-  const stdoutCallback = (stdout: string): void => {
+  const publishExecutionResult = (
+    prompt_count: any,
+    data: any,
+    metadata: any
+  ): void => {
+    const bundle = {
+      execution_count: formatResult(prompt_count),
```
jtpio, line 115: Does `prompt_count` need to be passed through `formatResult`, or could it be used directly if it already comes as a `number`?

jtpio:
Also as a follow-up, we might want to elaborate a bit more on the `Pyolite` user docs to mention `IPython` (could definitely be done separately): https://jupyterlite.readthedocs.io/en/latest/kernels/pyolite.html

packages/pyolite-kernel/src/worker.ts
madhur-tandon: Yes, it can be passed directly, fixed it.

examples/pyolite.ipynb
```diff
@@ -32,7 +32,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from IPython.display import Markdown, HTML, JSON, Latex"
+    "from IPython.display import display, Markdown, HTML, JSON, Latex"
```
martinRenou, line 35: Do we need to import display? Is it not injected into the user scope automatically?

madhur-tandon: Yeah, I think we need not import it. Let me change it then.

jtpio:
Looking great on RTD :+1: 

@madhur-tandon is there anything else to add, or is this ready for review and merge?

madhur-tandon:
I think this is ready. We can do more refactoring + docs separately.

martinRenou (approved):
LGTM!

jtpio (approved):
Thanks!!

jtpio:
Will cut a new release with this next.

martinRenou:
🚀🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

jtpio:
https://pypi.org/project/jupyterlite/0.1.0a3/

https://github.com/jtpio/jupyterlite/releases/tag/v0.1.0a3

PR #172, opened 2021-06-24
Split up CI into multiple jobs, test on more platforms

### LINKED ISSUE #163: Expand CI Coverage
### Problem

After #147, we'll have python wheels we hope to actually start (pre-)releasing, without having taken a look on non-Linux platforms.

### Proposed Solution

Split up the current core workflow:
- `build`
  - publishes core empty site build and PyPI/npm-bound distributions 
  - even more minimal linux-only `build` that just generates stuff we would actually publish, without actually exercising features
- `lint`, which runs all the lint tools (on linux)
  - probably get _more_ aggressive here, maybe add CodeQL, etc.
- `docs`, build the docs site
- `test`, against the as-built python packages on platforms x pythons

### Additional context

See:
- [jupyterlab-lsp](https://github.com/krassowski/jupyterlab-lsp/blob/master/.github/workflows/job.test.yml)
- [retrolab](https://github.com/jupyterlab/retrolab/blob/main/.github/workflows/build.yml)

### Design Ideas

- add some more environment variable flags for `doit` to control whether a `dev` or `prod` install is desired

bollwyvl:
## References

- fixes #163

## Code changes

- [x] splits up the build job, to get to python testing on different platforms faster
  - all told, seems to shave about a minute of total CI (while testing more)
- [x] uploads test reports
- [x] uses `SOURCE_DATE_EPOCH` for builds
- [x] another pass over requirements, etc.

## User-facing changes

- n/a

- n/a/

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -42,7 +42,6 @@ dependencies:
   - pytest-check-links
   # test
   - ansi2html
-  - hypothesis-jsonschema
   - pytest-console-scripts
   - pytest-cov
   - pytest-html
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,13 +19,24 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Setup pip (base)
+        run: pip install -U pip setuptools wheel
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: |
+            ${{ env.CACHE_EPOCH }}-${{ runner.os }}-pip-${{ hashFiles('requirements*.txt') }}
+          restore-keys: |
+            ${{ env.CACHE_EPOCH }}-${{ runner.os }}-pip-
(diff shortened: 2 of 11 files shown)


bollwyvl:
CI should work the first time, but wasn't testing against RTD. Hooray more RTD debugging!

bollwyvl:
Ha, worked the first time. when does that happen?

https://jupyterlite--172.org.readthedocs.build/en/172/

bollwyvl:
What this doesn't add is any cross-platform reproducibility checks... can add that here, or do later... or not care about cross-platform reproducible builds, which is probably also fine.

As the docs site does a lot of _stuff_, it would be a good candidate for checking reproducibility of some of our complex behavior, but we'd probably want to figure out the wheel story for `federated_extensions`: we shouldn't need to install the entire interactive scientific computing stack to get one hash.

jtpio:
Thanks!

> What this doesn't add is any cross-platform reproducibility checks... can add that here, or do later... or not care about cross-platform reproducible builds, which is probably also fine.

Yes it would be fine to think about that separately :+1:

jtpio (approved):
Thanks!

PR #173, opened 2021-06-25
fix up copy, rename, delete for directories

### LINKED ISSUE #99: Renaming directories doesn't rename children
## Description

When using the _Rename_ context menu item against a local directory, the director gets renamed, but not its children, so they disappear.

## Reproduce

1. Use the `[folder+]` icon
2. enter `Untitled Folder`
3. make a notebook `Untitled.ipynb`
4. navigate back up
5. rename the folder to `foo`
6. re-enter folder, see nothing
7. navigate back up
8. rename the folder back to `Untitled Folder`
9. re-enter folder, see `Untitled.ipynb`

## Expected behavior

On step 6, see `Untitled.ipynb`

## Context

- discovered while "testing" #94 :laughing: 
  - that PR will certainly make robust renaming more complicated, but not sure if it needs to be fixed on that PR...

bollwyvl:
## References

- fixes #99

## Code changes

- uses the `Contents.get` method (instead of `Contents._contents.getItem`) in most places in `contents.ts`
- append a `/` to the `key` check to avoid deleting `DELETE ME, NEVER.txt` when deleting `DELETE ME/`

## User-facing changes

- Should be able to rename files and directories more confidently

- n/a

### DIFF
--- a/packages/contents/src/contents.ts
+++ b/packages/contents/src/contents.ts
@@ -149,13 +149,13 @@ export class Contents implements IContents {
     let name = PathExt.basename(path);
     toDir = toDir === '' ? '' : `${toDir.slice(1)}/`;
     // TODO: better handle naming collisions with existing files
-    while (await this._storage.getItem(`${toDir}${name}`)) {
+    while (await this.get(`${toDir}${name}`, { content: true })) {
       const ext = PathExt.extname(name);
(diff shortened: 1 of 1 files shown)


bollwyvl:
quacks ok on RTD...

jtpio (approved):
Thanks!

PR #175, opened 2021-06-28
add --user for initial pip install in CI

### LINKED ISSUE #174: Failing windows check on CI
Looks like the windows build job is now failing on CI during the `pip install -U pip setuptools wheel` step: https://github.com/jtpio/jupyterlite/runs/2931131528

image

Maybe one of the suspects could be the new version of pip (`21.1.3`) released 2 days ago: https://pypi.org/project/pip/21.1.3/

bollwyvl:
## References

- fixes #174

## Code changes

- Per error message, add `--user` to initial `pip install`.
  - didn't have to do this for a few months, but :shrug:  

## User-facing changes

- n/a

- n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           python-version: 3.8
       - name: Setup pip (base)
-        run: pip install -U pip setuptools wheel
+        run: pip install --user -U pip setuptools wheel
       - name: Cache (pip)
         uses: actions/cache@v2
         with:
@@ -81,7 +81,7 @@ jobs:
         with:
           python-version: 3.8
       - name: Setup pip (base)
-        run: pip install -U pip setuptools wheel
+        run: pip install --user -U pip setuptools wheel
       - name: Cache (pip)
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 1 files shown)


bollwyvl:
@martinRenou every once and a while the CI environments vs Pip get confused, and there are permission issues on one more more platforms. I've seen this on OSX as well as Windows. This, of course, is odd, as the CI workers generally run with way more perms than they need, but again :woman_shrugging:

jtpio:
Thanks @bollwyvl!

I was about to check with `21.1.2` just to see, but it might also be related to some changes on the github runners so yes it should be fine to have `--user` on CI.

jtpio:
So it looks like it's passing with `21.1.2` in https://github.com/jtpio/jupyterlite/pull/176.

.github/workflows/build.yml
```diff
@@ -151,7 +151,7 @@ jobs:
           python-version: 3.8
       - name: Setup pip (base)
         run: |
-          pip install -U pip setuptools wheel
+          pip install --user -U pip setuptools wheel
```
jtpio, line 154: Alternatively we could also add this `--user` flag just for this job since it's the one that was failing?

And the others run on Ubuntu. Or do you think it's better to have `--user` everywhere for consistency?

bollwyvl: >  everywhere for consistency?

yeah, until we get better reuse mechanisms (ahem, yaml anchors)  i'm inclined to keep everything the same as possible. heck, i almost added a matrix to to all the singleton jobs for this reason.

jtpio: :+1:

jtpio (approved):
Thanks!

PR #176, opened 2021-06-28
[tmp] test with pip==21.1.2 on CI

jtpio:
## References

#174 , just out of curiosity


bollwyvl:
yes again, :shrug: ! It may also be some conflict with `python` vs `python3`, though i don't think GHA win has _any_ python...

jtpio:
Closing as fixed by #175

PR #178, opened 2021-06-28
Try the latest dev Pyodide

jtpio:
## References

The latest dev version might include some `micropip` fixes as noticed and discussed in https://github.com/giswqs/leafmap/issues/55.

We can use the RTD preview for testing: https://jupyterlite--178.org.readthedocs.build/en/178/

## User-facing changes

Using the latest (dev) version of Pyodide.

Probably some, so keeping as a draft for now.

### DIFF
--- a/packages/pyolite-kernel-extension/src/index.ts
+++ b/packages/pyolite-kernel-extension/src/index.ts
@@ -12,7 +12,7 @@ import { PyoliteKernel } from '@jupyterlite/pyolite-kernel';
 /**
  * The default CDN fallback for Pyodide
  */
-const PYODIDE_CDN_URL = 'https://cdn.jsdelivr.net/pyodide/v0.17.0/full/pyodide.js';
+const PYODIDE_CDN_URL = 'https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js';
 
 /**
  * A plugin to register the Pyodide kernel.
--- a/packages/pyolite-kernel/src/worker.ts
+++ b/packages/pyolite-kernel/src/worker.ts
@@ -7,13 +7,16 @@ let kernel: any;
 // eslint-disable-next-line
 // @ts-ignore: breaks typedoc
 let interpreter: any;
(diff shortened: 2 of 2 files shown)


bollwyvl:
Got some interest in the INDEXDB/FS stuff, as well... and I see that `bokeh` (and therefore `pyyaml` (and therefore `libyaml`)) also got in recently, all of it running on 3.9. Fun stuff coming up!

bollwyvl:
I think we're circling the drain on getting access to more filesystems on https://github.com/pyodide/pyodide/pull/1596. The API won't be super-well documented, but here's at least [a test](https://github.com/pyodide/pyodide/blob/b3f9d837873e79f49fcae0d28aa3db8f36243859/src/tests/test_filesystem.py) that shows some how some stuff might be used.

jtpio:
Really nice to see this landing upstream soon!

If anyone would like to pick up this PR and start updating to the latest Pyodide, feel free to do so. Thanks!

bollwyvl:
Yeah, so everything's in upstream, but the jsdelivr `dev` assets haven't quite rolled over yet.

Once it is, we should be able to interactively build stuff like:

```py
from pyodide_js import FS
FS.mount()
# put some stuff into the filesystem FROM indexedDB
FS.syncfs(true, ...)
# do some stuff
# put some stuff out of the filesystem TO indexedDB
FS.syncfs(false, ...)
```

The challenge here is that `syncfs` does _all_ the filesystems, so we if were to add some mounts, they would _also_ get updated

Still kinda planning how to apply the new `FS` features most effectively... for example, getting an indexedDB wheel cache going might make it easier to iteratively work on the right stuff on the python side (since kernels would predictably start _much_ faster).

bollwyvl:
Some things that _kinda_ work with what's currently on `dev`:

https://gist.github.com/bollwyvl/616802a1765dc702ee97a0f065adad8c

jtpio:
Closing in favor of #274.

PR #180, opened 2021-06-28
Add JupyterLab Miami Nights theme to the demo site, fix theme unloading

jtpio:
## References

Add `jupyterlab_miami_nights` as a new theme on the RTD demo site.

## Code changes

None

## User-facing changes

image

None

### DIFF
--- a/packages/theme/src/manager.ts
+++ b/packages/theme/src/manager.ts
@@ -36,19 +36,10 @@ export class ThemeManager extends LabThemeManager {
         reject(`Stylesheet failed to load: ${href}`);
       });
 
-      this._link = link;
-      this._unloadCSS();
       document.body.appendChild(link);
+      this['_links'].push(link);
     });
   }
 
-  /**
-   * Unload the previous theme.
-   */
-  private _unloadCSS(): void {
-    this._link?.parentElement?.removeChild(this._link);
-  }
-
-  private _link: HTMLLinkElement | undefined = undefined;
   private _themesUrl = '';
 }
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -56,6 +56,7 @@ dependencies:
   - jupyterlab-tour >=3.1.0
   - jupyterlab-fasta >=3,<4
   - jupyterlab-geojson >=3,<4
+  - jupyterlab_miami_nights
   - plotly >=5,<6
   - theme-darcula
   ### DOCS ENV ###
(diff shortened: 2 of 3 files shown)


jtpio:
Ah looks like the RTD build might have caught an issue with the `jupyter lite build` command:

```
########################################
Execution aborted.
Two different tasks can't have a common target.'/home/docs/checkouts/readthedocs.org/user_builds/jupyterlite/checkouts/180/build/docs-app/lab/build/themes/index.css' is a target for post_build:federated_extensions:copy:theme:lab:@telamonian/theme-darcula and post_build:federated_extensions:copy:theme:lab:@timkpaine/jupyterlab_miami_nights.
.  docs:app:build
.  docs:app:pack
DependencyError - taskid:docs:app:pack
ERROR: Task 'docs:app:pack' saving success: Dependent file '/home/docs/checkouts/readthedocs.org/user_builds/jupyterlite/checkouts/180/build/docs-app/SHA256SUMS' does not exist

```

jtpio:
Build succeeds, seeing some issues when unloading the theme: 

https://user-images.githubusercontent.com/591645/123872758-cf56d300-d935-11eb-94da-3d8dcef16e7c.mp4

jtpio:
Looks like it's related to the links being unloaded in this private method in upstream theme manager in lab:

https://github.com/jupyterlab/jupyterlab/blob/8c15f776dd7fcd221169d282d23eaa1eb27e23a5/packages/apputils/src/thememanager.ts#L396-L402

packages/theme/src/manager.ts
```diff
@@ -36,19 +36,10 @@ export class ThemeManager extends LabThemeManager {
         reject(`Stylesheet failed to load: ${href}`);
       });
 
-      this._link = link;
-      this._unloadCSS();
       document.body.appendChild(link);
+      this['_links'].push(link);
```
jtpio, line 40: Not the most elegant since we are extending `LabThemeManager` from upstream, and `_links` is managed as a private attribute.

jtpio:
> Looks like it's related to the links being unloaded in this private method in upstream theme manager in lab:

https://github.com/jtpio/jupyterlite/pull/180/commits/4a1595b10666f52f1477b85bcc715f24bf921e98 should fix it:

https://user-images.githubusercontent.com/591645/123925313-1327f780-d98b-11eb-9bb2-f837c3b9a22b.mp4

packages/theme/src/manager.ts
jtpio: As a follow-up, we should check whether we can just reuse the upstream plugin instead of having our own here.

Probably this was to be able to define the custom themes url, but maybe that is already configurable.

PR #182, opened 2021-06-28
Add `jsonschema` as dependency

jtpio:
## References

Looks like it was missing when installing `jupyterlite` in a bare environment (without `jupyterlab` / `jupyterlab_server`).

## Code changes

None

## User-facing changes

`jupyter lite build` in a bare environment would not give an error.

None

### DIFF
--- a/py/jupyterlite/pyproject.toml
+++ b/py/jupyterlite/pyproject.toml
@@ -10,6 +10,7 @@ requires-python = ">=3.7"
 requires = [
     "doit",
     "entrypoints",
+    "jsonschema >=3.0.1",
     "jupyter_core >=4.7",
     # TODO: support wheels https://github.com/jtpio/jupyterlite/issues/151
     # "wheel",


bollwyvl:
Good call. `jupyter_server` was bringing it along in test.

I'd rather make it optional, as it is:
- only used in `check`
- still a sticky wicket on pyodide, due to the ever-popular `pyrsistent`, which might impair our ability to use it for self-hosting in the browser... i'd wager the version that doesn't use it probably doesn't support draft07...

We could add an extra up-front step of just installing the dev package, try to `build`...

bollwyvl:
Over on #184, I added `jsonschema` as an optional dependency that throws warnings if missing... was actually useful to have it fail on PyPI :scream:, as it reminded me to use `--no-index`, etc. when smoke testing.

jtpio:
> Over on #184, I added `jsonschema` as an optional dependency that throws warnings if missing.

Ah nice, thanks! This looks good so we can close this one in favor of #184.

PR #184, opened 2021-06-29
improve python distribution artifacts

### LINKED ISSUE #183: Missing README on the PyPI page
`jupyterlite` is now available on PyPI: https://pypi.org/project/jupyterlite/

However the page would look nice with the readme.

image

We could for example use the top-level readme file.

bollwyvl:
## References

- fixes #183

## Code changes

- [x] adds `dist:twine` doit tasks which runs `twine check` on each ~~dist~~ wheel
  - > `flit` sdists don't pass `twine check` :crying_cat_face: 
- [x] patch up the metadata
  - [x] `description-file`
  - [x] `classifiers`
  - [x] `license`
  - [x] `keywords` 
- [x] make `jsonschema` an optional dependency
  - [x] throw some `warnings` but don't fail 
- [x] copies the root README to `py/jupyterlite` before building
  - [x] adds a short table about installing, using the CLI
- [x] smoke tests the CLI on each platform _before_ installing test dependencies
  - > `linux` and `osx` build the same `sha256sum`, a la :tada: https://github.com/jtpio/jupyterlite/pull/172#issuecomment-867625254 

## User-facing changes

- users should see something on PyPI :blush:

- n/a

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -26,7 +26,7 @@ integration with other Jupyter tools.
 To get the [Python CLI](./cli.ipynb) and [API](./api/index.md) from [PyPI]:
 
 ```bash
-pip install jupyterlite
+python3 -m pip install --pre jupyterlite
 # TODO: mamba install jupyterlite
 ```
 
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -26,7 +26,7 @@ integration with other Jupyter tools.
 To get the [Python CLI](./cli.ipynb) and [API](./api/index.md) from [PyPI]:
 
 ```bash
-python3 -m pip install --pre jupyterlite
+python -m pip install --pre jupyterlite
 # TODO: mamba install jupyterlite
 ```
 
--- a/py/jupyterlite/README.md
+++ b/py/jupyterlite/README.md
@@ -37,20 +37,22 @@ archive.
 python3 -m pip install --pre jupyterlite
 ```
 
-Use the [`jupyter lite` CLI][cli] to `build`, `check`, or create a reproducible,
-remixable `archive` of your site, then [deploy] your built site to any static host, such
-as GitHub Pages or ReadTheDocs.
+Use the [`jupyter lite` CLI][cli] to `build`, `check`, or create a [reproducible],
+remixable `archive` of your site, then [publish] your built site to any static host,
+such as GitHub Pages or ReadTheDocs.
 
 | `jupyter lite` | description                                         | extras                                |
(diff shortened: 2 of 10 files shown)


bollwyvl:
Ha, turns out I wrote the test poorly. Also, `flit` sdists are apparently flagged as non-comforming by `twine check`... i'm _really_ hoping it pulls it off the wheel... i shan't be writing any rst today, thanks.

bollwyvl:
I appear to have gummed up the RTD works, but otherwise this is good to go

py/jupyterlite/README.md
```diff
@@ -1,58 +1,118 @@
...
+Install `jupyterlite` from PyPI, which comes with the CLI and a pre-built, empty site
+archive.
 
 ```bash
-# TBD pip install jupyterlite
-# or...
-# TBD mamba install -c conda-forge jupyterlite
-# or...
-# TBD conda install -c conda-forge jupyterlite
+python3 -m pip install --pre jupyterlite
```
jtpio, line 37: Minor preference for dropping the `python3` for conciseness:

```suggestion
pip install --pre jupyterlite
```

docs/deploying.md
```diff
@@ -26,7 +26,7 @@ integration with other Jupyter tools.
 To get the [Python CLI](./cli.ipynb) and [API](./api/index.md) from [PyPI]:
 
 ```bash
-pip install jupyterlite
+python3 -m pip install --pre jupyterlite
```
jtpio, line 29: Probably some distributions are still on Python 2 by default and require the explicit `python3`?

Otherwise I have a minor preference for just `python` since this is user-facing:

```suggestion
python -m pip install --pre jupyterlite
```

jtpio:
Just pushed a minor change to use `python` instead of `python3` for the user facing `pip` commands.

jtpio (approved):
Thanks!

bollwyvl:
Yeah, I prefer python better as well, but ran into at least one problem
during that mess of commits. Also apparently some distros are dropping the
`python` alias, which seems kinda crazy to me but I don't make the rules. I
guess Ubuntu 20.04 will be with us for a good while.

I'm fine with changing it back in docs, but would keep in ci... And
probably didn't even do it in dodo...

jtpio:
> I'm fine with changing it back in docs, but would keep in ci... And probably didn't even do it in dodo...

Yes it's just for docs and README.md, since we rarely use `python3` for user facing docs these days.

PR #190, opened 2021-06-29
Handle copying multiple themes

jtpio:
## References

As noticed in https://github.com/jtpio/jupyterlite/pull/180, the `jupyter lite build` would fail when multiple themes are installed in the virtual environment, as the corresponding doit tasks would define the same targets.

## Code changes

Update the theme copy task.

## User-facing changes

Be able to use multiple federated themes.

None


bollwyvl:
Good find! ☺️

jtpio:
Will rebase https://github.com/jtpio/jupyterlite/pull/180 right after which should make the build pass.

PR #205, opened 2021-07-01
Remove the custom theme plugin

### LINKED ISSUE #193: Reuse the upstream theme plugin?
We should check whether we can just reuse the upstream plugin instead of having our own here.

We currently define a custom plugin here:

https://github.com/jtpio/jupyterlite/blob/c8cbb818598435d5fb101ac0cfda29e135388fe4/packages/theme-extension/src/index.ts#L20-L89

The upstream plugin is defined here:

https://github.com/jupyterlab/jupyterlab/blob/8c15f776dd7fcd221169d282d23eaa1eb27e23a5/packages/apputils-extension/src/themesplugins.ts#L33-L153

Probably this was to be able to define the custom themes url, but maybe that is already configurable.

_Originally posted by @jtpio in https://github.com/jtpio/jupyterlite/pull/180#discussion_r661246909_

jtpio:
## References

Fixes #193 

Remove the custom theme plugin, and reuse the one from upstream directly.

## Code changes

The custom theme plugin was coming from the early iterations on the p5 notebook, to handle themes differently (bundling light and dark by default instead of loading them on the fly).

## User-facing changes

None

None


bollwyvl:
More reuse good!

As pertains to federate-all-the-things on #118: might this end up coming back as `@jupyterlite/themes` which is just an `index.ts` and `plugin.ts` which re-exposes everything from `@jupyterlab/themes`, with the federated junk in `package.json`?

jtpio:
> As pertains to federate-all-the-things on #118: might this end up coming back as `@jupyterlite/themes` which is just an `index.ts` and `plugin.ts` which re-exposes everything from `@jupyterlab/themes`, with the federated junk in `package.json`?

Probably yes. Not sure if or when we'll reach this point (everything federated), so I guess we can still remove it in the meantime.

jtpio:
Hmm looks like RTD doesn't find the css files:

image

Seems to be good locally though after a clean.

jtpio:
Ah looks like there is an extra `en/205/_static` in the URL.

jtpio:
> Ah looks like there is an extra `en/205/_static` in the URL.

Probably because the upstream plugin joins the base and themes URLS:

https://github.com/jupyterlab/jupyterlab/blob/c957e3f2b9a31f819a28a99cbbbfd1d08abaea21/packages/apputils-extension/src/themesplugins.ts#L50

bollwyvl:
Do we need to special-case this in `config-utils.js`?

bollwyvl:
> Do we need to special-case this in config-utils.js?

Looked good locally, so i pushed :blush:

jtpio:
Ah thanks!

Sounds like we can indeed special-case this at least for now. Maybe there will be other [`JupyterFontEnd.IPaths`](https://github.com/jupyterlab/jupyterlab/blob/c957e3f2b9a31f819a28a99cbbbfd1d08abaea21/packages/application/src/lab.ts#L226-L240) to consider later on.

jtpio:
Put it back to draft as Retro doesn't seem to pick it up on rtd.

bollwyvl:
It's a bit hard to track where things are after the force-pushes, but it appears the last thing i had on a620e5be was

```ts
if (k === 'themesUrl') {
  // themesUrls is joined in code with baseUrl, leave as-is
  continue;
}
config[k] = `${urlBase}${v.slice(2)}`;
```

bollwyvl:
As to other uses of `URLExt.join(baseUrl, ...)`: it appears _I_ did when adding the [license stuff](https://github.com/jupyterlab/jupyterlab/blob/dbdefed9db9332381fe4104bdf53ec314451951f/packages/help-extension/src/index.tsx#L469-L473). Otherwise it's all kernel assets.

jtpio:
Ah thanks. Looks like I missed the sync with the remote branch before rebasing.

jtpio:
cc @bollwyvl if you want to have a look

Looks like retro also needed this: https://github.com/jtpio/jupyterlite/pull/205/commits/0b58f6a652d4d72aafa4001efc63b756c80be812

app/config-utils.js
```diff
@@ -206,6 +206,10 @@ export function fixRelativeUrls(url, config) {
   let urlBase = new URL(url || here()).pathname;
   for (const [k, v] of Object.entries(config)) {
     if (k.endsWith('Url') && v.startsWith('./')) {
+      if (k === 'themesUrl') {
```
bollwyvl, line 209: This was kind of a fast hack, but maybe we should hoist this to `UNPREFIXED_PATHS`?

app/lab/package.json
```diff
@@ -178,7 +176,6 @@
   "jupyterlab": {
     "extensions": [
       "@jupyterlite/application-extension",
-      "@jupyterlite/theme-extension",
```
bollwyvl, line 181: does `@jupyterlab/theme-extension` need to added to this list? the singleton list?

app/webpack.config.js
```diff
@@ -120,24 +120,6 @@ const extensionAssetConfig = Build.ensureAssets({
...
-  path.resolve(buildDir, 'schemas/@jupyterlab/apputils-extension', 'themes.json'),
-  path.resolve(jupyterliteApputilsPlugin, 'themes.json'),
-  { overwrite: true }
-);
-
-fs.copySync(
-  path.resolve(buildDir, 'schemas/@jupyterlab/apputils-extension', 'package.json.orig'),
-  path.resolve(jupyterliteApputilsPlugin, 'package.json.orig')
-);
-
```
bollwyvl, line 140: :tada: ideally we'd get all this stuff into a plugin so it can be managed by webpack more directly, e.g. use `CleanWorkspacePlugin`.

app/config-utils.js
jtpio: Yes it's probably cleaner in the long run.

app/lab/package.json
jtpio: Or `@jupyterlab/apputils`? Since the upstream plugin comes from:

https://github.com/jupyterlab/jupyterlab/blob/14abba086b5cae2ad0bcfaf0f5e08a7dc1ca5678/packages/apputils-extension/src/themesplugins.ts#L37

bollwyvl:
There appears to be much less theme leakage, so am very much :+1: to merging this.

Maybe the FLOUC is a bit longer now in retro, but i must confess i haven't been monitoring it.

I can handle hoisting `UNPREFIXED_PATHS=['licensesUrl', 'themesUrl']` when i get to it... which i should issue anyway.

Bottom line: I don't see any showstoppers. :ship:

app/config-utils.js
bollwyvl: Like i said below, i'll hit it on #233

app/lab/package.json
bollwyvl: hey, it builds! just looking at diffs

app/config-utils.js
jtpio: :+1:

jtpio:
Fixed a couple of conflicts (the latest `main` is now on alpha 4)

PR #200, opened 2021-06-30
Resolve absolute path of `out_dir`

benabel:
Fix #196.

Issue #185, opened 2021-06-29
broken doit?


madhur-tandon:
## Description

Running `doit` used to build the setup locally. However, with the latest `main` branch, it gives
```
ERROR: Invalid parameter: "schema". Must be a command, task, or a target.
Type "doit help" to see available commands.
Type "doit list" to see available tasks.
```

Running `doit build` does the job though.

bollwyvl:
Moved schema to lint... You can see all the stuff it will do with:

doit list --all --status

jtpio:
Probably the default `doit` tasks can run the tasks like before, so it's consistent with the docs: https://jupyterlite.readthedocs.io/en/latest/contributing.html#task-and-action-defaults

(or we update the docs)

bollwyvl:
oh, snap, yeah, that's my bad on not updating the `default_tasks`. We should add the `docs:app:build` to it as well. PR incoming...

PR #206, opened 2021-07-01
move more path logic to traitlets

bollwyvl:
## References

- further refines solution to #185, as provided by #200 

## Code changes

- [x] cargo cult `TypedTuple` from `ipywidgets.trait_types`
- [x] makes all path-based config traits `CPath` or `Tuple(CPath())`
- [x] does more `.resolve`ing 

## User-facing changes

- n/a

- n/a


jtpio (approved):
Thanks!

Issue #132, opened 2021-06-03
Problem Saving Notebook


psychemedia:
Trying to save a created and untitled notebook in JupyterLab view, the notebook appears to save under a newly given name, but the dialog remains with the original default name and blocks further action:

![image](https://user-images.githubusercontent.com/82988/120677289-d43b7a80-c48e-11eb-9c95-90e04df0285f.png)

UPDATE: clicking the *Enter* button lots of times very very quickly appears to clear the dialog.

jtpio:
Thanks @psychemedia for reporting :+1: 

It might be an issue with JupyterLite, but could also be from upstream JupyterLab. For reference this feature was added in https://github.com/jupyterlab/jupyterlab/pull/10043.

We're using the very latest packages (`3.1.0-alpha.11` at the moment), and I believe there are still a couple of things upstream to fix with the new rename flow.

jtpio:
For reference, https://github.com/jtpio/jupyterlite/pull/136 disabled this dialog on the demo RTD site.

psychemedia:
I've just noticed another problem with erratic save behaviour whilst running `jupyerlite` using a simple pyhton http server (`python3 -m http.server`): specificially, the save behaviour seems very hit and miss. Sometimes saves appear to occur, sometimes they don't. I'm seeing the effect with both manual and autosaves. 

The notebook download option appears to work correctly, which means I can:

- edit a notebook and see the changes;
- download the changed notebook correctly;
- close and open the notebook and the changes have disappeared,

At other times, changes do appear to be saved persistently but I can't figure out when.

jtpio:
> specifically, the save behaviour seems very hit and miss

Would you be able to try with stock JupyterLab? Using the latest alpha release with `pip install --pre jupyterlab`, just to check whether the same happens there (which looks like it does)

psychemedia:
I think it's Chrome.. Just noticed that my current version of Chrome seems to be completely broken cachewise trying to refresh other pages: things not refreshing at all (Firefox sees updated links fine). Not sure if this a Chrome version thing, or a side effect of the large number of open tabs (but they shouldn't have side effects, right?) With browsers continually updating themselves, it's hard to know what's going on at times..

bollwyvl:
Over on #173, reworked some  of the contents stuff... maybe try it out on RTD?

https://jupyterlite--173.org.readthedocs.build/en/173/

PR #204, opened 2021-07-01
Add tbump configuration to bump the jupyterlite Python package

jtpio:
## References

A small incremental step towards better release scripts #121 

## Code changes

Config

## User-facing changes

None

## Backwards-incompatible changes

None

PR #198, opened 2021-06-30
Add `jupyterlab-kernelspy` to the demo site

jtpio:
## References

Fixes #197 

## Code changes

Demo site update.

## User-facing changes

Users of the demo site should be able to inspect kernel messages.

## Backwards-incompatible changes

None

Issue #197, opened 2021-06-30
Add `jupyterlab-kernel-spy` to the demo site


jtpio:
On that end, we should probably consider adding [kernelspy](https://github.com/jupyterlab-contrib/jupyterlab-kernelspy) to the demo site, and ensuring we generate content there, as losing the browser websocket playback, which _just_ got good enough to use, is pretty crippling when you're building something and _it goes wrong_.

_Originally posted by @bollwyvl in https://github.com/jtpio/jupyterlite/issues/195#issuecomment-871411572_

jtpio:
Conda forge PR: https://github.com/conda-forge/staged-recipes/pull/15456

PR #207, opened 2021-07-02
Upgrade to JupyterLab 3.1.0rc1

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlite/bump-lab-beta-0?urlpath=lab)

## References

- started from #198 (of #197)

## Code changes

- [x] bump all the (upstream) versions The Hard Way
  - not sure i got them all :blush:
  - don't know if there's a more automated way for everything a la #204)
    - ... but maybe as a datapoint for tracking stuff down?
- [x] add `appName` and `appVersion` to `jupyter-lite(.schema).json`
  - [x] validate in `dodo`
  - [ ] consider version of last resort from a `package.json` if available? `yarn.lock` for better reproducibility?
- [x] implement `setPath` in `packages/application-extension/src/index.tsx` 
  - > didn't have time to go run this one down, yet
  - [x] consider deferring in #109?

## User-facing changes

- _About_ now shows app name and version

Screenshot from 2021-07-02 20-51-00

[previous screenshot](https://user-images.githubusercontent.com/45380/124219917-f4c91580-daca-11eb-95ad-eeba16f746c6.png)


jtpio:
Thanks!

> don't know if there's a more automated way for everything

There is the `update:dependency` script from upstream, for example:

```bash
yarn run update:dependency --regex @jupyterlab next
```

Although we also need to fix the `resolutions` with the new versions after.

.binder/environment.yml
```diff
@@ -14,12 +15,13 @@ dependencies:
...
   # demo stuff
   - jupyter-server-proxy
-  - jupyterlab >=3.1.0a11,<4
   - retrolab
   # extra docs tools
   - sphinx-autobuild
 
   ### DOCS ENV ###
+  # primary upstream
+  - jupyterlab >=3.1.0a11,<4
```
jtpio, line 24: Looks like this could also be bumped?

```suggestion
  - jupyterlab >=3.1.0b0,<4
```

py/jupyterlite/src/jupyterlite/__init__.py
```diff
@@ -1,3 +1,3 @@
 """tools for building JupyterLite sites
 """
-__version__ = "0.1.0a2"
+__version__ = "0.1.0a3"
```
jtpio, line 3: How about leaving that bump (and the JS ones) for a future release commit? Like https://github.com/jtpio/jupyterlite/commit/e6a20fbbb76fba4e7d87c6204e99914c42458763

At some point I'm hoping we'll get all that nicely automated with a script doing the version bumps for us.

.binder/environment.yml
bollwyvl: noice

py/jupyterlite/src/jupyterlite/__init__.py
bollwyvl: i generally like the practice of bumping the version as soon as the tag goes out... but in this case... it's already done, as i was running into cache issues (see flailing commit spam) that went away after rebuilding. happens upstream, too, sometimes.

bollwyvl: though maybe going to the upstream `b` should trigger a higher bump? it's all kinda funny with pre-releases, anyway...

jtpio: > i generally like the practice of bumping the version as soon as the (previous) tag goes out.

Sure. We can probably do that for the next release? So releasing `0.1.0a3` and then bump to `0.1.0a4` after?

Although it might be a matter of tastes and preferences, but having a separate commit on `main` is also nice to make it easier to see when things got bumped (instead of having them "hidden" in a PR).

jtpio: > though maybe going to the upstream `b` should trigger a higher bump? it's all kinda funny with pre-releases, anyway...

Right. I guess we could also stay on alpha here for now.

bollwyvl: says your name on the tin, so you drive... I can bump everything down if that will work better with the automation! been meaning to nuke my miniforge from orbit for a while... one errant `pip install --pre` in the wrong env and _who knows_ what I'm running anymore :rofl:

bollwyvl: ok... version what _we_ ship hopefully reverted by be4bb5a

bollwyvl:
From checking out on binder:

Screenshot from 2021-07-02 21-19-24

probably not a blocker, as we still haven't fleshed out the content for #31... and I'm sure we'll get around to either: 

- upgrading the tour upstream  https://github.com/jupyterlab-contrib/jupyterlab-tour/issues/39
  - potentially fixing whatever regression occurred _further_ upstream in lab core
    - it kinda seems like this might be the case, as i'd think it's kind of a semver fail to expect every extension developer to update for a `3.1` release.

jtpio:
> * as i'd think it's kind of a semver fail to expect every extension developer to update for a `3.1` release.

Agree that 3.1 shouldn't cause a downstream update. 

Wondering if this is caused by the new menu backed by settings feature, which was added in the latest alpha. I remember seeing something similar when trying to update RetroLab: https://github.com/jupyterlab/retrolab/pull/167

jtpio:
Just released [RetroLab `0.3.0b0`](https://github.com/jupyterlab/retrolab/releases/tag/0.3.0b0), so we can also update the `@retrolab` packages to `0.3.0-beta.0`.

bollwyvl:
I haven't had time to run down more extension incompatibilities... i think we probably want to wait until some of those are resolved vs core before throwing too much effort at it.

jtpio:
Looks like we should be able to finish this one once a `b2` is out for lab.

bollwyvl:
Is a `b2` even forthcoming?

jtpio:
> Is a `b2` even forthcoming? asked on [gitter](https://gitter.im/jupyterlab/jupyterlab?at=60e7166b515cab3e4fa7e68c)

Let's see, but there have already been a couple of merges so it could be worth it.

bollwyvl:
rc1 coming today https://github.com/jupyterlab/jupyterlab/issues/10513#issuecomment-878214972 :tada:

jtpio:
Nice!

Planning to cut a new `0.1.0a5` today with the recent changes, so we can update this PR and merge it right after.

bollwyvl:
Will start once this clears:

https://github.com/conda-forge/jupyterlab-feedstock/pull/282

jtpio:
Sounds good :+1: 

Will be updating `@retrolab` too.

bollwyvl:
Kinda gave up trying to do it the right way, and fell back to mostly manually merging.

The `fasta` file example doesn't load, but the mimetype does. Otherwise, everything else is GTG in lab.

bollwyvl:
(didn't look at rtc, the relationship to the upstream implementation is... vague to me)

examples/overrides.json
```diff
@@ -13,8 +13,5 @@
         "options": {}
       }
     ]
-  },
-  "@jupyterlab/docmanager-extension:plugin": {
```
bollwyvl, line 17: this "feature" was reverted :tada:

app/lab/package.json
```diff
@@ -160,7 +161,6 @@
     "jest": "^26.4.2",
     "mini-css-extract-plugin": "~0.9.0",
     "npm-run-all": "^4.1.5",
-    "playwright": "^1.8.0",
```
bollwyvl, line 163: as we're not using this yet, might as well save the 1gb...

app/lab/package.json
```diff
@@ -254,7 +254,8 @@
       "@lumino/virtualdom",
       "@lumino/widgets",
       "react",
-      "react-dom"
+      "react-dom",
+      "yjs"
```
bollwyvl, line 258: just copying upstream...

examples/overrides.json
jtpio: This should fix #132.

jtpio:
This also fixes #210 (bringing https://github.com/jupyterlab/jupyterlab/pull/10572):

image

packages/application-extension/src/index.tsx
```diff
@@ -45,10 +45,14 @@ import React from 'react';
 
 class WebRtcProvider extends WebrtcProvider implements IDocumentProvider {
   constructor(options: IDocumentProviderFactory.IOptions & { room: string }) {
-    super(`${options.room}${options.guid}`, options.ymodel.ydoc);
+    super(`${options.room}${options.path}`, options.ymodel.ydoc);
     this.awareness = options.ymodel.awareness;
   }
 
+  setPath() {
+    // TODO: this seems super useful
```
jtpio, line 53: This was added in https://github.com/jupyterlab/jupyterlab/pull/10470 to support renaming of files.

It would be fine to tackle this separately as it might require a different logic than upstream (we don't have that common websocket endpoint)

jtpio:
> (didn't look at rtc, the relationship to the upstream implementation is... vague to me)

RTC seems to be working fine on the preview.

For the usernames and colors, we'll need to implement something similar to what is upstream since we replace the provider: https://github.com/jupyterlab/jupyterlab/pull/10411

jtpio:
Looks like there is an issue with markdown cells going out of focus, which looks like it should also be happening upstream (need to check there):

https://user-images.githubusercontent.com/591645/125589002-b3be049b-b755-46a2-aadd-6835ebd3e32b.mp4

jtpio:
> which looks like it should also be happening upstream (need to check there):

Happens upstream too: https://github.com/jupyterlab/jupyterlab/issues/10619

packages/application-extension/src/index.tsx
bollwyvl: hm... is a no-op fine, or do we need some kind of message?

bollwyvl: > For the usernames and colors, we'll need to implement something similar to what is upstream since we replace the provider: jupyterlab/jupyterlab#10411

Sounds fun. Alas, even though Callisto is big enough to have moons, we haven't found any. Maybe names of asteroids? Or entirely made-up moons from various sci-fi universes?

So do we collect all of the RTC issues into one follow-on?

jtpio: > So do we collect all of the RTC issues into one follow-on?

Yes better to fix them outside of this PR. The username / color should be straightforward normally.

jtpio:
Jut pushed the update to the latest `@retrolab/*`.

Will do a quick check on RTD, and then we should be good to go with this one.

jtpio (approved):
Thanks!

jtpio:
@bollwyvl looks like this is ready to go?

bollwyvl:
Seems :+1: to me

bollwyvl:
You want me to punch it?

jtpio:
Let's get it in :rocket:

PR #214, opened 2021-07-06
Refactor the kernel implementation

### LINKED ISSUE #211: Refactor kernel's executeResult and executeError handling
### Problem

As discovered in @madhur-tandon 's IPython PR (https://github.com/jtpio/jupyterlite/pull/171), the `executeResult` is handled in the base kernel [here](https://github.com/jtpio/jupyterlite/blob/ad01d63864a46a9166632cdf73e676490d5e0c50/packages/kernel/src/kernel.ts#L345), but handling `executeResult` and `executeError` should be kernel specific. For example, when using IPython in pyolite, we'll let IPython's display hook trigger the `executeResult`.

We need to refactor this a bit.

You can assign this task to me if you want.

martinRenou:
Fix #211 

The base kernel used to handle the "execute_result" and "error" iopub messages automatically, we now let each kernel implementation handle those messages.

Also, the `executeRequest` BaseKernel's method was expecting an `executeResult` type instead of an `executeReply` as return value.

### DIFF
--- a/packages/javascript-kernel/src/kernel.ts
+++ b/packages/javascript-kernel/src/kernel.ts
@@ -92,17 +92,41 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
    */
   async executeRequest(
     content: KernelMessage.IExecuteRequestMsg['content']
-  ): Promise<KernelMessage.IExecuteResultMsg['content']> {
+  ): Promise<KernelMessage.IExecuteReplyMsg['content']> {
     const { code } = content;
-    const result = this._eval(code);
-    // TODO: move executeResult and executeError here
-    return {
-      execution_count: this.executionCount,
-      data: {
-        'text/plain': result
-      },
-      metadata: {}
-    };
+    try {
+      const result = this._eval(code);
+
+      this.publishExecuteResult({
+        execution_count: this.executionCount,
+        data: {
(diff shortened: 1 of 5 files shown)


packages/javascript-kernel/src/kernel.ts
```diff
@@ -92,17 +92,41 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
    */
   async executeRequest(
     content: KernelMessage.IExecuteRequestMsg['content']
-  ): Promise<KernelMessage.IExecuteResultMsg['content']> {
+  ): Promise<KernelMessage.IExecuteReplyMsg['content']> {
     const { code } = content;
-    const result = this._eval(code);
-    // TODO: move executeResult and executeError here
```
martinRenou, line 98: Actually fixing this TODO, nice!

martinRenou:
Also renaming `executeResult` to `publishExecuteResult`, this is a personal preference (what we do in xeus), but don't hesitate to ask me to revert this change

jtpio:
Thanks!

> Also renaming `executeResult` to `publishExecuteResult`

Sounds good :+1:

jtpio (approved):
Thanks!

martinRenou:
Thank you!

PR #216, opened 2021-07-06
ipython refactor

madhur-tandon:
1. `run` and `complete` methods are now a part of `Pyolite` class / kernel.
2. reply is now constructed at the end for all messages of type `reply` (except for callbacks inside `execute`)
3. `get_parent` is implemented to access parent headers which enables the use of `Output` widget.

### DIFF
--- a/packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
+++ b/packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
@@ -27,15 +27,11 @@ class Pyolite:
 
         txt, matches = self.interpreter.complete("", line, line_cursor)
         return {
-            "parentheader": self._parent_header,
-            "type": "reply",
-            "results": {
-                "matches": matches,
-                "cursor_end": cursor_pos,
-                "cursor_start": cursor_pos - len(txt),
-                "metadata": {},
-                "status": "ok",
-            },
+            "matches": matches,
+            "cursor_end": cursor_pos,
+            "cursor_start": cursor_pos - len(txt),
+            "metadata": {},
+            "status": "ok"
         }
 
     async def run(self, code):
@@ -62,8 +58,4 @@ class Pyolite:
             results["evalue"] = last_traceback["evalue"]
             results["traceback"] = last_traceback["traceback"]
 
-        return {
-            "parentheader": self._parent_header,
-            "type": "reply",
-            "results": results,
-        }
+        return results
--- a/packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
+++ b/packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
@@ -34,7 +34,7 @@ class Pyolite:
         }
 
     async def run(self, code):
-        self._last_traceback = None
+        self.interpreter._last_traceback = None
         exec_code = self.interpreter.transform_cell(code)
         await _load_packages_from_imports(exec_code)
         if self.interpreter.should_run_async(code):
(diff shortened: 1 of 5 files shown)


packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
```diff
@@ -15,3 +17,30 @@ def comm_info(self, target_name=""):
                 comms[comm_id] = dict(target_name=comm.target_name)
 
         return comms
+
+    def do_complete(self, code, cursor_pos):
```
martinRenou, line 21: Maybe we can now rename it to `complete`, just like we have `run`?

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
```diff
@@ -15,3 +17,30 @@ def comm_info(self, target_name=""):
...
+        self.interpreter._last_traceback = None
+        exec_code = self.interpreter.transform_cell(code)
+        await _load_packages_from_imports(exec_code)
+        if self.interpreter.should_run_async(code):
+            self.result = await self.interpreter.run_cell_async(
+                code, store_history=True
+            )
+        else:
+            self.result = self.interpreter.run_cell(code, store_history=True)
+        return self.result
```
martinRenou, line 46: Your `do_complete` returns a message that respects the kernel protocol, maybe we could move some logic that you have in the `worker.ts` here, to return the proper protocol message from here in Python.

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
madhur-tandon: Doing.

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
madhur-tandon: so instead of constructing a `reply` on the JS side, we return the `reply` on the python side?

martinRenou: Yes. Well, it's just that because you do it for the `complete`, it seems more homogeneous to me to do it for the `run` as well. Also this makes the code closer to ipykernel, and could potentially help us to refactor to share more code.

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
```diff
@@ -15,3 +18,52 @@ def comm_info(self, target_name=""):
...
+                code, store_history=True
+            )
+        else:
+            self.result = self.interpreter.run_cell(code, store_history=True)
+
+        results = {}
+        results["payload"] = self.interpreter.payload_manager.read_payload()
+        self.interpreter.payload_manager.clear_payload()
+
+        if self.interpreter._last_traceback is not None:
```
martinRenou, line 56: `if self.interpreter._last_traceback is None:`

martinRenou (changes requested):

packages/pyolite-kernel/src/worker.ts
```diff
@@ -327,26 +309,27 @@ self.onmessage = async (event: MessageEvent): Promise<void> => {
 
   const messageType = data.type;
   const messageContent = data.data;
+  const parentHeader = data.parentHeader;
```
martinRenou, line 312: Maybe we could do `kernel._parent_header = data.parentHeader` here instead, and not change the signature of `execute`, `complete` etc

madhur-tandon: Done!

martinRenou (approved):
LGTM! If everything still works as expected, it's a 👍🏽 for merging from me

madhur-tandon:
Output Widget now works too!

Screenshot 2021-07-09 at 4 40 06 PM

jtpio:
Neat, thanks Madhur!

Probably we could even add the `Output` widget to the example widget notebook (could be done later of course)

jtpio (approved):
Thanks!

PR #217, opened 2021-07-06
Skip serve tests on MacOS CI

### LINKED ISSUE #215: macos CI failing
## Description

Looks like CI is now failing on macos: https://github.com/jtpio/jupyterlite/runs/2998239544

Started on this commit: https://github.com/jtpio/jupyterlite/commit/99dcbe7e445901bd09b34b7f0a19ca13af37312a

This is also happening on new PRs; https://github.com/jtpio/jupyterlite/pull/214/checks?check_run_id=2998598392

bollwyvl:
## References

- fixes #215

## Code changes

- skip `jupyter lite serve` tests on MacOS... but only when running in CI
  - if it's a real problem, maybe some helpful contributor running on real MacOS will get frustrated and fix it :blush: 

## User-facing changes

- n/a

- n/a

### DIFF
--- a/py/jupyterlite/src/jupyterlite/tests/test_serve.py
+++ b/py/jupyterlite/src/jupyterlite/tests/test_serve.py
@@ -1,5 +1,6 @@
 """Test that various serving options work"""
 
+import os
 import subprocess
 import sys
 import time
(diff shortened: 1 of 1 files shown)


bollwyvl:
First fail was missing the coverage target... i like keeping tests under coverage to find dead code, but in this case...

anyhow, should be good to go now.

jtpio (approved):
Thanks!

Make sense to skip on macos since they indeed look flaky.

PR #219, opened 2021-07-06
Add ipympl example

martinRenou:
https://user-images.githubusercontent.com/591645/124771250-4514fd00-df3b-11eb-834b-96d32aec99a7.mp4

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -51,6 +51,7 @@ dependencies:
   - ipywidgets
   - bqplot
   - ipyleaflet
+  - ipympl
   - ipycanvas
   - jupyterlab-drawio
   - jupyterlab-tour >=3.1.0
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -38,6 +38,7 @@ dependencies:
   - ipywidgets
   - bqplot
   - ipyleaflet
+  - ipympl
   - ipycanvas
   - jupyterlab-drawio
   - jupyterlab-tour >=3.1.0
--- a/examples/ipympl.ipynb
+++ b/examples/ipympl.ipynb
@@ -1,0 +1,135 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ec40ad8e",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import micropip\n",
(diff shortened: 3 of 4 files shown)


martinRenou:
It seems to work only with this ugly workaround for now:
```python
import sys
import types

sys.modules['tornado'] = types.ModuleType('tornadofake')
```

Tracking this in ipympl directly: https://github.com/matplotlib/ipympl/issues/334

jtpio:
> It seems to work only with this ugly workaround for now:

It would be totally fine to have that in the examples for now.

Or maybe we should even bake that in https://github.com/jtpio/jupyterlite/blob/main/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py directly?

martinRenou:
Yes! I can do that :)

bollwyvl:
Not to pick on this PR (which is _raaaad_), but could we, to the extent possible, be more intentional about some of the imports hacks so that these things run under `nbconvert --execute` and `nbval`? e.g.

```py
try:
    import pyolite
    # do hacky things
except ImportError:
    pass
```

It's also worth adding a `<details><summary>Limitations</summary>...</details>` markdown cell above any cells like these to all of our examples. Totally aside, but dang i sure wish there was a good markdown flavor that supported `details`...

martinRenou:
> be more intentional about some of the imports hacks so that these things run under nbconvert --execute and nbval?

Yeah! I guess all the `micropip.install` commands should go there as well, for all notebooks?

jtpio:
> Yeah! I guess all the `micropip.install` commands should go there as well, for all notebooks?

Maybe, at the expense of making the code less readable?

> be more intentional about some of the imports hacks so that these things run under `nbconvert --execute` and `nbval`?

Maybe we could track this in a separate issue. For now the example notebooks are really meant to be used in JupyterLite so they could probably stay as they are?

jtpio:
@martinRenou should we go ahead and merge this PR? I guess we can always go back to it later once https://github.com/matplotlib/matplotlib/pull/20591 is fixed

martinRenou:
If there is no other blocker, yes :)

jtpio:
I've added a screencast to the top comment for reference.

jtpio (approved):
Thanks!

PR #226, opened 2021-07-07
Add patch for PIL.Image.Image._repr_png_

### LINKED ISSUE #223: PIL Image does not display
## Description

PIL Image should display, but does not.

## Reproduce

Small reproducible example:

```python
from PIL import Image
Image.new("RGB", (100, 100))
```

Gives this output:

Screenshot from 2021-07-06 15-45-59

which shows that the error is:

```html
<img src="data:image/png;base64,[object Object]">
```
but it should be:

Screenshot from 2021-07-06 15-50-19

I'd be glad to look at this, but not sure where to begin. Maybe a missing tostring() on Bytes?

dsblank:
## References

This fixes #223 by add a monkey patch to pyolite.py. A future fix would be to find out where bytes need to be encoded from the JavaScript form, and fix there.

## Code changes

Added patch, and call it.

## User-facing changes

PIL Images will now display correctly.

None


bollwyvl:
Thanks for this!

Hm, it looks like we do already call [`transform_cell`](https://github.com/jtpio/jupyterlite/blob/2ff4cf17d57a514704b68a51fe05695ff888ac12/packages/pyolite-kernel/py/pyolite/pyolite/interpreter.py#L60) which in turn calls the same-named function on [InputTransformer](https://github.com/ipython/ipython/blob/7.25.0/IPython/core/inputtransformer2.py#L534). I'd rather see us figure out how to start doing this the right way than adding to every kernel startup another 3mb Pill plus libjpeg, etc... 

I think an example notebook would be a fine place to demonstrate it, with the ugly in-line hack, and then removing the hack can be a victory lap.

jtpio:
Thanks @dsblank!

Looking good on the RTD preview: https://jupyterlite--226.org.readthedocs.build/en/226/

image

> I think an example notebook would be a fine place to demonstrate it, with the ugly in-line hack, and then removing the hack can be a victory lap.

So should we focus on the example notebook first? I haven't checked the added overhead introduced by the new PIL import, but if it's significant then we should probably start thinking of a different approach indeed.

examples/pyolite - images.ipynb
```diff
@@ -0,0 +1,8 @@
+{
+  "metadata": {
+    "orig_nbformat": 4
+  },
+  "nbformat_minor": 4,
+  "nbformat": 4,
+  "cells": []
```
jtpio, line 7: Looks like the notebook is empty?

dsblank: Indeed... looks like a bug in jupyterlite that when I save the notebook and download it, it is empty.

dsblank: Manually recreated the notebook, as there doesn't seem a way to make a new notebook (with contents) from jupyterlite.

jtpio: Ah looks like this might be a bug then. Would you like to open a new issue to track this?

Normally the download command from the command palette should work:

image

jtpio: Might be related: https://github.com/jtpio/jupyterlite/issues/229

dsblank: There is an issue in opening a file in the notebook. I'll take a look at that. Perhaps there is a standard way of opening files in jupyterlite?

jtpio: By opening a file, do you mean something like in https://github.com/jtpio/jupyterlite/discussions/91?

dsblank: From the notebook:

Screenshot from 2021-07-07 11-03-43

dsblank: BTW, PIL is already imported, so this additional patch is not adding any additional time to the current load time. PIL is import via matplotlib. So this is going to be great to have a better system!

Re: moving to a lazy import-patching system, there are at least two options:

1. write the bare minimum necessary ourselves for our minimal needs
2. use a package like [importhook](https://github.com/brettlangdon/importhook) that appears to work, is tested, and maintained

What would you like to pursue?

bollwyvl: importhook looks nice and has no dependencies. a perusal of some of the crazy stuff needed in the example notebooks shows some of the challenges, where to get to something functioning in `sys.modules`, we need to (for now) to install particular versions of upstreams.

Looking at those, we might need something that lets us codify `preimport` (and `preinstall`) hooks. However, as we don't have persistence in the WASM filesystem, these are the same... but we're working on that upstream.

The `preinstall` might be solveable through a patch of `micropip` (which _could_ be done via `importhook`). 

I feel like the long con would be for site builders to be able to add required patches via config file, as their end users shouldn't even know they're doing anything special (aside from _perhaps_ a context manager or a magic).

jtpio: > BTW, PIL is already imported, so this additional patch is not adding any additional time to the current load time. PIL is import via matplotlib. So this is going to be great to have a better system!

Thanks @dsblank

So we can probably go ahead with this patch and try to improve things in follow-up issues and PRs?

examples/pyolite - images.ipynb
```diff
@@ -0,0 +1,97 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "115e688f-9a77-422b-9801-485ba4c7401b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "logo = Image.open(\"jupyter.png\")\n",
```
jtpio, line 71: Maybe we can remove that once since reading from existing files from Python is not supported yet?

Or we generate / write an image to the Python file system, and then open it with PIL?

dsblank: I'll make loading a file work; I just didn't know what was wrong :) So if I save an image, and then load it, it should work?

bollwyvl: or use `await js.fetch`... a `py - js.ipynb` that collects tricks for that would be rad... the `micropip` implementation does some [pithy stuff for working with binary downloads](https://github.com/pyodide/pyodide/blob/main/packages/micropip/micropip/micropip.py#L38).

We should probably have a notebook just for js-in-py-js tricks like this until we figure out how to e.g. make stdlib (or `requests`, `httpx`, etc) work.

jtpio: >  So if I save an image, and then load it, it should work?

I think it should yes

jtpio:
btw @dsblank for reference Nick has been making good progress on persistence in upstream Pyodide: https://github.com/pyodide/pyodide/pull/1596

So hopefully this might land in JupyterLite some time soon when we update to the new Pyodide (#178)

bollwyvl:
> persistence in upstream Pyodide

Yeah, it's a slow ship to turn, but making progress.

Having a persistent `whl` cache will be _wonderful_, and we'll turn that on as soon as it's available... this will lessen (but not eliminate) the CDN hits.

But regarding contents: I'm not sure if _magically_ bidirectionally syncing everything from the composite server/localforage contents into every kernel runtime is going to be feasible... however, if we implement `IDrive`, and the lite kernel API had an optional `getDrive` hook, each virtual computer could be mounted, and one could drag/drop files in and out.

We could actually try this out _today_, without persistence (as the _does_ require recompiling the pyodide universe) as there is a hidden `_module.FS` member hanging off the API

examples/pyolite - images.ipynb
dsblank: > So we can probably go ahead with this patch and try to improve things in follow-up issues and PRs?

Sounds like a plan. I updated the notebook to save and then load, and that appears to work well. So, I think this PR is complete. 

I'll open up another issue so we can discuss the patching issue. 

Thanks all!

dsblank:
Very cool, @bollwyvl ! That will benefit the whole pyodide ecosystem!

jtpio (approved):
Thanks!

jtpio:
> I'll open up another issue so we can discuss the patching issue.

@dsblank that would be great, thanks!

Issue #67, opened 2021-04-26
Deployment Documentation


bollwyvl:
### Problem

It's still a little unclear how one would actually deploy and "own" jupyterlite.

### Proposed Solution

Add a _Deployment_ section which includes:
- how to configure
  - where to put a `jupyterlite.json` 
  -  JSON Schema 
- how to get the code
  - we could be shipping the tarball to the docs site, which is as good as any, for now
- configuring different SaaS providers
  - vercel
  - github pages
  - readthedocs
- on-premises
  - nginx?
  - inside other host apps
    - jupyterhub
    - jupyter_server (#41)
      - jupyter-server-proxy 
    - django?
    - fastapi?
    - flask?
  - how to configure CDN replacements (#45)

### Additional context

- #30 probably required before this is actually important

jtpio:
Dumping some for Vercel in this issue:

- Create an account
- Run `vercel` from the repo locally to get the secrets, see https://github.com/amondnet/vercel-action#how-to-use for some details
- Add the action to the repo (see this repo, or add code snippet)
- There should be a comment on the next PR

bollwyvl:
Started a skeleton (lots of TBD) for this on #58:
https://github.com/jtpio/jupyterlite/pull/58/files#diff-27ccda254fe500e13705e63554cdf4dcaf4c8b8e2fc74024f61776d407a30c56R1

areeves87:
Very excited about this project. Had a lot of fun trying the jupyterlite demo notebooks out today!

I want to share an interactive jupyerlite notebook-style report with my colleagues in the near future. Will it be possible to provide a step-by-step guide for how to publish a jupyterlite notebook to amazon s3? Since s3 hosts static web pages and supports `application/wasm` content I suppose it's possible to do. But I haven't had time to figure out what exactly are the steps required.

bollwyvl:
Yeah, there are a lot of distracting directions (e.g. #141) right now, but getting a minimal CLI out is probably more important. Once written, It's going to be much easier to keep _that_ documented in a semi-automatic way, e.g. `jupyter lite --help`, rather than a bunch of narrative docs. 

Once that works, it will be:

```bash
pip install jupyterlite <any labextensions you want>
jupyter lite init ./my-jupyterlite
cp <your notebooks> ./my-jupyterlite/files
jupyter lite build ./my-jupyterlite
```

[...]

PR #228, opened 2021-07-07
Add docs to deploy on Netlify

jtpio:
## References

Consolidate the documentation about deployments: #67 

I quickly tried deploying to Netlify using a similar structure as for the demo repo: https://github.com/jupyterlite/demo

And it seems to be working so far, as long as we specify 3.7 as the Python runtime.

image

## Code changes

Documentation

## User-facing changes

None

None

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -225,7 +225,7 @@ automatically get a preview link when opening a new pull request:
 
 ### Netlify
 
-[Netlify](https://www.netlify.com/) makes it easy and convenient to host static website
+[Netlify](https://www.netlify.com/) makes it easy and convenient to host static websites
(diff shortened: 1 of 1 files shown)


jtpio:
This would be very similar for Vercel except they seem to be stuck on Python 3.6 at the moment...

bollwyvl:
We can make the CLI py-3 compatible... just a few inline type annotations not even being checked with a typechecker.

jtpio:
Right, although if it's just for Vercel we might want to wait for them to support newer version. Looks like it it's mentioned in https://github.com/vercel/vercel/discussions/5486

jtpio:
Since this is additive to the documentation, let's merge that and iterate if needed.

PR #234, opened 2021-07-08
Add route for licenses, include licenses in app bundle

### LINKED ISSUE #233: Show bundled JS libraries, versions, and licenses
### Problem

As a JupyterLite site builder or end user, it is pretty hard to know what software are in my final, shipped site.

### Proposed Solution

- enable the [license file webpack plugin](https://github.com/jupyterlab/jupyterlab/blob/v3.1.0b1/jupyterlab/staging/webpack.prod.config.js#L18) 
- add `licensesUrl` to `jupyter-config-data`? (it doesn't load the UI if this is missing)
- add a serverlite route for `/api/licenses`
- implement a license bundle finder
- enable the `@jupyterlab/help-extension:licenses` plugin (if not already done)

### Additional context

Somewhere in the lab 3.1 alpha progression, we got `/api/licenses` into `jupyterlab_server` (and some helper consumer stuff in `jupyterlab`) which reveals the precise name, version, license, and full license text of the core bundle and all federated extensions (there are none in the wild yet) from `(app_dir/static/|labextensions/(*/)?*/)third-party-licenses.json`.

A number of changes occurred along the way, as well as a few regressions, so we might have to wait for #207  `b2` (with the menu fix) but otherwise should be fairly straightforward.

bollwyvl:
## References

- [x] fixes #233
- [x] blocked by #207 (which in turn is blocked by an upstream release of `rc0`)
- [x] will conflict with #205

## Code changes

- [x] add `JSONLicenseWebpackPlugin` to webpack
- [x] add `licensesUrl` to schema 
- [x] add `licensesUrl` to `jupyter-lite.json` 
- [ ] adds server route
  - [x] skeleton handler
  - [x] fetch `third-party-licenses.json` from server 
    - [x] core
    - [x] federated extensions
      - [x] add `the_smallest_extensions`, build with license-aware `@jupyterlab/builder`
        - [x] remove checked-in fixtures
        - [x] build in CI 
      - [x] improve types for `federated_extensions`
      - [x] hoist to `@jupyterlite/types`
      - [x] add fetcher
  - [ ] bundle for JSON
  - [ ] downloads?

## User-facing changes

- the _Licenses_ menu will appear, and show the grid (once it works)

Screenshot from 2021-07-13 21-11-38

- n/a


bollwyvl:
Buh, that was gross. Sorry for all the commit spam.

So the challenge we now have is that the _name_ of the owning package of the license bundle is _not in_ the license bundle. However, as we probably want the version-of-last-resort from `package.json` over in #207, maybe we can source it from there...

bollwyvl:
So this looks okay locally, but still has a fair amount to do.

I don't think the report download is super worth it, so might try to figure out how to hide those... or disable the commands or something.

Getting an actual federated extension, built with a `@jupyterlab/builder` that will include the data is... annoying. I guess that "simplest" one we have checked in is what it's going to be... maybe i add `left-pad` as a dependency?

bollwyvl:
Well, another run might with a warm cache might drag the time down a bit, but at first blush, the conda-build step takes 30s longer than all the other build stuff :crying_cat_face: . Can probably cache the built artifacts, if I can convince myself that given `yarn.lock`, `package.json` and `setup.cfg` and all the ts/py we'd get the same wheel?conda package out... but at least i can now start testing against actual built licenses...

bollwyvl:
Yeah, i think caching the heavy fixtures is going to work fine.

bollwyvl:
Ha, looks a little janky with just two deps:

Screenshot from 2021-07-23 08-52-37

Points to a few issues:
- upstream, can we get `style-loader` and `css-loader` out of every extension, ideally without every extension having to configure the peril-fraught `sharedPackages`
- guess i should fix the upstream CSS... somehow :blush:

bollwyvl:
Going to continue on #431.

Issue #165, opened 2021-06-22
Add server(lite)/labextension for building sites.


bollwyvl:
### Problem

#147 adds the beginning of the CLI, but this is still not super friendly. Supporting the same options in a labextension setting, with live preview, would showcase the tool.

### Proposed Solution

- add a labextension 
  - add an (extensible) editor that can
    - edit `jupyter-lite.json`
    - edit `jupyter-lite.ipynb` metadata
  - add a button that builds a site with...
- add a serverextension that shares the `LiteManager` in `jupyter_server`
  - builds sites, serves them (in-loop)
  - offers archives for download

### Additional context

While it should probably not constrain the initial implementation, it would likely be possible to either reuse the python implementation, or re-implement portions in JS as a serverlitextension. From this, and a listing of all files used in a site (such as the `SHA256SUMS`), its conceivable we could get to a self-hosting, live-remixing environment, where a user could:

- visit the docs site
- write a notebook or two
- fiddle with some settings
- download a new archive of the site
  - use it locally a bit
  - save it again
- send the archive to someone else
  - they could drag it into the docs site
  - see everything the other person had chosen to add

Crazy!

bollwyvl:
Over on #238, i committed a wheel and conda package for testing purposes... they're both about 20k. We should probably put the source in as a standalone package, so that we can adapt the fixtures as needed.

bollwyvl:
I've added the `extension idea` label, as perhaps it _would_ make sense to get this started out-of-band... but it could be that we _should_ keep it in this repo, at least as long as the schema/config is still somewhat in flux.

On that note, now that `rjsf` is in upstream, I think we could go ahead and start doing this.

Issue #151, opened 2021-06-17
Better manage the extra Python packages


jtpio:
As a follow-up to #145, we should consider improving the way extra dependencies are handled for the Python kernel.

There is some metadata here:

https://github.com/jtpio/jupyterlite/blob/5f38f5cfbf373ad2c8a6c4bfd4f2d54124c36525/packages/pyolite-kernel/package.json#L66-L73

And the wheels are referenced here: 

https://github.com/jtpio/jupyterlite/blob/5f38f5cfbf373ad2c8a6c4bfd4f2d54124c36525/packages/pyolite-kernel/src/kernel.ts#L11-L14

Ideally there should be a single source of truth and a more streamlined way for managing these extra packages.

bollwyvl:
I'm also passing the buck from #147 on federated-extensions-by-wheel here, as these approaches likely needs to work together _and_ be user-serviceable.

I'm leaning towards a `requirements.txt`-compatible approach, but don't know how one would differentiate between wheels meant to meant to be:
- served raw to pyodide
- unpacked and have `data_files` extracted

bollwyvl:
With #238 in which handles wheel/conda packages for non-kernel-specific extensions, and hopefully close to merging, we have:

- a cache dir
- some basic download tooling

I think to finish out this one, and take some strides towards #225, some of the pyolite issues can start to be addressed.

This would do some new stuff during the lifecycle:
- `post_init`: copy our pre-built patched wheels for `ipy*` to the `{output_dir}/pyolite/wheels` 
- `pre_build`: copy user-specified local/downloaded wheels to same
- `post_build` create a cut-down `https://pypi.org/pypi/jupyterlite/json` in, i guess, `{output_dir}/pyolite/api/{pkg}.json`, for each of the packages, 

This would require patching `micropip` to be aware of multiple sources of files, effectively additional indexes (which we could support at the same time, though I don't know if everyone's custom registry supports the JSON API... so maybe the old-school `https://pypi.org/simple/` approach is better).

#45 is harder. The distribution stance for `pyodide` (196mb off github actions) is still a bit further off, as we'd want to see what 0.18 does, as well as how "normal means" distribution through `npm` works. If it ends up that just the stack up to `micropip` is available as a single package, we would probably be able to ship that on `pip`.

bollwyvl:
I don't think we can do much more here!

PR #238, opened 2021-07-09
Support more sources of federated_extension

bollwyvl:
## References

- tangentially addresses #151
  - though that's more about the py-in-js packages for `pyolite`
    - but the `.whl` downloading (which definitely _should_ be shared between them) is reusable (and really dumb)

## Code changes

- [x] add a `--cache-dir` (default: `{lite_dir}/.cache`)
  - for downloading archives
  - ideally would be shareable between multiple builds
- [x] adds to `LiteBuildConfig`
  - [x] `--ignore-sys-prefix` to disable the env scraping stuff 
  - [x] `--federated-extensions`, which accepts
    - [x] local and remote `.whl` files
    - [x] local folders that contain a `package.json` 
    - [x] conda `.tar.bz2`
      - [x] local
      - [x] remote 
      - [x] ~~`anaconda.org`~~ doesn't work because of redirects, not going to add special case
    - [ ] other well-known archive formats/layouts? 
      - I don't want to go mucking around too much...
- [x] use `--federated-extensions` in docs build  for ~~_all_~~ of the extensions
  - [x] will make RTD ~~_much_~~ somewhat faster
  - [x] keep the goods in binder
  - [x] cache the downloaded conda packages, rather than trying to find them
- [x] update docs
- [x] add tests (currently fails coverage threshold :crying_cat_face:)
  - [x] checked-in smallest extension (which has a theme, mime, settings, etc)
  - [ ] ideally we'd already be _building_ a labextension to test with, (#165), but will probably just download a fixture, pull it out of the local wheel cache, etc.

## User-facing changes

- CLI users will have more options, especially in CI, for builds

- renamed a few members on the addons... no too worried about breakage at that layer at this time


bollwyvl:
Of note, a conda-forge URL can be be predictably transformed from:

```http
https://conda.anaconda.org/conda-forge/noarch/jupyterlab-gt-coar-theme-0.3.0-pyhd8ed1ab_0.tar.bz2
```

to:

```http
https://github.com/conda-forge/releases/releases/download/noarch/jupyterlab-gt-coar-theme-0.3.0-pyhd8ed1ab_0.tar.bz2/jupyterlab-gt-coar-theme-0.3.0-pyhd8ed1ab_0.tar.bz2
```

> yerp, `releases/releases` and `.bz2/....bz2` :shrug: 

This URL is _probably_ throttled, though... but _maybe_ we could offer it as a translation.

bollwyvl:
My previous [build](https://readthedocs.org/projects/jupyterlite/builds/14200068/) for 24763bc took:
- 792 seconds

Imma guess... ~740 :crossed_fingers: :popcorn:

bollwyvl:
Welp, it failed... but it failed _faster_! ` 636`

Was:

```
[rtd-command-info] start-time: 2021-07-09T14:13:01.498605Z, end-time: 2021-07-09T14:13:54.219586Z, duration: 52, exit-code: 0
```

Now:

```
[rtd-command-info] start-time: 2021-07-09T03:54:47.647757Z, end-time: 2021-07-09T03:56:02.804552Z, duration: 75, exit-code: 0
mamba env create --quiet --name 238 --file docs/environment.yml
```

So _maybe_ not worth the complexity. It kinda blows me away that they implemented the mamba thing as _use conda to install mamba_ when they could roll a container:

```
[rtd-command-info] start-time: 2021-07-09T03:51:57.916166Z, end-time: 2021-07-09T03:54:47.494956Z, duration: 169, exit-code: 0
conda install --yes --quiet --name=base --channel=conda-forge python=3.7 mamba
```

bollwyvl:
Gah, apparently the docs notebooks `import pandas` (which is no longer installed)

bollwyvl:
So, actually _passing_ it takes 767s.

Thirty seconds isn't nothing, of course... and every time we add another extension, there's the risk of it driving it up further.

Very possibly we could run more `doit` jobs at once (except for `build:prod`) and knock some more time off.

Also the reproducibility check (and the archive creation in the first place) is pretty much vanity (and help catch errors in a visual way) but these are checked in the per-platform tests, anyway. We can probably take them out (or gate them behind an env var check, so we're at least running them locally. Similarly, the demo app archive that gets built isn't _linked to_ from anywhere in the docs... so maybe we drop that, too, and just keep it in the release workflow.

Of course, making these _faster_ with an optional `libarchive-c` dependency might be better... while folk won't be _building_ archives all the time, they will be unpacking that 20mb every time.

bollwyvl:
RTD builds now seem to hover right around 720-740 ms, which seems like it's probably worth the changes.

I've got `the-smallest-extension` locally, and could add it to `packages` (or a new top-level/nested folder) and squash out the checked-in `.whl` and conda `.tar.bz2`... even with `micromamba` or whtatever, adding `conda-build` ratchets up the complexity/time/flakiness pretty substantially, but could be a parallel concern that _only_ worries about that... and would cache quite well.

Otherwise, this is probably good to go.

jtpio:
Thanks!

Looks like there is a minor conflict to fix, but otherwise looks good :+1:

jtpio:
Arf, one more conflict so we can add `ipycytoscape` (from #256) to the demo site.

bollwyvl:
will try resolving with the web ui...

bollwyvl:
Yeah, `ipyctoscape` and its deps are not exactly lightweight... `Build took 813 seconds`...

jtpio:
> Yeah, `ipyctoscape` and its deps are not exactly lightweight... `Build took 813 seconds`...

Maybe we could start offloading some of these dependencies and examples to https://github.com/jupyterlite/demo?

And keep https://jupyterlite.github.io/demo/ as the "stable" demo, and https://jupyterlite.rtfd.io/en/latest/try/lab  be the "nightly". This could help keep snappier previews here for development.

jtpio (approved):
Thanks!

Issue #235, opened 2021-07-08
Lazy Patching


dsblank:
### Problem

Currently, jupyterlite needs to tweak code from some Python libraries. Currently, there are [two patches](https://github.com/jupyterlite/jupyterlite/blob/main/packages/pyolite-kernel/py/pyolite/pyolite/patches.py) to make these tweaks:

* `matplotlib.pyplot.show`
* `PIL.Image.Image._repr_png_`

Eventually we may be able to remove these changes, but, for now, we need to monkey patch them. Currently we import the libraries, and change the methods. This is not ideal because it forces these libraries to be imported, even if they are not used, creating a large startup delay.

### Proposed Solution

A better solution would be to only patch them when we import the code. There are at least two possibilities for monkey patching on import:

* write a custom code loader
* use a existing library, such as [importhook](https://github.com/brettlangdon/importhook)

@bollwyvl said:

> `importhook` looks nice and has no dependencies. a perusal of some of the crazy stuff needed in the example notebooks shows some of the challenges, where to get to something functioning in sys.modules, we need to (for now) to install particular versions of upstreams.

> Looking at those, we might need something that lets us codify `preimport` (and `preinstall`) hooks. However, as we don't have persistence in the WASM filesystem, these are the same... but we're working on that upstream.

> The `preinstall` might be solveable through a patch of `micropip` (which could be done via `importhook`).

> I feel like the long con would be for site builders to be able to add required patches via config file, as their end users shouldn't even know they're doing anything special (aside from perhaps a context manager or a magic).

See also:

1. #227

bollwyvl:
As of #913, we would no longer have _any_ heavyweight patches :tada: !

Issue #45, opened 2021-04-15
Allow/default to disabling all CDN operations


bollwyvl:
Thanks again for getting this going.

As I've hinted at over on #41, a huge benefit of static sites is that they can be, well _static_. By this I mean: once you get a build working and put someplace, it will _work_, provided some future browser hasn't disabled a critical API, even if that _someplace_ is a USB stick or DVD in a computer that can't access the internet.

I don't think it's reasonable in the long term to trust any of the major CDNs to necessarily be around in five years, much less bespoke ones like mathjax and pyodide. Further, CDNs are a source of lapses in user privacy and can be abused/disabled to affect harm/censorship by any number of agents.

I would be tempted to give a pass to content-addressable systems (e.g. git, fossil, bittorent, hypercore, ipfs) that can be verifiably cached, locally, and differentiably updated, but for now I think part of being able to fully realize the potential of this system is its ability to generate archival-grade content that can be used without access to The Internet, even if that isn't the _default_ mode of operation.

## The Challenges

Right now, this looks like:
- mathjax: 
  - understandable
    - there is (at least) [jupyter_server_mathjax](https://github.com/jupyter-server/jupyter_server_mathjax)
  - I feel like lab core or server should probably just suck it up and distribute it
    - but failing that, another, _another_ copy of mathjax in a distributed package a la #41 isn't going to kill anybody's internet
- pyodide: 
  - sticky indeed!
    - `import pandas, matplotlib` actually only pulls 32mb of data, which honestly is within the range of "sane," i'd say, to distribute as an sdist... and even a `whl`, if it doesn't get duplicated.
      - over on `ipydrawio`, i've done some of the [dirty work](https://github.com/deathbeds/ipydrawio/blob/master/packages/ipydrawio-webpack/lib/_static.js) to get path-perfect things in place with `!!file-loader!`, so it's definitely doable to make something that quacks like their CDN.

Once it _is_ working, "offline," keeping it that way is always going to be a challenge... I've had some luck with gross `/etc/hosts` stuff in VM testing, but perhaps there are better, browser-based ways to do it now.

jtpio:
Thanks!

Right this should be able to run offline. Especially when publishing assets via GitHub releases (https://github.com/jtpio/jupyterlite/issues/16). Folks should just be able to download the tarball, run `python -m http.server` and have things work out of the box.

[...]

PR #239, opened 2021-07-09
Use importhook for patches

dsblank:
Use importhook to lazily patch needed tweaks.

## References

Possible fix for #235 

## Code changes

Instead of import and patch, use importhook for setting patch on import.

## User-facing changes

No longer need to load all of matplotlib (and all dependencies) on first execution.

None.

### DIFF
--- a/packages/pyolite-kernel/src/worker.ts
+++ b/packages/pyolite-kernel/src/worker.ts
@@ -24,6 +24,7 @@ async function loadPyodideAndPackages() {
     import micropip
     await micropip.install([
       'traitlets',
+      'importhook',
       '${_widgetsnbextensionWheelUrl}',
       '${_nbformatWheelUrl}',
       '${_ipykernelWheelUrl}'
--- a/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
+++ b/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
@@ -19,12 +19,7 @@ sys.modules["resource"] = types.ModuleType("resource")
 # This is needed for some Matplotlib backends (webagg, ipympl)
 sys.modules["tornado"] = types.ModuleType("tornados")
 
-from .patches import ensure_matplotlib_patch, ensure_pil_patch
-
-# apply patches for available modules
-ensure_matplotlib_patch()
-ensure_pil_patch()
-
+from .patches import register_patches
 from .display import LiteStream
 from .interpreter import LitePythonShellApp
 
@@ -38,3 +33,5 @@ kernel_instance = ipython_shell.kernel
 
 sys.stdout = stdout_stream
 sys.stderr = stderr_stream
+
+register_patches()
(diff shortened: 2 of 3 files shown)


dsblank:
This isn't quite right. There is an error in accessing matplotlib.pyplot. But if anyone has comments on the structure so far, I'd be glad for any advice.

dsblank:
@bollwyvl @jtpio I believe that this is now correct and ready for review. This PR reduces the number of python modules loaded from 937 down to 665 (all of the modules loaded by the patches: matplotlib and PIL and their dependencies) and seems to cut ready-to-use time by about 50%. To test, try the image and matplotlib notebooks.

bollwyvl:
hmmm... i'm still seeing matplotlib and numpy and pillow and kiwi and... load on an empty kernel launch on [the pr docs](https://jupyterlite--239.org.readthedocs.build/en/239/_static/lab/index.html).

Screenshot from 2021-07-10 10-15-13

We're also still pulling `pytest` and `nose`, which seems odd... but maybe `micropip` just grabs all the extras...

dsblank:
@bollwyvl Thanks for testing! Yes, I see those too, but I think that is separate from the import into the actual kernel sys.modules. The patching (and now the importhooks) are done when the kernel starts. I suspect what you are looking at is a deeper level that just gets those packages ready to be imported (aka, micropip.installs). I don't see pytest or nose in sys.modules. Also my regular JupyterLab IPython sys.modules has 756 entries, so I think this PR is properly limiting the imports.

bollwyvl:
Welp, baby steps then.

Of note: some stuff that messes with `sys.meta_path` just landed upstream in `pyodide`:

https://github.com/pyodide/pyodide/blob/main/src/py/_pyodide/_importhook.py

bollwyvl:
So this is good for stuff we know about, (eventually) test, and ship.

How might we allow other packages to create patches like this?

One approach might be `entry_point` metadata, which can be created by any of the (sane) package build chains. We'd then want to use these ourselves

```toml
# pyproject.toml
[tool.flit.entrypoints."pyolite.v0.import_patches"]
matplotlib_what_it_does = "pyolite.patches.matplotlib:patch_matplotlib_what_it_does"
pillow_what_it_does = "pyolite.patches:patch_pillow_what_it_does"
```

We would then want to hook the end of each `micropip.install` to apply these, have to be careful about not double applying them, etc.

bollwyvl:
Ah, found where all the [matplotlib stuff](https://github.com/jupyterlite/jupyterlite/blob/52fad689c18144032853fedc345741c62670e350/packages/pyolite-kernel/src/worker.ts#L22
) is coming from.

It would be interesting to see the impacts of cutting out the early imports of everything except the bare minimum, `ipykernel`, `traitlets`, `ipython`.

We'd probably have to continue to "take over" more `micropip` behavior, and short-circuit requests for our shipped wheels, but I guess it's all a lot of trades between:
- time-to-execute-response
- parity with canonical upstream APIs
- maintanability/resilience in the face of upstream API changes at the pyodide and jupyter levels

jtpio (approved):
Thanks!

packages/pyolite-kernel/src/worker.ts
```diff
@@ -24,6 +24,7 @@ async function loadPyodideAndPackages() {
     import micropip
     await micropip.install([
       'traitlets',
+      'importhook',
```
jtpio, line 27: We can track how to best handle these extra packages in https://github.com/jupyterlite/jupyterlite/issues/45#issuecomment-878103407, so they can be configurable and not prevent offline use.

Issue #194, opened 2021-06-30
ESLint allocation failure errors


jtpio:
## Description

This seems to be happening more often lately.

Sometime running ESLint automatically on commits might trigger the following error:

```
<--- Last few GCs --->

[950841:0x55dffd287ed0]    25909 ms: Mark-sweep (reduce) 2046.1 (2051.7) -> 2045.4 (2053.0) MB, 815.8 / 0.0 ms  (average mu = 0.132, current mu = 0.008) allocation failure scavenge might not succeed
[950841:0x55dffd287ed0]    27405 ms: Mark-sweep (reduce) 2046.5 (2052.0) -> 2045.7 (2053.0) MB, 1487.0 / 0.0 ms  (average mu = 0.053, current mu = 0.006) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x7fe812c92489 node::Abort() [/home/jtp/miniforge3/envs/jupyterlite-dev/bin/../lib/libnode.so.83]
 2: 0x7fe812ba5dbe std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > node::SPrintFImpl<char const*>(char const
```

## Reproduce

It might be possible to reproduce by just running `doit lint` at the top-level.

## Expected behavior

ESLint shouldn't crash.

## Context

Probably we are missing some ignore clauses in the `.eslintignore`?

jtpio:
Seeing that one locally even when committing a single file: https://github.com/jupyterlite/jupyterlite/blob/main/packages/application-extension/src/index.tsx

Could be because this one is a `.tsx`.

bollwyvl:
So setting `DEBUG='*' doit lint` shows... everything. There's some graphviz in there. Last time a spent a bit with this, it seemed to reliably fail on initializing typescript. Similar to the webpack fail, we can try `--cache`, and _hope_ that if it OoMs, it will be able to pick up where it left off, but with slightly more done.

Getting to where we can use clean-webpack-plugin would probably _also_ help, as i just don't know whether spoor from multiple builds is causing issues (we ignore `**/build/**` but :shrug: )

jtpio:
Seeing that one less often now with #240, although the eslint check still takes quite a bit of time on commit.

jtpio:
[...]

PR #240, opened 2021-07-09
Update ESLint dependencies

jtpio:
## References

Attempt at fixing #194 by updating the ESLint dependencies.

Use `--cache` to hopefully mitigate the OoM issues.

## Code changes

Dependency update.

## User-facing changes

None

None


jtpio:
Even if this doesn't fix the memory issue in #194 and tests still pass, we can still do the update anyway.

jtpio:
The allocation failure seems to be reproducible even with this change:

```bash
# edit packages/application-extension/src/index.tsx

git add packages/application-extension/src/index.tsx
git commit
```

bollwyvl:
So can we... cache the cache in CI? Though maybe that's pointless as `lint-js-test` doesn't block anything else....

jtpio:
Checking a little bit locally with `--cache`. Looks like it might have improved a bit, haven't faced that oom yet.

> So can we... cache the cache in CI?

I guess we could?

PR #243, opened 2021-07-12
add inspect message

madhur-tandon:
Adds the ability to use the `inspect` message protocol in JupyterLite.

### DIFF
--- a/packages/pyolite-kernel/src/worker.ts
+++ b/packages/pyolite-kernel/src/worker.ts
@@ -222,7 +222,7 @@ function complete(content: any) {
  *
  * @param content The incoming message with the code to inspect.
  */
-function inspect(content: any) {
+function inspect(content: { code: string; cursor_pos: number; detail_level: 0 | 1 }) {
   const res = kernel.inspect(content.code, content.cursor_pos, content.detail_level);
   const results = formatResult(res);
   return results;
--- a/packages/pyolite-kernel/src/worker.ts
+++ b/packages/pyolite-kernel/src/worker.ts
@@ -218,6 +218,17 @@ function complete(content: any) {
 }
 
 /**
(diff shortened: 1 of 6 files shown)


madhur-tandon:
Screenshot 2021-07-12 at 4 18 55 PM

jtpio:
Nice, thanks Madhur!

jtpio:
Looks like there is a small conflict to resolve now that https://github.com/jupyterlite/jupyterlite/pull/239 is in.

Would you mind doing a quick rebase?

packages/pyolite-kernel/src/worker.ts
```diff
@@ -216,6 +216,17 @@ function complete(content: any) {
   return results;
 }
 
+/**
+ * Inspect the code submitted by a user.
+ *
+ * @param content The incoming message with the code to inspect.
+ */
+function inspect(content: any) {
```
jtpio, line 224: Looks like we could use a stronger type than `any` here?

And add `content` to the docstring above.

madhur-tandon: `content` is already in docstring. I have now added type definition.

bollwyvl:
Additionally, as you're closer to it, can you take a pass over the `docs/kernels` matrix and update with the features _actually_ supported? Probably need `comm` as well, now :laughing:. I kinda slammed out the original, and it's probably still kinda wrong. In an ideal world, we'd derive it (or more likely, check it, because no browsers are invited to RTD) from tags in #6, but for now it's all elbow grease...

jtpio (approved):
Looks good, thanks!

jtpio:
> can you take a pass over the `docs/kernels` matrix and update with the features _actually_ supported?

Yes that would be great :+1:  Opening a new issue to track this.

PR #245, opened 2021-07-12
add is_complete message protocol

madhur-tandon:


jtpio (approved):
Thanks!

packages/kernel/src/kernel.ts
```diff
@@ -512,6 +515,25 @@ export abstract class BaseKernel implements IKernel {
     this._sendMessage(message);
   }
 
+  /**
+   * Handle an is_complete_request message
+   *
+   * @param msg The parent message.
+   */
+  private async _is_complete_request(msg: KernelMessage.IMessage): Promise<void> {
```
jtpio, line 523: Minor nit-pick: since this part is in TypeScript we could follow the rest of the code use camel casing for the name of the method:

```suggestion
  private async _isCompleteRequest(msg: KernelMessage.IMessage): Promise<void> {
```

madhur-tandon: just fixed

jtpio:
Thanks!

PR #246, opened 2021-07-12
Add RELEASE.md

jtpio:
## References

Dumping some information about cutting new releases for reference.

This is still a manual process for now, but should improve with #121.

## Code changes

Dev docs.

## User-facing changes

None

None

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,52 @@
+# Releasing JupyterLite
+
+The current release process is still a bit manual, but should hopefully improve and be more automated soon.
+
+## Getting a clean environment
+
+To get a clean environment, the simplest is to follow the documentation for contributing:
+https://jupyterlite.readthedocs.io/en/latest/contributing.html
+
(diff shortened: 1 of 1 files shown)


bollwyvl:
I highly prefer releasing CI-built assets (even if CI isn't doing the actual permissioned upload), as it's almost impossible to screw up with a dirty env, and then the SHASUMS have a clear provenance to a particular build.

I also generally use `npm publish` (rather than `yarn` or `lerna`), and have even gone so far as to template out a release script .

jtpio:
Yes if we adopt the jupyter releaser then it will be releasing CI assets.

For now we create the GitHub release after pushing the tag to the repo with the workflow (will add that to RELEASE.md), so the steps are a bit out of order.

jtpio:
But we could already update the process to first push the release commit to generate the new assets on CI, and then download locally and publish them (for now it's just the PyPI package).

jtpio:
Should be good enough for now, to be improved and iterated on soon.

Issue #248, opened 2021-07-12
First cell execution hangs


davidbrochart:
## Description

Trying https://jupyterlite.rtfd.io/en/latest/try/lab, and opening `ipympl.ipynb`, executing the first cell hangs.

## Reproduce

1. Go to https://jupyterlite.rtfd.io/en/latest/try/lab
2. Open the `ipympl.ipynb` notebook
3. Execute the first cell
4. See error in the browser console: `RecursionError: maximum recursion depth exceeded while calling a Python object`

## Expected behavior

It should not hang.

## Context

- Operating System and version: Ubuntu 21.04
- Browser and version: Google Chrome version 91.0.4472.114 (Official Build) (64-bit)
- JupyterLite version: "Version unknown" (from Help -> About JupylerLite)

<details><summary>Browser Output</summary>
<pre>
jupyter-lite.ipynb:1 Failed to load resource: the server responded with a status of 404 ()
/en/latest/_static/jupyter-lite.ipynb:1 Failed to load resource: the server responded with a status of 404 ()
config-utils.js:200 failed get jupyter-config-data from https://jupyterlite.readthedocs.io/en/latest/_static/lab/jupyter-lite.ipynb
getLiteConfig @ config-utils.js:200
config-utils.js:200 failed get jupyter-config-data from https://jupyterlite.readthedocs.io/en/latest/_static/jupyter-lite.ipynb
getLiteConfig @ config-utils.js:200
remoteEntry.ca1efc27dc965162ca86.js:1 Unsatisfied version 3.1.0-alpha.11 of shared singleton module @jupyterlab/notebook (required ^3.0.0)
s @ remoteEntry.ca1efc27dc965162ca86.js:1
remoteEntry.ca1efc27dc965162ca86.js:1 Unsatisfied version 3.1.0-alpha.11 of shared singleton module @jupyterlab/logconsole (required ^3.0.0)
s @ remoteEntry.ca1efc27dc965162ca86.js:1
remoteEntry.ca1efc27dc965162ca86.js:1 Unsatisfied version 3.1.0-alpha.11 of shared singleton module @jupyterlab/settingregistry (required ^3.0.0)
s @ remoteEntry.ca1efc27dc965162ca86.js:1
remoteEntry.ca1efc27dc965162ca86.js:1 Unsatisfied version 6.1.0-alpha.11 of shared singleton module @jupyterlab/services (required ^6.0.0)
[...]

PR #249, opened 2021-07-12
Tweak `sys.setrecursionlimit()`

jtpio:
## References

Attempt at fixing #248


jtpio:
So we can test on the RTD preview with the example notebooks.

jtpio:
Bumping to 175 seems to be solving the `ipympl` / `micropip` calls, but breaks `altair`:

image

PR #250, opened 2021-07-12
Revert importhook for now

jtpio:
## References

Another attempt at fixing #248 by reverting https://github.com/jupyterlite/jupyterlite/pull/239

Closes #249

### DIFF
--- a/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
+++ b/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
@@ -19,9 +19,14 @@ sys.modules["resource"] = types.ModuleType("resource")
 # This is needed for some Matplotlib backends (webagg, ipympl)
 sys.modules["tornado"] = types.ModuleType("tornados")
 
+from .patches import ensure_matplotlib_patch, ensure_pil_patch
+
+# apply patches for available modules
+ensure_matplotlib_patch()
+ensure_pil_patch()
+
 from .display import LiteStream
 from .interpreter import LitePythonShellApp
-from .patches import register_patches
 
 stdout_stream = LiteStream("stdout")
 stderr_stream = LiteStream("stderr")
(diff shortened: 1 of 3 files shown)


jtpio:
It's possible the extra decorators add a couple more level of depths.

We can try on https://jupyterlite--89.org.readthedocs.build/en/250/_static/lab/index.html when the preview is ready.

jtpio:
Looks good on the preview with Chromium 91:

image

jtpio:
@davidbrochart would you be able to test with the preview? https://jupyterlite--250.org.readthedocs.build/en/250/_static/lab/index.html

jtpio:
cc @dsblank as you might have more insights on `importhook`

dsblank:
No, no real insights.... just watching and testing :)

davidbrochart:
> @davidbrochart would you be able to test with the preview? https://jupyterlite--250.org.readthedocs.build/en/250/_static/lab/index.html

Yes, it works fine.

jtpio:
Merging to fix #248 and get a release out.

@dsblank the work in #239 is great and works great in Firefox. Maybe we can try to get it back in right after and see how we can deal with the recursion limit issue with Chromium?

dsblank:
@jtpio Thanks! Can you say more about what exactly is happening, and why jupyterlite is setting the recursion limit? Is this an issue with history in jupyterlite?

jtpio:
@dsblank there was some investigation done by @lrowe in https://github.com/jupyterlite/jupyterlite/pull/113#issuecomment-851072065 which led to setting an arbitrary value for the recursion limit, so it also works with Altair.

Looks like it's a tricky one to get right. Maybe the next version of Pyodide will make that easier to deal with.

dsblank:
Wow, that is a tricky one, and explains the problem. I'll keep an eye on that so that we can bring back importhooks for lazy patching.

PR #253, opened 2021-07-13
add input_request message protocol

madhur-tandon:
This is not done yet, I need some sense of direction as to how to proceed further. Maybe links to some reference implementations?

Also, I changed the return type to `Promise<KernelMessage.IInputReplyMsg['content']>` instead of `Promise<void>`. I hope this is the way to go according to here: https://jupyter-client.readthedocs.io/en/stable/messaging.html#messages-on-the-stdin-router-dealer-channel

When this is done, it should allow us to use:
```
a = input()
```
in JupyterLite. (cc: @jtpio @martinRenou)

### DIFF
--- a/packages/javascript-kernel/src/kernel.ts
+++ b/packages/javascript-kernel/src/kernel.ts
@@ -201,9 +201,9 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
    *
    * @param content - The content of the request.
    */
-  async inputReply(
-    content: KernelMessage.IInputReplyMsg['content']
-  ): Promise<KernelMessage.IInputRequestMsg['content']> {
+  async inputRequest(
+    content: KernelMessage.IInputRequestMsg['content']
+  ): Promise<KernelMessage.IInputReplyMsg['content']> {
     throw new Error('Not implemented');
   }
 
--- a/packages/javascript-kernel/src/kernel.ts
+++ b/packages/javascript-kernel/src/kernel.ts
@@ -201,7 +201,7 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
    *
    * @param content - The content of the request.
    */
-  async inputReply(content: KernelMessage.IInputReplyMsg['content']): Promise<void> {
+  inputReply(content: KernelMessage.IInputReplyMsg['content']): void {
     throw new Error('Not implemented');
   }
 
--- a/packages/javascript-kernel/src/kernel.ts
+++ b/packages/javascript-kernel/src/kernel.ts
@@ -197,9 +197,9 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
   }
 
   /**
-   * Send an `input_request` message.
+   * Send an `input_reply` message.
    *
-   * @param content - The content of the request.
+   * @param content - The content of the reply.
    */
   inputReply(content: KernelMessage.IInputReplyMsg['content']): void {
     throw new Error('Not implemented');
--- a/packages/javascript-kernel/src/kernel.ts
+++ b/packages/javascript-kernel/src/kernel.ts
@@ -201,9 +201,7 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
    *
    * @param content - The content of the request.
    */
-  async inputRequest(
-    content: KernelMessage.IInputRequestMsg['content']
-  ): Promise<KernelMessage.IInputReplyMsg['content']> {
+  async inputReply(content: KernelMessage.IInputReplyMsg['content']): Promise<void> {
     throw new Error('Not implemented');
   }
 
--- a/packages/javascript-kernel/src/kernel.ts
+++ b/packages/javascript-kernel/src/kernel.ts
@@ -203,7 +203,7 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
    */
   async inputRequest(
     content: KernelMessage.IInputRequestMsg['content']
-  ): Promise<void> {
+  ): Promise<KernelMessage.IInputReplyMsg['content']> {
     throw new Error('Not implemented');
   }
 
--- a/packages/javascript-kernel/src/kernel.ts
+++ b/packages/javascript-kernel/src/kernel.ts
@@ -201,9 +201,9 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
    *
    * @param content - The content of the request.
    */
-  async inputRequest(
-    content: KernelMessage.IInputRequestMsg['content']
-  ): Promise<KernelMessage.IInputReplyMsg['content']> {
+  async inputReply(
+    content: KernelMessage.IInputReplyMsg['content']
+  ): Promise<KernelMessage.IInputRequestMsg['content']> {
     throw new Error('Not implemented');
   }
 
--- a/packages/kernel/src/kernel.ts
+++ b/packages/kernel/src/kernel.ts
@@ -195,9 +195,9 @@ export abstract class BaseKernel implements IKernel {
   ): Promise<KernelMessage.ICommInfoReplyMsg['content']>;
 
   /**
-   * Send an `input_request` message.
+   * Send an `input_reply` message.
    *
-   * @param content - The content of the request.
+   * @param content - The content of the reply.
    */
   abstract inputReply(content: KernelMessage.IInputReplyMsg['content']): void;
 
--- a/packages/kernel/src/kernel.ts
+++ b/packages/kernel/src/kernel.ts
@@ -196,7 +196,7 @@ export abstract class BaseKernel implements IKernel {
    *
    * @param content - The content of the request.
    */
-  abstract inputReply(content: KernelMessage.IInputReplyMsg['content']): Promise<void>;
+  abstract inputReply(content: KernelMessage.IInputReplyMsg['content']): void;
 
   /**
    * Send an `comm_open` message.
@@ -259,14 +259,7 @@ export abstract class BaseKernel implements IKernel {
    *
    * @param content The input_request content.
    */
-  protected inputRequest(
-    content: KernelMessage.IInputRequestMsg['content'],
-    parentHeader: any
-  ): void {
-    console.log('parent header from worker');
-    console.log(parentHeader);
-    console.log('this parent header');
-    console.log(this._parentHeader);
+  protected inputRequest(content: KernelMessage.IInputRequestMsg['content']): void {
     const message = KernelMessage.createMessage<KernelMessage.IInputRequestMsg>({
       channel: 'stdin',
       msgType: 'input_request',
--- a/packages/kernel/src/kernel.ts
+++ b/packages/kernel/src/kernel.ts
@@ -259,13 +259,16 @@ export abstract class BaseKernel implements IKernel {
    *
    * @param content The input_request content.
    */
-  protected _inputRequest(content: KernelMessage.IInputRequestMsg['content']): void {
+  protected inputRequest(
+    content: KernelMessage.IInputRequestMsg['content'],
+    parentHeader: any
+  ): void {
     const message = KernelMessage.createMessage<KernelMessage.IInputRequestMsg>({
       channel: 'stdin',
       msgType: 'input_request',
       // TODO: better handle this
-      session: this._parentHeader?.session ?? '',
-      parentHeader: this._parentHeader,
+      session: parentHeader?.session ?? '',
+      parentHeader,
       content
     });
     this._sendMessage(message);
--- a/packages/kernel/src/kernel.ts
+++ b/packages/kernel/src/kernel.ts
@@ -263,12 +263,16 @@ export abstract class BaseKernel implements IKernel {
     content: KernelMessage.IInputRequestMsg['content'],
     parentHeader: any
   ): void {
+    console.log('parent header from worker');
+    console.log(parentHeader);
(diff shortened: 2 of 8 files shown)


jtpio:
Thanks Madhur for starting this :+1: 

> This is not done yet,

Converting to a draft for now.

bollwyvl:
Cool stuff!

We have `kernelspy` in the binder environment... (it's the little JSON icon in the notebook toolbar) if you fire up an `ipykernel`, you'd be able to see the message exchange for something simple...

packages/pyolite-kernel/src/worker.ts
```diff
@@ -227,6 +246,20 @@ function inspect(content: { code: string; cursor_pos: number; detail_level: 0 |
   return results;
 }
 
+/**
+ * Deal with Inputs
+ *
+ * @param content The incoming message with the input prompt.
+ */
+function inputReq(content: { prompt: string; password: boolean }) {
```
martinRenou, line 254: This should be removed.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -309,6 +342,10 @@ self.onmessage = async (event: MessageEvent): Promise<void> => {
       results = await execute(messageContent);
       break;
 
+    case 'input-request':
```
martinRenou, line 345: This should be removed.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -100,6 +101,24 @@ async function sendComm(
...
+    prompt,
+    password
+  };
+  postMessage({
+    type: 'input_request',
+    parentHeader: formatResult(kernel._parent_header['header']),
+    content
+  });
+}
+
```
martinRenou, line 121: You are missing a function, let's call it `input`. It's the function you should set to the interpreter with `interpreter.input_request = input`.

This function might be tricky to implement, you should find a way for this to be a blocking call which resolves only when the input-reply has been received.

So:
- send input request to the `kernel.ts` with `sendInputRequest`
- wait for input reply. This may be done by waiting for a Promise that is resolved only when the `input-reply` is resolved. 
- return the content of the input reply to Python

martinRenou (changes requested):
The workflow when calling `input()` from Python is the following:

- worker (Python process) send input-request to the UI
- UI shows a text input that is "resolved" only when the user presses "enter". While it's waiting for user input, the worker should be blocked (waiting for reply)
- reply is sent to the worker, and the worker can continue execution

packages/pyolite-kernel/src/worker.ts
martinRenou: And replaced by `case 'input-reply'`, in this case resolve the Promise that your `input` function created so that it can continue

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -245,8 +245,8 @@ export class PyoliteKernel extends BaseKernel implements IKernel {
    */
   async inputRequest(
     content: KernelMessage.IInputRequestMsg['content']
-  ): Promise<void> {
-    throw new Error('Not implemented');
+  ): Promise<KernelMessage.IInputReplyMsg['content']> {
+    return await this._sendWorkerMessage('input-request', content);
```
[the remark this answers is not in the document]
madhur-tandon: What should be the arguments and return type of this?
Is the argument `KernelMessage.IInputRequestMsg['content']` and the return type `Promise<KernelMessage.IInputReplyMsg['content']>` OR perhaps this is the other way around?
Or should I revert back to what it was before i.e. the argument is `KernelMessage.IInputRequestMsg['content']` and the return type is `Promise<void>`?

packages/pyolite-kernel/src/worker.ts
madhur-tandon: Done.

packages/pyolite-kernel/src/kernel.ts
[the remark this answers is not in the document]
martinRenou: > Or should I revert back to what it was before i.e. the argument is KernelMessage.IInputRequestMsg['content'] and the return type is Promise<void>?

That seems sensible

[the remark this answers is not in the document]
martinRenou: I guess it should take a `KernelMessage.IInputReplyMsg['content']` as argument and return nothing.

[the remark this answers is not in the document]
madhur-tandon: Thanks!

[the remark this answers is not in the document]
martinRenou: There shouldn't be an `inputRequest` method in this class at all.

Instead, in `_processWorkerMessage` you should call the parent's `inputRequest` method that you implemented.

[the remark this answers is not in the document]
martinRenou: But it is not implemented properly, I will comment on it.

packages/kernel/src/kernel.ts
```diff
@@ -534,6 +537,25 @@ export abstract class BaseKernel implements IKernel {
...
+  /**
+   * Handle an input_request message
+   *
+   * @param msg The parent message.
+   */
+  private async _inputRequest(msg: KernelMessage.IMessage): Promise<void> {
+    const inputMsg = msg as KernelMessage.IInputRequestMsg;
+    const content = await this.inputRequest(inputMsg.content);
+    const message = KernelMessage.createMessage<KernelMessage.IInputReplyMsg>({
+      msgType: 'input_reply',
```
martinRenou, line 549: `msg_type: 'input_request'`

packages/kernel/src/kernel.ts
```diff
@@ -534,6 +537,25 @@ export abstract class BaseKernel implements IKernel {
...
   }
 
+  /**
+   * Handle an input_request message
+   *
+   * @param msg The parent message.
+   */
+  private async _inputRequest(msg: KernelMessage.IMessage): Promise<void> {
+    const inputMsg = msg as KernelMessage.IInputRequestMsg;
+    const content = await this.inputRequest(inputMsg.content);
```
martinRenou, line 547: This is not good. Subclasses should not implement an abstract `inputRequest` method.

This is a kernel-to-UI message, just like messages on the `iopub`. So this should look more like those messages.
Something like
```typescript
    const message = KernelMessage.createMessage<KernelMessage.IInputRequestMsg>({
      channel: 'stdin',
      msgType: 'input_request',
      // TODO: better handle this
      session: this._parentHeader?.session ?? '',
      parentHeader: this._parentHeader,
      content
    });
    this._sendMessage(message);
```

madhur-tandon:
@martinRenou hopefully the communication is now correct, can you have a quick overview?

packages/kernel/src/kernel.ts
```diff
@@ -256,6 +254,23 @@ export abstract class BaseKernel implements IKernel {
     this._sendMessage(message);
   }
 
+  /**
+   * Send a `input_request` message to the client.
+   *
+   * @param content The input_request content.
+   */
+  protected _inputRequest(content: KernelMessage.IInputRequestMsg['content']): void {
```
martinRenou, line 262: No `_` prefix if it's `protected`, otherwise looking good!

martinRenou (commented):
LGTM so far!

packages/kernel/src/kernel.ts
madhur-tandon: Doing!

packages/kernel/src/kernel.ts
```diff
@@ -259,13 +259,16 @@ export abstract class BaseKernel implements IKernel {
...
+    content: KernelMessage.IInputRequestMsg['content'],
+    parentHeader: any
+  ): void {
     const message = KernelMessage.createMessage<KernelMessage.IInputRequestMsg>({
       channel: 'stdin',
       msgType: 'input_request',
       // TODO: better handle this
-      session: this._parentHeader?.session ?? '',
-      parentHeader: this._parentHeader,
+      session: parentHeader?.session ?? '',
```
martinRenou, line 270: 👍🏽 It might be good to do that for all other messages (in a separate PR)

madhur-tandon: Okay.

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -288,8 +291,6 @@ export class PyoliteKernel extends BaseKernel implements IKernel {
    */
   private async _sendWorkerMessage(type: string, data: any): Promise<any> {
```
martinRenou: We should maybe rename this so that it's clearer that it must be used for requests **only**, something like `_sendRequestMessageToWorker`

packages/kernel/src/kernels.ts
```diff
@@ -75,7 +75,14 @@ export class Kernels implements IKernels {
...
           }
-          void processMsg(msg);
+
+          // TODO Find a better solution for this?
+          // input-reply is asynchrone, must not be processed like other messages
+          if (msg.header.msg_type === 'input_reply') {
+            kernel.handleMessage(msg);
+          } else {
+            void processMsg(msg);
+          }
```
martinRenou, line 85: @jtpio the mutex is preventing us from getting the `input_reply` while the `execute_request` is being processed, so we need to do this. We might want to find a better way to handle this case.

packages/kernel/src/tokens.ts
```diff
@@ -128,7 +128,7 @@ export interface IKernel extends IObservableDisposable {
    *
    * @param content - The content of the request.
```
martinRenou, line 129: The docs need to be updated input_request -> input_reply

martinRenou (changes requested):

packages/javascript-kernel/src/kernel.ts
```diff
@@ -201,9 +201,7 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
    *
```
martinRenou, line 201: Here as well input_request -> input_reply

martinRenou (changes requested):

packages/kernel/src/kernels.ts
jtpio: Looks like it should be fine like that yes.

For now the mutex is merely there so the kernel processes one message at a time. And have the queue of messages processed in order.

packages/kernel/src/kernels.ts
```diff
@@ -75,7 +75,14 @@ export class Kernels implements IKernels {
           } else {
             return;
           }
-          void processMsg(msg);
+
+          // TODO Find a better solution for this?
+          // input-reply is asynchrone, must not be processed like other messages
```
jtpio, line 80:
```suggestion
          // input-reply is asynchronous, must not be processed like other messages
```

packages/pyolite-kernel/src/kernel.ts
madhur-tandon: Done!

packages/kernel/src/tokens.ts
madhur-tandon: Done!

packages/javascript-kernel/src/kernel.ts
madhur-tandon: Done!

packages/kernel/src/kernels.ts
madhur-tandon: I am resolving this since it's okay for now. Feel free to "un-resolve" it again if needed.

madhur-tandon:
Screenshot 2021-07-18 at 1 18 57 AM

packages/pyolite-kernel/src/worker.ts
```diff
@@ -109,8 +109,8 @@ async function getpass(prompt: string) {
   const replyPromise = new Promise(resolve => {
     resolveInputReply = resolve;
   });
-  const result = await replyPromise;
-  return result;
+  const result: any = await replyPromise;
+  return result['value'];
```
martinRenou, line 113: You will need to call `toPy` to make it a Python object

packages/pyolite-kernel/src/worker.ts
```diff
@@ -119,8 +119,8 @@ async function input(prompt: string) {
...
```
martinRenou, line 123: > You will need to call `toPy` to make it a Python object

Here as well

martinRenou (changes requested):

packages/pyolite-kernel/src/worker.ts
madhur-tandon: We can directly access `value` which is a string and thus don't need a conversion.

packages/pyolite-kernel/src/worker.ts
madhur-tandon: Same as before, can directly access `value`

martinRenou (approved):
LGTM!!!

martinRenou:
We might want to find a way to make this non-async (if it's possible). But let's do that in another PR.

jtpio:
> PS -- I forgot to use `await` in front of `getpass.getpass()` and thus we see a `Promise` being returned. If I indeed use `await`, I can access the value of `b` later.

@madhur-tandon how about adding an example of using `await input()` in the default `pyolite` example notebook?

jtpio:
And thanks Martin and Madhur for working on this!

PR #256, opened 2021-07-13
Add ipycytoscape example

marimeireles:
## References

#255

## Code changes

Adds new notebook.

## User-facing changes

None

None


jtpio:
Thanks!

We also need to add `ipycytoscape` to the docs environment so the JupyterLab extension works on ReadTheDocs.

Here:

https://github.com/jupyterlite/jupyterlite/blob/65c9304cf89d311b8a48f227a0cbb2b7f50cf4bd/docs/environment.yml#L37-L50

And here for the Binder:

https://github.com/jupyterlite/jupyterlite/blob/65c9304cf89d311b8a48f227a0cbb2b7f50cf4bd/.binder/environment.yml#L50-L63

examples/pyolite - ipycytoscape.ipynb
```diff
@@ -0,0 +1,753 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "18a99a16",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import ipycytoscape as cy"
```
jtpio, line 51: We also need a first cell before to install `ipycytoscape` with `micropip`:

```python
import micropip

await micropip.install('ipycytoscape')
```

jtpio:
Looking great on the RTD preview: https://jupyterlite--256.org.readthedocs.build/en/256/_static/lab/index.html

image

.binder/environment.yml
```diff
@@ -59,6 +60,8 @@ dependencies:
   - jupyterlab-geojson >=3,<4
   - jupyterlab-kernelspy
   - jupyterlab_miami_nights
+  - networkx
+  - pandas
```
jtpio, line 64: We should be able to remove these two from the `environment.yml` files. They are only needed in the kernel.

We still need to keep `ipycytoscape` to pick up the JupyterLab prebuilt extension (using the new system in 3.0). There is some info in the docs for reference: https://jupyterlite.readthedocs.io/en/latest/configuring.html#adding-extensions

jtpio: At some point later we could also expand the contributing docs to show how to add new libraries and example notebooks to the demo site: https://jupyterlite.readthedocs.io/en/latest/contributing.html

docs/environment.yml
```diff
@@ -46,6 +47,8 @@ dependencies:
   - jupyterlab-geojson >=3,<4
   - jupyterlab-kernelspy
   - jupyterlab_miami_nights
+  - networkx
+  - pandas
```
jtpio, line 51: Removing them from here too:

```suggestion
```

jtpio (approved):
Thanks!

PR #263, opened 2021-07-15
Add RTC colors and usernames

jtpio:
## References

Follow-up to #207 

Similar to the implementation in JupyterLab: 

## Code changes

Handle username and colors in the custom RTC provider plugin.

## User-facing changes

Nicer usernames and colors, that can also be set via query string parameters.

image

None

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -16,7 +16,9 @@ import { IDocumentManager } from '@jupyterlab/docmanager';
 import {
   IDocumentProvider,
   IDocumentProviderFactory,
-  ProviderMock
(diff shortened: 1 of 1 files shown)


jtpio:
Should also work in Retro:

https://user-images.githubusercontent.com/591645/125748066-0a930793-5342-497b-b6a8-7e3f09af7e40.mp4

jtpio:
Opened https://github.com/jupyterlab/jupyterlab/issues/10625 as we seem to be inheriting this issue as well.

bollwyvl:
looks good on RTD

jtpio:
Nice, this is good to go then.

PR #269, opened 2021-07-15
Make sure display_data always contain metadata

### LINKED ISSUE #264: Problem running py3DMol - widget using publish_display_data() doesn't display
Installing and running the [`py3DMol`](https://pypi.org/project/py3Dmol/)  package for viewing 3D molecules, I note that the widget does not render:

image

```python
import micropip
await micropip.install('py3Dmol')

import py3Dmol

p = py3Dmol.view(query='pdb:1ycr')
p.setStyle({'stick': {'radius': .1}, 'sphere': {'scale': 0.25}})
```

This originally failed by explicitly testing for IPython.display tools (https://github.com/3dmol/3Dmol.js/blob/master/py3Dmol/py3Dmol/__init__.py) but while the package now runs and appears to generate some some of response, the widget is not displayed.

I note that `py3Dmol` makes use of `publish_display_data()` and wonder if that could be part of the issue?

martinRenou:
## References

Attempt to fix #264

### DIFF
--- a/packages/kernel/src/kernel.ts
+++ b/packages/kernel/src/kernel.ts
@@ -245,6 +245,9 @@ export abstract class BaseKernel implements IKernel {
    * @param content The display_data content.
    */
   protected displayData(content: KernelMessage.IDisplayDataMsg['content']): void {
+    // Make sure metadata is always set
+    content.metadata = content.metadata ?? {};
+
     const message = KernelMessage.createMessage<KernelMessage.IDisplayDataMsg>({
       channel: 'iopub',
       msgType: 'display_data',


jtpio:
Thanks!

Looks like having that default should indeed be fine anyway.

martinRenou:
(Edited the top comment with a screenshot, that seems to work)

martinRenou:
Checking the other notebooks before making this non-draft

bollwyvl:
seems legit!

martinRenou:
I honestly believe we shouldn't use it, and that py3dmol should not use `IPython.display.publish_display_data` directly. But if ipykernel is resilient with this, then maybe pyolite should be too.

jtpio (approved):
Thanks!

psychemedia:
@martinRenou Re: "I honestly believe we shouldn't do this, and that py3dmol should not use IPython.display.publish_display_data directly." Is there an approach you think they should take? I am happen to open an issue on their repo "observing" that their perhaps atypical/non-standard(?) approach caused this project to have to error trap on what appears to be an exceptional case so far.

martinRenou:
I am not 100% of this. But this `publish_display_data` seems somewhat like a private API to me. I would suggest using `display(obj, raw=True)` instead.

martinRenou:
Although this is documented API https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.publish_display_data. So maybe it's an issue on IPython's side which should set the metadata to an empty dict by default.

Issue #254, opened 2021-07-13
Add back `importhook` for patches


jtpio:
### Problem

#250 reverted the use of `importhook` as it seemed to be causing issues with the recursion limit in Chromium 91 (#248).

Ideally we should try to bring it back 

### Proposed Solution

We could try to open #239 again, and test against Chromium.

Have some end to end testing (#6) on multiple browsers (`firefox` and `chromium`) might also help catch such issues.

### Additional context

#239 #248 #250

cc @dsblank

jtpio:
There might be some changes in the next Pyodide version regarding the recursion limit: https://github.com/pyodide/pyodide/pull/1733

dsblank:
Yes, looks like the recursion limit is doubling. Let's try the PR again once we use the new version of pyodide.

bollwyvl:
There's [a bunch](https://github.com/pyodide/pyodide/pull/1756) of additional changes landing, so I'm a bit hesitant to move forward on much on the pyolite wheel side of the house at present. 

So, yeah, not worth doing much of anything until 0.18 final... though there still appears to be a number of things on [the roadmap](https://github.com/pyodide/pyodide/milestone/5), though i think some of those are already done...

bollwyvl:
#274 is in, so it's worth another look at this.

bollwyvl:
@dsblank no pressure: did you find anything else? 

It would really help our performance/bandwidth posture on e.g. jupyter.org/try if we could get `matplotlib` and `numpy`, etc. off the critical path to starting kernels.

dsblank:
@bollwyvl I haven't had a chance to revisit this. I'll see if I can catch up with recent development in jupyetrlite and pyodide and report back.

bollwyvl:
Semi-related: at present, we only ship one version of the `widgetsnbextension` shim package, and [install it](https://github.com/jupyterlite/jupyterlite/blob/v0.1.0b3/packages/pyolite-kernel/src/worker.ts#L53) at startup.

However, on #563, it appears this version is incompatible with the `ipywdigets 7.7,0` that just came out:

```
ValueError: Requested 'widgetsnbextension~=3.6.0', but widgetsnbextension==3.5.0 is already installed
```

will look into it, as it _might_ be the case that we can just ship multiple versions...

PR #274, opened 2021-07-17
Upgrade to Pyodide 0.18.0

bollwyvl:
## References

- [x] follow-on / contains #178
  - got a bit confused with the history, and didn't want to push against the main repo branch :blush:
- [ ] blocked by #254
  - really need to get some of the heavy hitters like `matplotlib` out of the critical load path for interactive testing!

## Code changes

- [x] explicit call to install `micropip`
- [x] make `copy`s of long-running python objects like `kernel`, `interpreter` and `*_stream`s to avoid GC/ref errors 
- [x] update outer environment(s) to python 3.9 (as pyodide is now _also_ 3.9)
  - [ ] add a `matrix` for `python-version` in CI to ensure we maintain compat back to advertised `3.7` (this could be a separate PR)
- [x] **remove** debug messages for determining current pyodide `dev` API support for `fileSystem`/`FS`
- [ ] add a `py - filesystem.ipynb` with some interactive examples of this API
  - [x] show using `pyodide_js.FS` with stdlib python, e.g. `pathlib` 
  - [x] show using `pyodide_js.FS.filesystems.IDBFS` directly 
  - [ ] contribute back to `pyodide` upstream

## User-facing changes

- n/a  (yet, though we'd like to see some performance benefits)
 
- going to 3.9 shouldn't be too big of an issue... but probably non-negotiable!

## Notes

- some work-in-progress notebook stuff, showing how user python code will be able to interact with `FS`:
  - https://gist.github.com/bollwyvl/616802a1765dc702ee97a0f065adad8c


jtpio:
Nice, thanks for starting this!

>     * really need to get some of the heavy hitters like `matplotlib` out of the critical load path for interactive testing!

Maybe it will be possible to bring `importhook` back with the latest Pyodide if it changed something regarding the recursion limit.

bollwyvl:
An update: it looks like the nightlies with `FS` are up... but `Pygments.js` is not loading correctly, so still blocked for a tick...

bollwyvl:
Ok, I've moved that notebook from the gist into this PR for the time being, so that it shows up under the docs build. The example now includes persisting to `IndexedDB` from user python code, which is a little crazy. Certainly a lot left to do, but getting to a live interactive state is what this is all about, right? Here's hoping i didn't mess up the docs build or something :crossed_fingers: .

packages/pyolite-kernel-extension/src/index.ts
```diff
@@ -12,7 +12,7 @@ import { PyoliteKernel } from '@jupyterlite/pyolite-kernel';
 /**
  * The default CDN fallback for Pyodide
  */
-const PYODIDE_CDN_URL = 'https://cdn.jsdelivr.net/pyodide/v0.17.0/full/pyodide.js';
+const PYODIDE_CDN_URL = 'https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js';
```
jtpio, line 15: If there was a way to pin on a commit / hash, maybe we could already start shipping it.

bollwyvl: ha, would rather not, just yet... i think waiting for the npm releases will be worth it...

bollwyvl:
SNAAAP:

https://github.com/pyodide/pyodide/releases/tag/0.18.0

bollwyvl:
It appears we semi-frequently get into trouble with our reference to `kernel` in the worker

```
Uncaught (in promise) Error: This borrowed attribute proxy was automatically destroyed in the process of destroying the proxy it was borrowed from. Try using the 'copy' method.
    at _getPtr (pyodide.js:1)
    at pyodide.js:1
    at Object.set (pyodide.js:1)
    at self.onmessage (75af6e0a-7e2c-4e54-ad63-8249388b1216:315)
```

bollwyvl:
Was able to locally run through all of the examples except `folium` (which needs a lot of hacks _anyway_, apparently).

bollwyvl:
On RTD, it looks like all of the cruft from `folium` can be boiled down to:

```
import micropip
await micropip.install("folium==0.12.1")
```

MarkupSafe is now available directly from `pyodide` as a pre-built wasm package.

bollwyvl:
The only hard regression appears to be a `maximum call stack` error in `ipyleaflet`: using the download technique employed by [micropip](https://github.com/pyodide/pyodide/blob/0.18.0/packages/micropip/src/micropip/micropip.py#L60), was able to get around it...

bollwyvl:
Well I guess we need to merge this in order to start looking into some of the other open PRs and issues. Hold on to your hats!

PR #276, opened 2021-07-19
Basic support for renaming the RTC session

### LINKED ISSUE #267: RTC: implement `setPath`
### Problem

`setPath` is new in JupyterLab `3.1.0rc1` and was added in https://github.com/jupyterlab/jupyterlab/pull/10470

We should also implement it here:

https://github.com/jupyterlite/jupyterlite/blob/424c4ad2ca1b25cd608ae4cf152d7aa44cf7ec59/packages/application-extension/src/index.tsx#L52-L54

### Proposed Solution

Not sure yet what the preferred way would be, but maybe @dmonad would have some pointers for how to deal with this without a server and with the WebRTC provider.

jtpio:
## References

Fixes #267 

## Code changes

Implement simple logic to rename a file, and continue the RTC session seamlessly.

## User-facing changes

Users should be able to rename a file. However since the files are distributed (everyone has a copy in indexeddb), users must explicitly save the notebook to get a copy.

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -61,15 +61,6 @@ class WebRtcProvider extends WebrtcProvider implements IDocumentProvider {
     }
     this._room = options.room;
     this._path = options.path;
-
-    this.on('rename', (args: any) => {
-      const path = args.path;
-      if (path !== this._path) {
-        this._path = path;
-        console.log('new path', path);
-        this._reconnect(path);
-      }
-    });
   }
 
   private _reconnect(path: string): void {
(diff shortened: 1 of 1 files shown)


jtpio:
Not really working for now.

> users must explicitly save the notebook to get a copy.

Also we'll probably want to rename the existing files for the other peers, instead of creating new ones with new names.

jtpio:
Posting this PR here as reference as it might be relevant when we resume work here: https://github.com/jupyterlab/jupyterlab/pull/11197

jtpio:
Closing as this would likely have to be implemented in https://github.com/jupyterlite/jupyterlab-webrtc-docprovider now (or take a different approach).

Issue #80, opened 2021-04-28
Add architecture diagram to README or docs somewhere


telamonian:
There were some whispers at today's jlab dev meeting about a architecture/explainer diagram that had been made for an earlier presentation of `jupyterlite`. I for one would love to see it.

@jtpio Any chance that we can get that diagram posted publicly somewhere? I think for now it would go nicely in the README, and can then be migrated to the deeper docs once those exist

jtpio:
Thanks @telamonian.

Yes for now there is one in this temp gist: https://gist.github.com/jtpio/3763af75ec8cd7bfdce86501563fef09

But it could be improved and other diagrams covering different parts could also be added (https://github.com/jtpio/jupyterlite/issues/34).

bollwyvl:
So... as a systems enginerd, i love building diagrams (obviously) but having them _purely_ for documentation means they get out of date almost immediately. Over on [jupyterlab-lsp](https://jupyterlab-lsp.readthedocs.io/en/latest/Architecture.html), we spent a bunch of effort getting them set up, but then I have shamefully been really bad at updating them while @krassowski goes off and builds :fire: stuff. And graphviz is _barely_ fit for purpose in this case. `ipydrawio` (which i want _REAAAALY BADLY_ in jupyterlite) is better at both being a diagram _and_ being a model. The reason I haven't PRd the initial (or any other) diagram from #34 yet is I need to think real hard about how to keep them accurate and get real value out of that model.

Another challenge we have is this architecture is best described as a delta from a reference architecture (`pip` -> `@jupyterlab/application-top` -> `jupyterlab` -> `jupyter_server`), for which there are no systems models.

Anyhow, onto testability: the initial chartjunk of `ls packages/*/package.json apps/*/package.json` is relatively easy, as we can check the symmetric difference of the labels without even understanding the underlying XML very much. So I can probably get something out pretty quick that would keep this up-to-date, and add ipydrawio to the docs/binder environment, and deal with dumb-old-regexen or stdlib xml parsing and xpath to be more precise.

An area of (sometimes-funded) research is turning visual representations into executable test cases #6. For example, #79, that's a relatively complex sequence diagram involving multiple re-entrance to a browser session... and it might not make that great of an explainer, aside from showing that multiple apps can reuse the contents manager.

PR #278, opened 2021-07-20
Add the architecture diagram

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/80

Using the diagram that has been used so far. So we have it somewhere now and can more easily be edited.

It is of course incomplete, but should still be accurate for now. Ideally we could produce something better later that would cover other aspects of JupyterLite (contents management, configuration) while making sure they don't get out of date too quickly.

## Code changes

Docs

## User-facing changes

In the developer docs:

image

None


bollwyvl:
Perhaps as follow-on, but... 
- put the image in `examples` so it's something people can play with... 
  - assuming `jupyter-drawio` can handle SVG... 
  - otherwise the XML there, and `.svg` in `docs`
    - and a check that the svg is always newer than the xml, gah
- can we put under any sort of test? 
  - from the codebase, we pull out number/names of:
    - apps
    - kernels
    - storage drivers
    - CDN providers
  - with e.g. `lxml` (or even standard lib) we pull out shapes/strings that should match the above
  - test fail if they do not agree
    - or even have them as notebooks that would show what's out-of-date (but not strictly _fail_)

i did a bit about this on [ipydrawio](https://ipydrawio--52.org.readthedocs.build/en/53/tutorials/designing-jupyter-extensions/index.html#testing)... that's specifically with the `.dio.svg` format, but would probably work with the straight-up `mxModel` dialect.

jtpio:
> i did a bit about this on [ipydrawio](https://ipydrawio--52.org.readthedocs.build/en/53/tutorials/designing-jupyter-extensions/index.html#testing)

Maybe we can reuse some of this here too?

It would indeed be useful to have it in the examples and elaborate the pipeline. For now the idea was just to dump that diagram somewhere so it's more visible, but we can definitely improve that in follow-up PRs.

bollwyvl:
Welp if you get the source xml into examples/, and a check that the svg is
up-to-date, we have enough to move forward with...

bollwyvl:
The example is coming up empty on RTD...

jtpio:
Ah that's weird.

I uploaded the svg on diagrams.net, and then download the `.drawio` from there.

bollwyvl:
>  then download 

make sure the "compress and obfuscate xml" isn't on :rofl:

bollwyvl:
Although, if you're dropping the already-generated SVG into a drawio doc, it will end up as a base64-encoded data URI, and its value for maintainability and testing is... diminished.

jtpio:
The "source" is the SVG + metadata.

jtpio:
Dropping https://github.com/jupyterlite/jupyterlite/pull/278/commits/a8728ed0167a1893a7e8510482c1e7b3ec7d8001 for now.

We can track it and the other items mentioned in https://github.com/jupyterlite/jupyterlite/pull/278#issuecomment-883448265 in separate issue(s).

jtpio:
Merging so we don't block on this.

Happy to tackle the rest as part of #279, or @bollwyvl if you want to have a closer look feel free!

PR #281, opened 2021-07-21
Pin jupyter widgets versions

martinRenou:
Currently, the bqplot example on the JupyterLite readthedocs deployment is failing. This is because the federated extension for `0.12.29` is installed, while the latest version of bqplot (installed by micropip) is `0.12.30`, which is incompatible.

This can be fixed by pinning versions


jtpio:
> This is because the federated extension for `0.12.29` is installed, while the latest version of bqplot (installed by micropip) is `0.12.30`, which is incompatible.

Shouldn't they be compatible because of semver? Or does it mean we need to have a strict version matching between the federated extension installed on the deployment, and the version of the Python package installed with `micropip`?

bollwyvl:
> This can be fixed by pinning versions

For today!

>  strict version matching between the federated extension installed on the deployment, and the version of the Python package installed with micropip?

I think once we extend the meaning of pre-indexed wheels during the build step, and make %pip install those instead of going out to pypi, (if requested) this problem will mostly go away, for the case where it's mostly known in advance what packages will be needed at run time and have federated components. 

As _nearly_ every lite site will have `federated_extensions`, i'd think the python side (which every lite site _shouldn't_ need to have) would inherit from that... but only for wheels, of course.

martinRenou:
> Shouldn't they be compatible because of semver? 

Due to the way bqplot is implemented (which is definitely debatable, though most widgets seems to do it that way), the Python module requires a certain front-end extension version. Here bqplot 0.12.30 (latest) requires the bqplot federated extension version 0.5.31. Though we installed 0.5.30 on the server. So pulling the latest bqplot version with micropip results in a failing setup.

> Or does it mean we need to have a strict version matching between the federated extension installed on the deployment, and the version of the Python package installed with micropip?

Ideally we wouldn't have to. But this is how it works currently for bqplot and most widgets. Ideally those widgets packages would have a "widget model version" defined, and if the Python module and the front-end extension are working against the same model version they would be compatible.

But this would mean having a Python module version, a model version, and a front-end version which is less convenient to track.

> For today!

Yep...

jtpio:
> Here bqplot 0.12.30 (latest) requires the bqplot federated extension version 0.5.31. Though we installed 0.5.30 on the server. So pulling the latest bqplot version with micropip results in a failing setup.
> 
> >

Ah right, I though ipywidgets was checking the semver range for compatibility.

And it's true that up until now in the majority of cases with traditional jupyter environments, the Python and frontend versions would usually always be in sync (when installed in the same environment).

jtpio:
So yeah pinning the versions like in this PR sounds like a good thing to do for now :+1:

bollwyvl:
In the near term, can we add _something_ that will fail if the `environment.yml` and notebook pins get out of sync? I don't even care if they can be auto updated (you know, versions change things!) but at least we'd know before shipping/manually checking things.

martinRenou:
> In the near term, can we add something that will fail if the environment.yml and notebook pins get out of sync?

What do you mean? It's actually failing if they get out-of-sync and not display the widget

bollwyvl:
Right, checking the actual notebook in a real browser is the goal...
Someday.

For before, if there was a "task_check" item that:

- Loaded the federated extension chunk from environment.yml
- looked at the notebooks for `micropip. install` that matches the part
before ==
- fail if any are different

Then the docs job would fail (it's running check for broken links) and we
wouldn't ship something broken.

jtpio (approved):
Thanks!

Issue #286, opened 2021-07-22
(Disable) Incompatible retro extensions?


bollwyvl:
### Problem

- `jupyterlab-tour` uses StateDB, which isn't available in (any?) of the retro pages
- also `jupyterlab-drawio` doesn't work, apparently

<details>
<summary>
These throw a bunch of scarrors:
</summary>
<pre>
Plugin '@jupyterlab/drawio-extension:plugin' failed to activate.
(anonymous) @ index.es6.js:282
Promise.catch (async)
(anonymous) @ index.es6.js:281
e.start @ index.es6.js:280
(anonymous) @ index.js:250
async function (async)
(anonymous) @ index.js:42
60880 @ index.js:256
v @ bootstrap:19
Promise.then (async)
(anonymous) @ bootstrap.js:90
async function (async)
(anonymous) @ bootstrap.js:75
37559 @ bootstrap.js:64
v @ bootstrap:19
(anonymous) @ startup:5
(anonymous) @ bundle.js:1
index.es6.js:283 Error: No provider for: @jupyterlab/filebrowser:IFileBrowserFactory.
    at f.e.resolveRequiredService (index.es6.js:197)
    at index.es6.js:155
    at Array.map (<anonymous>)
    at f.e.activatePlugin (index.es6.js:155)
    at index.es6.js:281
    at Array.map (<anonymous>)
    at f.e.start (index.es6.js:280)
    at index.js:250
(anonymous) @ index.es6.js:283
Promise.catch (async)
(anonymous) @ index.es6.js:281
e.start @ index.es6.js:280
(anonymous) @ index.js:250
async function (async)
(anonymous) @ index.js:42
60880 @ index.js:256
v @ bootstrap:19
Promise.then (async)
(anonymous) @ bootstrap.js:90
async function (async)
(anonymous) @ bootstrap.js:75
37559 @ bootstrap.js:64
v @ bootstrap:19
(anonymous) @ startup:5
(anonymous) @ bundle.js:1
index.es6.js:282 Plugin 'jupyterlab-tour:plugin' failed to activate.
(anonymous) @ index.es6.js:282
Promise.catch (async)
(anonymous) @ index.es6.js:281
e.start @ index.es6.js:280
(anonymous) @ index.js:250
async function (async)
(anonymous) @ index.js:42
60880 @ index.js:256
v @ bootstrap:19
Promise.then (async)
(anonymous) @ bootstrap.js:90
async function (async)
[...]

PR #287, opened 2021-07-22
Provide `IStateDB` in retro

jtpio:
## References

To address the `jupyterlab-tour` part of https://github.com/jupyterlite/jupyterlite/issues/286

## Code changes

Add `@jupyterlab/apputils-extension:settings` to the retro app.

## User-facing changes

Users should be able to take tours with RetroLab if the `jupyterlab-tour` extension is installed and tour exists.

None

### DIFF
--- a/app/retro/index.template.js
+++ b/app/retro/index.template.js
@@ -75,6 +75,7 @@ async function main() {
       [
         '@jupyterlab/apputils-extension:palette',
         '@jupyterlab/apputils-extension:settings',
+        '@jupyterlab/apputils-extension:state',
         '@jupyterlab/apputils-extension:themes',
         '@jupyterlab/apputils-extension:themes-palette-menu'
       ].includes(id)


jtpio:
So this seems to fix it, although the existing tours are not usable:

image

bollwyvl:
Notebook tour works like a frigging champ:

Screenshot from 2021-07-22 14-52-25

jtpio:
> Notebook tour works like a frigging champ:

Ah right there is one for the notebook, great!

bollwyvl:
Unfortunately, the notebook and main tours are bundled together as a single plugin :crying_cat_face:  

Our boring custom tour uses `jp-JupyterLogo` to fire it up, so that could probably be fixed, e.g. `#jp-JupyterLogo, #jp-RetroLogo`... should only be one on the page at a time

jtpio:
> Unfortunately, the notebook and main tours are bundled together as a single plugin crying_cat_face

Sounds like we could split them into multiple plugins upstream? Which should make it easier to disable / enable here.

jtpio:
So this should then be good to go!

We can iterate on the tours separately.

jtpio:
> Sounds like we could split them into multiple plugins upstream?

Opened https://github.com/jupyterlab-contrib/jupyterlab-tour/issues/40 to discuss.

Issue #285, opened 2021-07-22
Pyolite: handle syntax errors


jtpio:
## Description

Syntax errors seem to be failing silently, but should ideally be bubbled up to the user.

![image](https://user-images.githubusercontent.com/591645/126653864-7ab65640-6f3a-4053-990a-1c1dc13629f9.png)

## Reproduce

1. Open https://jupyterlite.github.io/demo/lab/
2. Create a new notebook
3. Enter and execute the following:

```python
def foo():
    
```

## Expected behavior

The error should appear in the output cell.

## Context

- Operating System and version: Ubuntu 20.04
- Browser and version: Firefox 89
- JupyterLite version: 0.1.0a5

madhur-tandon:
If I try `import whatever` in the cell, the error occurs in the output cell itself.
While `def foo():` fails.

Are these 2 related? or they follow a different code path to propagate the error?
Ideally if `import whatever` works, `def foo():` should work too I guess?

jtpio:
It could be because `def foo():` is considered as incomplete input?

madhur-tandon:
Sending a PR in a few minutes.

martinRenou:
IIRC the syntax error is a special case in ipykernel

martinRenou:
IIRC the syntax error is a special case in ipykernel

jtpio:
Thanks @martinRenou for the pointer.

I guess we can try to check what `ipykernel` does and mimic the logic here.

jtpio:
Reopened and updated the title so we can track this.

martinRenou:
Actually my memory is failing me, there is no special case for syntax errors in ipykernel (at least I don't find it).

The actual issue was that pyodide's `_load_packages_from_imports` was failing against syntax errors, this failure was not catch by the interpreter (hence not forwarded to the Jupyter client properly).

I opened another PR https://github.com/jupyterlite/jupyterlite/pull/304 that wraps the `_load_packages_from_imports` call inside a try/except, catching the exception if it occurs with `self.interpreter.showtraceback()`.

PR #290, opened 2021-07-23
fix syntax error showing in JS console

madhur-tandon:
Screenshot 2021-07-23 at 1 30 27 PM

Fixes https://github.com/jupyterlite/jupyterlite/issues/285

### DIFF
--- a/packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
+++ b/packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
@@ -71,7 +71,8 @@ class Pyolite:
     async def run(self, code):
         self.interpreter._last_traceback = None
         exec_code = self.interpreter.transform_cell(code)
-        await _load_packages_from_imports(exec_code)
+        if self.is_complete(exec_code)["status"] != "incomplete":
+            await _load_packages_from_imports(exec_code)
         if self.interpreter.should_run_async(code):
             self.result = await self.interpreter.run_cell_async(
                 code, store_history=True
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks!

Looks like there might be other cases, for example `x = `:

image

As a comparison in stock JupyterLab / ipykernel:

image

madhur-tandon:
@jtpio I have added the case when the status could be `invalid`. There is one more case where this can be `unknown`. I haven't added that yet. But do you have an example where the `status` could indeed be `unknown`? Anyway, should I also add `unknown` to the list?

madhur-tandon:
Attaching a screenshot of the fix.

Screenshot 2021-07-23 at 2 26 22 PM

jtpio:
Thanks @madhur-tandon, this already looks like a nice improvement :+1: 

> There is one more case where this can be unknown.

Need to check, but this could also be tracked and fixed in a separate PR.

madhur-tandon:
The 4 message types are given here: https://jupyter-client.readthedocs.io/en/stable/messaging.html#code-completeness
We only need to see if we want to do this for `unknown` type.

jtpio (approved):
Looks good, thanks!

packages/pyolite-kernel/py/pyolite/pyolite/kernel.py
```diff
@@ -71,7 +71,8 @@ def complete(self, code, cursor_pos):
     async def run(self, code):
         self.interpreter._last_traceback = None
         exec_code = self.interpreter.transform_cell(code)
-        await _load_packages_from_imports(exec_code)
+        if self.is_complete(exec_code)["status"] not in ["incomplete", "invalid"]:
```
martinRenou, line 74: I believe we shouldn't do an is_complete check in the run request. It might be handled differently in ipykernel.

PR #304, opened 2021-08-10
Prevent calling "is_complete" from execution request

martinRenou:
Wrapping the `load_packages_from_imports` inside a try/except instead.


bollwyvl:
Does this fix any existing issues, or is it for something new entirely?

martinRenou:
It's a follow up of this PR #290

PR #306, opened 2021-08-16
access header key after formatResult on whole object

madhur-tandon:


martinRenou:
I guess we would need to actually use this value. It was not failing before simply because we don't make use of this yet.

madhur-tandon:
Yes, gotta use it next in `kernel.ts`

PR #307, opened 2021-08-16
use parent header passed from worker

madhur-tandon:


packages/kernel/src/kernel.ts
```diff
@@ -228,13 +228,18 @@ export abstract class BaseKernel implements IKernel {
    * @param parentHeader The parent header.
    * @param content The stream content.
    */
-  protected stream(content: KernelMessage.IStreamMsg['content']): void {
+  protected stream(
+    content: KernelMessage.IStreamMsg['content'],
+    parentHeader: any = undefined
```
martinRenou, line 233:
```suggestion
    parentHeader: IHeader | {} | undefined = undefined
```

madhur-tandon: Done, Thanks!

martinRenou:
Works like a charm 💯

bollwyvl:
I just tried out this, works well:

```python
from ipywidgets import interact
from IPython.display import Markdown, display

@interact(cookies=slider)
def cookies(cookies=slider.value, calories=(0,150)):
    total_calories = calories * cookies
    if cookies:
        display(Markdown(f"If each cookie contains _{calories} calories_, _{cookies} cookies_ contain **{total_calories} calories**!"))
    else:
        display(Markdown(f"No cookies!"))
    if total_calories > 2000:
        display(Markdown(f"> Maybe that's too many cookies..."))
```

Given this can be a tricky one, but also a super useful example, may let's add that to the end of `pyolite - interactive-widgets.ipynb`?

madhur-tandon:
@bollwyvl added the example in the latest commit. Thanks!

martinRenou:
Great thanks!

Issue #225, opened 2021-07-07
Avoid too many requests to PyPI


jtpio:
The biggest win, in my mind, is we could then lobby for an efficient, in-browser `https://jupyter.org/try`. However, we couldn't release and then promote a DDoS of  `pypi.org`. This would need, as table stakes, 
- #137 (to cut down on repeat bandwidth needs)
- ideally #45 (and have a self-hosted subset of whls needed to enable any shipped demo content)

_Originally posted by @bollwyvl in https://github.com/jtpio/jupyterlite/issues/218#issuecomment-874895720_

jtpio:
@bollwyvl let's track this in a separate issue, based on your comment in https://github.com/jtpio/jupyterlite/issues/218#issuecomment-874895720.

It indeed sounds like something that could already be improved with some tooling around managing wheels and self-host them.

Also we might want to check / profile whether the wheels are or can be cached when downloaded via `micropip`.

martinRenou:
As a good start, we could maybe self-host `traitlets` and `IPython` which are used by `pyolite`

bollwyvl:
Thanks for opening this.

> tooling around managing wheels and self-host them.

Right. We already know we need some wheel handling for:
- better labextension support (so `jupyter lite build` doesn't have to install matplotlib in docs)
  - just need to copy 
- runtime packages for python-based kernels

It's also highly likely these sets will overlap... but i think trying to load labextensions from the whls directly in the browser is a fool's errand.

I have do some thinking/tinkering with the `jupyter-lite.schema.json` and find some patterns we like, as at the end of the day, the browser needs a heads-up about when all this happening... even if it ends up being a directory of wheels, a look at pyodide's [micropip](https://github.com/pyodide/pyodide/blob/1485732211816a545887b154fc41ce3781785ff7/packages/micropip/micropip/micropip.py) indicates we might need to patch that, re-implementing a subset of the PyPI warehouse API in an `all.json` or something.

bollwyvl:
Unlike #137, the merged #655 doesn't cache wheels offline, but should.

bollwyvl:
> if nobody does in the meantime

I think if you've got runway, you should go for it... I'm really strapped for open source contribution bandwidth at the moment.

In light of #679, i think the biggest near term thing is a graceful "off switch" for all serviceworker related stuff: we really don't want the whole system to come crashing down entirely because a user is stuck using e.g. a fixed browser in a library.

bollwyvl:
[...]

Issue #186, opened 2021-06-29
Pyolite configuration


martinRenou:
### Problem

It is currently not possible to automatically install Python packages inside the pyolite kernel.

### Proposed Solution

It would be nice to be able to configure Pyolite to include some packages that would automatically be installed with `micropip`.

cc. @psychemedia who suggested something here https://github.com/jtpio/jupyterlite-demo/pull/7#issuecomment-870144941

### Additional context

It might be possible to have it once we use IPython for the execution (see related work https://github.com/jtpio/jupyterlite/pull/171), as it's possible to configure IPython itself to run code at the shell start.

bollwyvl:
I feel we need to figure out some way to get these hosted with the build,
get the service worker going, use dat/ipfs, etc. before installing more
stuff magically. Spamming pypi with requests without any kind of caching
isn't very nice... They probably weren't signing up to be a static cdn for
web page assets.

jtpio:
Maybe one way could be to make that part that @psychemedia linked to in https://github.com/jtpio/jupyterlite-demo/pull/7#issuecomment-870144941 configurable at build time:

https://github.com/jtpio/jupyterlite/blob/5225266445d65c05cc2f545fceb1fa9f44380709/packages/pyolite-kernel/src/worker.ts#L20-L25

To download the wheels only once and place them at the right location.

But that would be specific to the pyolite kernel, and the jupyterlite toolchain should ideally be kernel agnostic.

bollwyvl:
I haven't dug enough into `micropip`, but presumably it can accept additional (preferred) index pages... given a list of wheels, that should be easy enough to add. What we _don't_ want to do is end up having the maintenance burden of little PRs for every little patch/build someone needs.

At any rate, wheels _will_ get special treatment in the toolchain _anyway_ as they are the most reliable source of labextensions, as we can predict precisely where the assets will be inside them... until we have a `wasm-32` target from conda-forge, which would probably be the most desirable way for _all_ non-basically-pure JS kernels. Further, there are a number of wrapper kernels that reuse `ipykernel`, etc. which would be able to use wheel-based mechanisms.

psychemedia:
One of the things I hacked to get the folium demo working was a prebuilt universal wheel for one of the packages served from a Github page: https://github.com/OpenComputingLab/vce-wheelhouse

It's probably a good idea to separate out the ideas of installing from pypi, from a JupyterLite wheelhouse, or from a local wheel.

[...]

PR #310, opened 2021-08-29
add piplite for customizing pyolite packages, automate wheel management

bollwyvl:
## References

- starts on #151

## Code changes

- [x] removes all the hard-coded `whl` URLs from the `pyolite` extension
- [x] hoist pyodide-specific config to new namespaced `litePluginSettings`
  - [x] adds `pipliteUrls` to `jupyter-lite.json`
  - [x] create a schema for a new well-known `all.json` describing the necessary subset of the warehouse API
  - [x] optionally `disablePyPIFallback` access altogether
- [x] wraps `micropip` _really_ early to obey these 
  - [x] encapsulated as `piplite`,  could be upstreamed, but we might need other stuff here (e.g. `%pip`)
- [x] implement `jupyter_lite_config.json` and CLI-serviceable options for
  - [x] specify custom `micropipUrls`
  - [x] validating the metadata schema
- [x] adds CLI, `jupyter lite pip index` for _just_ generating an `all.json` for a directory of wheels
- [x] anticipate wheels, etc. in `federated_extensions` (including this thing itself in #386)
  - [x] move wheels into `pypi` during webpack build of app, handle more robustly
  - [x] define `package.json` conventions/schema for declaring location of wheels, e.g. `#/piplite/wheelDirs`
  - [x] determine when to find wheels in `package.json`, weighing flexibility/resilience and performance (e.g. _2n+1 requests_)
    - [x] at build time, updating `jupyter-lite.json`
    - [ ] at run time... could be done in parallel with initial pyodide load? 

## User-facing changes

- for GUI users
  - might be _slightly_ faster, as it would save ~5 requests to `pypi.org`... but not really until more stuff is indexed
  - if fully cached, dependencies that change "in the wild" will not impact users of an already-built site, leading to less breakage a la #311
- for CLI users
  - eventually should be able to
    - point to alternate PyPI indexes, mirrors
    - easily include additional, pre-cached packages
      - e.g. offer `pip freeze` (or equivalent) from a known set of "leaf" packages (plus the kernel dependencies)
      - make no requests to pypi.org _at all_

- n/a, probably, as this was either working and you didn't notice it, or was broken and you couldn't tell


bollwyvl:
Looks alright on RTD. I guess if i get the CLI/build thing squared away, we could go ahead and remove all of the labextenion-based version pins in the docs build, and just ship those whls as well...  certainly open to other feedback, however!

jtpio:
Thanks for starting this!

bollwyvl:
One effect this has: the `pyolite-kernel` package no longer ships the `whl` files, so if it _did_ get published to npmjs, they wouldn't have enough data. We might need to rethink
Yeah, between labextensions, pyodide, and micropip, there's a lot to juggle, but I think we'll benefit from having some sane approaches _without_ having to implement (or even strictly _require_) a package manager.

bollwyvl:
Of note: if we do go to where the `docs` app provides a full archive of all the _known_ wheel dependencies (starting with the `ipykernel` stack), this would insulate a site from something like #311 where dependencies shift after deployment.

jtpio:
This should also fix https://github.com/jupyterlite/jupyterlite/issues/225

bollwyvl:
>  fix #225

Right, but it kinda depends... as currently implemented, it's only serving our custom patched `whl`s for a few files. We _could_ merge as is, just so that's cleaned up, but still have to do some more dancing with it to e.g. merge the "built-in" wheels with user wheels. 

So when a site builder does a build, they'll get _at least_ these couple wheels. They will then have the _option_ of adding additional wheels, so that would hit pypi each _build_, but not _each page hit_ by a site visitor. Practically, yeah, visitors to our fairly large docs site would pull make _many_ fewer requests to pypi, and would even make fewer _overall_ requests, because the whole warehouse API will come down the wire in one JSON request (as opposed to one JSON request per package).

These requests will still hit the server hosting lite, until we determine the appropriate approach for offline support: #137 is the cadillac, but... it only takes premium, and might not work for everyone. Over on pyodide, there has been some progress on demonstrating some [novel emscripten file systems](https://github.com/pyodide/pyodide/issues/1715#issuecomment-900122795) that we might be able to back with contents (which we _also_ need to modularize... will make an issue).

bollwyvl:
My apologies for not making further progress on this: still dealing with moving, etc. and haven't yet found a way to get to work on this during `$DAY_JOB`. If someone else wants to pick this up in the meantime, I can provide low-velocity support via PR comments!

jtpio:
No problem Nick, thanks for the heads-up!

jtpio:
Nice this is looking good!

> this is 49mb... which really doesn't seem _too_ bad for theoretically being free of version drift concerns.

I also think it's ok, especially compared to the size of lab and retro combined.

dsblank:
I just noticed this work, and it warms my heart :) 

I was thinking about a similar interface for what we could use in my day-job: have a list of wheel repositories so that a request to install a wheel could look first in our local set of wheels, and if not found, go to the next repository, bottoming out at pypi (if it were in the list).

I was thinking that this code really belongs in micropip, rather than patching it here.

bollwyvl:
I'm semi-bullish on this coming up clean, finally, and have started some basic docs... so will take draft off so we can discuss moving forward.

jtpio:
Thanks @bollwyvl!

This looks good and works well on RTD.

Just wondering how the new `jupyter-lite.json#/micropipUrls` will work if we start moving the Pyolite kernel to a separate repo (https://github.com/jupyterlite/jupyterlite/issues/386). Should it be part of the schema even if it's specific to the Python kernel?

bollwyvl:
> Should it be part of the schema even if it's specific to the Python kernel?

A good point! I let this sit so long that the architecture has moved on, and should generally re-work this so extracting it can be all-but automated, and show a clear path for future "heavyweight" extensions. I'll add it over there, just to keep the story consistent...

app/jupyterlite.schema.v0.json
```diff
@@ -120,6 +114,34 @@
               }
             ]
           }
+        },
+        "litePluginSettings": {
```
bollwyvl, line 118: Thoughts on naming?

bollwyvl:
I've started separating out the config values (see earlier schema comment) and it seems mostly well-behaved. Still need to consider how we might provide and extensible validation mechanism.

It's going to take some doing to pick apart how multiple extensions will provide `micropipUrls`: basically the pre-cooked `app/lab/build/wheels` would go away, but there isn't anywhere else clever to put it yet. 

My current thinking is the `@jupyterlate/pyolite-extension` prebuilt assets, packaged inside the `jupyterlite-pyolite` build-time package will end up with the wheels  (e.g. the `pyolite` browser run-time wheel)  and pre-cooked `all.json`. Then, its build `entry_point` would be able to look inside each `federated_extension`'s `package.json` to find `#/pyolite/wheelDir` and add those dynamically to `micropipUrls`. I don't think the versions of the packaged wheels will end up being canonically useful, as we currently do, but it sure is nice to have a real string to parse someplace.

bollwyvl:
Back up to green, gotta put it down for a while, though...

bollwyvl:
The test fail looks spurious... the issue is very probably that the translation `entry_point` lookup dominates the actual activity. Likely we need to pre-cache the output of `get_language_packs` so we can calculate the expected paths, and change `one_translation_path` to _actually_ handle... one translation path... out of scope here, and probably needs an issue...

bollwyvl:
Well, things are feeling a _little_ better than before this weekend, but still not ready to ship. As I've been digging around in here, I think `piplite` needs to become a real wheel, and have started moving in that direction: some properties it would have:
- the _one_ hard-coded URL, installed from js with `micropip.install`
- would become the backing implementation for the `%pip` magic
  - all our example would change to look like `%pip install ...`
  - `micropip` would then remain, unchanged, and only use `pypi.org`
- any references to `micropip` introduced in this PR up to this point (e.g. `micropipUrls`, `addons/micropip.py`) would become `piplite`
- would (eventually) take over much of the annoying work of #235/#254
  - (maybe by vendoring to avoid more chicken/eggs)
  - _might_ become configurable to support #186

jtpio:
Thanks @bollwyvl for the patience and continuing pushing on this :+1:

app/jupyterlite.schema.v0.json
jtpio: I think `litePluginSettings` sounds good

bollwyvl:
Yeah got stuck going back to the hard-coded wheel for piplite. The code is
much cleaner, which is nice, but i don't yet know if it will work!

examples/jupyter_lite_config.json
```diff
@@ -3,7 +3,7 @@
     "contents": ["."],
     "federated_extensions": [
       "https://github.com/conda-forge/releases/releases/download/noarch/bqplot-0.12.30-pyhd8ed1ab_0.tar.bz2/bqplot-0.12.30-pyhd8ed1ab_0.tar.bz2",
-      "https://github.com/conda-forge/releases/releases/download/noarch/ipycanvas-0.9.1-pyhd8ed1ab_0.tar.bz2/ipycanvas-0.9.1-pyhd8ed1ab_0.tar.bz2",
```
martinRenou: Was there a reason for downgrading ipycanvas?

IIRC we need ipycanvas 0.9.1 specifically for the jupyterlite lua kernel. We can always pin ipycanvas correctly in the example deployments of lua though.

bollwyvl: no particular reason, probably a bad merge or something... this PR's getting a little long in the tooth.

bollwyvl:
Buh, it was rather a lot of bother, but `piplite` is now separated out as package, and the examples have been updated to at least reflect use of `piplite`. Further, did some optimizations of the demo workflow to not package wheels that are already packaged by pyodide... no doubt they've been patched for a good reason.

examples/jupyter_lite_config.json
bollwyvl: `0.9.1` is back live, again... we can now actually _drop_ all the in-notebook version pins, if we like, as `piplite` will take care of getting the correct version for us... but we can also hold off on that for a separate PR.

bollwyvl:
The amount of work to make wheels in `federated_extensions` is probably not that great, but it's a little hard to think about it with the code still in the codebase... otherwise, probably needs some more docs, but is probably getting pretty close now, provided we can deal with the various compromises needed to get to this point.

bollwyvl:
I think I'm happy with this for the time being... going too far down the _lemme `pip install` an env for you_ is probably out of scope, and wouldn't even work for some clutch use cases e.g "the sandbox": docs deploys of PRs where the package-under-documentation isn't on pypi yet.

app/jupyterlite.schema.v0.json
```diff
@@ -101,12 +101,6 @@
           "default": "./lab/build/themes",
           "$ref": "#/definitions/non-trailing-slash-uri"
         },
-        "pyodideUrl": {
```
jtpio, line 104: Nice to see that one removed from the top level :+1:

bollwyvl:
i know, per usual, this has gotten to be a beast of a PR but: I see happy little emoji, but no red-x or e-merge-y... do i need to break this up?

if we got this in, we'd be able to demonstrate a site that implemented #45 at the community call on 11/30... maybe even qualify as a non-alpha release at that point!

app/jupyterlite.schema.v0.json
bollwyvl: yeah, it was a bit of a haul to get there, but seems clearer. This is really the only thing that would constitute a `v1`, and would be _very_ easy to forward/back migrate... but anybody that has been customizing this already will be impacted by other changes in this PR even more.

once extracted is a dedicated extension, we can figure out some way to move the schema someplace over there that will make it to the built environment such that the `jupyter lite check` step can verify against sub-schema.

bollwyvl:
Over on #419, i've started a side investigation of self-hosted mathjax that has revealed some new `pip` and `flit` quirks in the build chain that i'll incorporate here... however, it may make sense to make self-hosted pyodide an opt-in thing on _this_ PR. While I usually like putting all the bells and whistles into the docs site, it might be that the ~170mb of `pyodide` is a bridge too far.

bollwyvl:
Kept fighting with `flit`, will just avoid `3.5.0` for now... https://github.com/takluyver/flit/issues/471

docs/configuring.md
```diff
@@ -48,43 +48,53 @@ schema-v0
...
 - have its theme information copied to `{output-dir}/{app/?}theme/`
 
+This discovery behavior can be disabled with the CLI flag `--ignore-sys-prefix` or
+`LiteBuildConfig/ignore_sys_prefix`.
+
 #### Extensions for a Specific App
 
 Similar to the above, by updating `$YOUR_JUPYTERLITE/{app}/jupyter-lite.json`, the
-federated extensions will only be avaialable for pages within that file tree.
+pre-built extensions will only be avaialable for pages within that file tree.
```
jtpio, line 97: Minor typo fix:

```suggestion
pre-built extensions will only be available for pages within that file tree.
```

packages/pyolite-kernel/py/piplite/piplite/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""A configurable Python package backed by Pyodide's micropip"""
+from .piplite import install
+
+__version__ = "0.1.0a16"
```
jtpio, line 4: Probably we need to add that one to bumpversion so it stays in sync with the other version numbers for now:

https://github.com/jupyterlite/jupyterlite/blob/b58ea7d358cc8b7438147812739b82e1cc43abc8/.bumpversion.cfg#L19

packages/pyolite-kernel/src/_pypi.ts
```diff
@@ -0,0 +1,8 @@
+// this file is autogenerated from the wheels described in ../package.json
+export * as allJSONUrl from '!!file-loader?name=pypi/[name].[ext]&context=.!../pypi/all.json';
+export * as ipykernelWheelUrl from '!!file-loader?name=pypi/[name].[ext]&context=.!../pypi/ipykernel-5.5.5-py3-none-any.whl';
+export * as nbformatWheelUrl from '!!file-loader?name=pypi/[name].[ext]&context=.!../pypi/nbformat-4.2.0-py3-none-any.whl';
+export * as pipliteWheelUrl from '!!file-loader?name=pypi/[name].[ext]&context=.!../pypi/piplite-0.1.0a16-py3-none-any.whl';
+export * as pyoliteWheelUrl from '!!file-loader?name=pypi/[name].[ext]&context=.!../pypi/pyolite-0.1.0a16-py3-none-any.whl';
```
jtpio, line 6: Looks like we would need to update the logic in the bump version script to handle this new location.

For now there is some basic search replace logic here:

https://github.com/jupyterlite/jupyterlite/blob/b58ea7d358cc8b7438147812739b82e1cc43abc8/scripts/bump-version.py#L44-L49

jtpio:
Thanks @bollwyvl this looks good!

I tried on RTD and also locally with the releaser assets from this run: https://github.com/jupyterlite/jupyterlite/actions/runs/1490564144

Left a couple of inline comments about versioning. We already have the issue of hardcoded version numbers in source files because of the wheel imports. Happy to take a look at it in a follow-up PRs to not block this one.

jtpio (approved):
Thanks!

PR #319, opened 2021-09-07
Add Jupyter Releaser config

### LINKED ISSUE #121: Release scripts

jtpio:
## References

Fixes #121 

Should also help address #7 (to check after a new release has been made with the releaser)

## Code changes

Add configuration to make the repo compatible with https://github.com/jupyter-server/jupyter_releaser

### TODO

- [x] Release hooks and config
- [x] Make sure we can build the Python package in the `py/jupyterlite` subfolder (not at the top-level)
- [x] Upload distributions built by the release as artifacts for the check release workflow
- [x] Bump version script
- [x] Use version to use for the changelog and tag? `0.1.0a6` or `0.1.0-alpha.6`? 
  - If we want the Python version we might have to fix `get_version` used by the releaser: https://github.com/jupyter-server/jupyter_releaser/blob/690b71130d99610f74c44f48746db9b21db57870/jupyter_releaser/util.py#L141-L148 
  - Otherwise we can go with the `npm` version read from the top-level `package.json` since that is handled by the releaser out of the box
- [x] Move integrity script to `dodo.py` and remove the `buildutils` package

## User-facing changes

None

None


jtpio:
Looking good with the Python and npm packages being uploaded in dry run at the end of the check release check:

image

The dist archive also seem to contain all the files needed:

image

jtpio:
>     * Otherwise we can go with the `npm` version read from the top-level `package.json` since that is handled by the releaser out of the box

This is what it looks like in the check release:

**Changelog**

image

**Tags**

image

jtpio:
cc @bollwyvl if you would like to have a look, also if you have some input on the few changes to the `doit` config.

If we are fine with having the top-level `npm` version used in the changelog and the GitHub tag (the next will be `0.1.0-alpha.7`), we should be able to go ahead and merge this. Then I'll cut an automated release with the releaser and hopefully everything will go through :crossed_fingers: 

> having the top-level `npm` version

Otherwise we could also format the changelog to convert to a PEP compliant version, and add more hooks and logic to push a custom tag. Probably implementing the custom get version option in the releaser would be better and easier long term anyway: https://github.com/jupyter-server/jupyter_releaser/issues/150

jtpio:
> If we are fine with having the top-level npm version used in the changelog and the GitHub tag (the next will be 0.1.0-alpha.7),

Looks like putting something in the form of `0.1.0a6` in the `package.json` `version` field is actually fine:

image

jtpio:
Let's try it!

jtpio:
So it looks like the Full Release workflow went through without hiccup for `v0.1.0a8` :tada: 

https://github.com/jtpio/jupyter_releaser/runs/3607208817?check_suite_focus=true

https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.0a8

PR #327, opened 2021-09-14
Lint changelog in `after-build-changelog`

jtpio:
## References

Follow-up to #319 

In #325 we notice the changelog is not linted and the CI check fails.

## Code changes

Releaser config.

## User-facing changes

None

None


jtpio:
Looks like it's now formatted:

image

https://github.com/jupyterlite/jupyterlite/pull/327/checks?check_run_id=3601093521

PR #335, opened 2021-09-15
Buffers are already bytes

dsblank:
I see that at some point in the past (and maybe for some situations today) the buffers had to be converted to bytes. In my situation (using ipywidgets.Image and setting the bytes from Python) they are already bytes. The changes in this PR preserve the previous fix, but doesn't apply when not necessary.

### DIFF
--- a/packages/pyolite-kernel/py/ipykernel/ipykernel/comm.py
+++ b/packages/pyolite-kernel/py/ipykernel/ipykernel/comm.py
@@ -40,10 +40,8 @@ class Comm:
         data = {} if data is None else data
         metadata = {} if metadata is None else metadata
         content = dict(data=data, comm_id=self.comm_id, **keys)
-        if ((buffers is not None) and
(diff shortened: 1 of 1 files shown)


dsblank:
@madhur-tandon What do you think about this proposed change?

madhur-tandon:
Hi @dsblank, Thanks for the PR.
I am not sure if we should only check for `hasattr()` for just the 1st element i.e. `buffers[0]`
So maybe we should do this check for each element.
And if it has it, convert it using `.tobytes()` and if not, append it as it is to the list

This should take care of cases where some elements are `bytes` object and some aren't.
Though I am not sure if we will encounter this case practically.
CC: @martinRenou

dsblank:
@madhur-tandon Thanks; makes sense! Updated PR.

martinRenou (approved):
LGTM :)

jtpio:
Thanks!

PR #336, opened 2021-09-16
Add support for localization

### LINKED ISSUE #317: Support for localization
### Problem

For now the JupyterLite UI is only available in English.

We should leverage the JupyterLab and Jupyter Server infrastructure to add support for localization, reusing the existing language packs.

### Proposed Solution

It looks like we'll have to place the locale content under `/api/translations/<locale>`, for example `api/translations/fr_FR`.

This endpoint should return the localization data as JSON:

image

Probably this will mean adding a new addon to: https://github.com/jupyterlite/jupyterlite/tree/main/py/jupyterlite/src/jupyterlite/addons

On the TS side, we remove the following plugin:

https://github.com/jupyterlite/jupyterlite/blob/a12f71c81a5c4bbb31e55bd61dcac358bbe63ba6/packages/application-extension/src/index.tsx#L361-L372

And replace it by the upstream extension.

jtpio:
## References

Fixes #317 

## Code changes

- [x] Add `@jupyterlite/translation` server package
- [x] Fetch the translation data from the server plugin
- [x] Add task to populate the `/api/translations/` endpoints with the locale data, fetched with:

```python
from jupyterlab_server.translation_utils import get_language_pack, get_language_packs

# /api/translations
get_language_packs()

# /api/translations/fr_FR
get_language_pack('fr_FR')
```
- [x] ~Or use the `jupyterlab.languagepack` entrypoint and avoid the optional dependency on `jupyterlab_server`~ -> maybe something to investigate later to only depend on `babel`
- [x] Update to latest retro (which adds support for localization)
- [x] Create one file per locale?
- [x] Clean up code
- [x] Make sure everything works fine for setups without extra language packs (only the default english)
- [x] Add tests / fix coverage
- [x] Add docs

## User-facing changes

When available, end users will be able to use lab and retro in other languages than English.

None


jtpio:
Looks like we'll need this PR to land upstream first: https://github.com/jupyterlab/jupyterlab/pull/11091

jtpio:
Looking good with retro on RTD: 

image

docs/environment.yml
```diff
@@ -35,4 +35,7 @@ dependencies:
   - pytest-cov
   - pytest-html
   - pytest-xdist
+  # language packs
+  - jupyterlab-language-pack-fr-FR
+  - jupyterlab_server >=2.8.1,<3
```
jtpio, line 40: Not sure why `jupyter_server` is not listed here to handle the example files on the demo site.

Maybe the example files are copied and indexed differently, need to check.

jtpio: Ah it's probably the chain of `myst-nb` -> `nbdime` -> `jupyter_server` like last time.

So we should probably list the dependency explicitly here so it's easier to know where it's coming from.

jtpio:
Looks good when no other packs are installed and `jupyterlab_server` is missing (like in JupyterLab):

image

jtpio:
This should be good to go now.

Will cut a new release right after.

PR #309, opened 2021-08-27
Add ipyvuetify example notebook

seidlr:
## Code changes
Adds notebook

## User-facing changes
None

## Backwards-incompatible changes
None

PR #344, opened 2021-09-22
Sync the demo site environment

jtpio:
## References

Follow-up to #309 

Update the demo environment with by recreating the `jupyterlite-dev` environment and running `doit env`.

Spotted when looking into #336.

## Code changes

Demo site config.

## User-facing changes

None

None

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -37,7 +37,7 @@ dependencies:
   - ipycytoscape ==1.2.2
   - ipyleaflet ==0.14.0
   - ipympl ==0.7.0
-  - ipyvuetify ==1.8.1
+  - ipyvue ==1.5.0
   - plotly ==5.1.0
   ### FEDERATED EXTENSIONS ###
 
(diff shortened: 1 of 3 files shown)


bollwyvl:
Again, sorry I haven't been able to drive #310 forward which would clean up many of this class of errors.

We _could_ add a CI task which attempted this (and dumped out what the new file would be) to make it a bit easier... but binder isn't that hard a way, either.

I don't remember what the notebook check was like, but do these versions need to be pinned, too?

jtpio:
> I don't remember what the notebook check was like, but do these versions need to be pinned, too?

I think it's related to this issue: https://github.com/jupyterlite/jupyterlite/pull/281

With newer packages pulled from PyPI, while the deployed federated extension is older.

jtpio:
This should be ready to go, looking good on RTD:

image

bollwyvl:
> With newer packages pulled from PyPI, while the deployed federated extension is older.

Right, that's why we've got hard `==` in a number of the notebooks... and check it in lint... if it _is_ version sensitive, we probably need that for the affected packages as well, until we _also_ cache the wheels.

PR #348, opened 2021-09-27
xeus 2.0 wasm based lua kernel

DerThorsten:


jtpio:
Thanks @DerThorsten, this looks very exciting!

We should probably start looking into #104. Then we could have this kernel (and eventually others too) live in the jupyterlite organization as a federated server extension (I can start a draft PR to add support for this).

Otherwise building this on the main repo might be too much for the readthedocs build and would complicate the dev setup.

bollwyvl:
Very cool! Started commenting here, but ended up moving the rant to https://github.com/jupyterlite/jupyterlite/issues/104#issuecomment-927950790

jtpio:
For reference current efforts are happening in https://github.com/DerThorsten/jupyterlite_xeus_lua

With `jupyterlite-xeus-lua` being developed as a federated server extension.

PR #352, opened 2021-09-27
Scaffolding for loading serverlite extensions

### LINKED ISSUE #104: Support federated serverlite extensions

jtpio:
## References

Fixes #104 

Still a very early draft for now, but opening early for visibility.

## Code changes

Add boilerplate and configuration for loading serverlite extensions such as new kernels.

- [x] Basic scaffolding for loading server extensions
- [x] ~Refactor some of the logic in `index.template.js` so it's almost the same for serverlite and federated lab extensions.~ -> use the same machinery
- [x] ~Add to the `jupyterlite` cli~ -> use the same machinery
- [x] Remove the p5 kernel from this repo
- [x] Move the `p5-kernel` outside of the main repo to its own repo under `jupyterite/p5-kernel`
- [x] Update docs
- [x] ~Handle disabled lite extensions~ -> already handled
- [x] Load serverlite extensions in retro

As a follow-up, we can:

- create a cookiecutter as a fork of the [upstream cookiecutter](https://github.com/jupyterlab/extension-cookiecutter-ts): https://github.com/jupyterlite/serverlite-cookiecutter-ts/pull/1
- add the `p5-kernel` to the demo template: https://github.com/jupyterlite/demo
- check how to handle shared packages added by `jupyter labextension build`. Probably nothing critical right now, but maybe we can use the [`--core-path`](https://github.com/jupyterlab/jupyterlab/blob/88ae23f798c2adfcaea6a8fd0385f23b8aa2b8a1/builder/src/build-labextension.ts#L54) pointing to the `jupyterlite` main `package.json`

## User-facing changes

Deployers of JupyterLite websites will be able to add new kernels that are developed as third-party extensions.

None


jtpio:
Manually copying the files from the p5-kernel prebuilt extension to `/lite/extensions`, looks like we can load the `remoteEntry` file:

image

jtpio:
> can we reuse _all_ of the existing `package.json#/jupyterlab/extension` metadata/detection (e.g. keeping them in `federated_extensions`) but then split them into `extensions` and `serverExtensions` on some _additional_ metadata, e.g. `package.json#/jupyterlite/liteExtension`?

Yes that sounds good. I'm currently iterating locally but will push the extracted `p5` repo to https://github.com/jupyterlite/p5-kernel soon. The good news is that the existing `jupyter labextension build` toolchain from the TS cookiecutter seems to be working well. So probably no change required there, although some tweaks could be done later if needed. For example we might not need the `@jupyterlab` packages to be added to the `share` webpack scope, only the `@jupyterlite` ones are needed for serverlite extensions.

jtpio:
Looks like we'll need this change to land upstream to properly handle the kernelspec logos: https://github.com/jupyterlab/jupyterlab/pull/11175

For now we can set an empty string so a default logo is generated: 

image

jtpio:
Maybe we can then even keep the `lite` extensions under the default `PREFIX/share/jupyter/labextensions` like other lab extensions. Sounds like that might be simpler at least to begin with.

And removing the `jupyterlab.extension` from the `package.json` and adding `jupyterlite.liteExtension` (or `jupyterlite.extension`) should be enough for the frontend to choose which extension to load. Since JupyterLab also checks `jupyterlab.extension` it will not try to load the lite extensions if they don't have the key.

That way we indeed keep an almost identical extension development workflow as regular federated lab extensions.

jtpio:
> removing the jupyterlab.extension

This seems to actually be required for the federated extension to properly be built by `jupyter labextension build`.

bollwyvl:
> identical extension development workflow

right, that's what i'm gunning for

> seems to actually be required for the federated extension to properly be built 

Yeah, there's just too much first- (core: builder, license), second- (us: settings, `lite build`) and third-party (???) stuff that is expecting `package.json#/jupyterlab/extension` (and `/lab/extensions`) that i think we need to accept them.

Some ideas:
- we _could_ require each `lite_extension` to offer a "real," no-op labextension
  - this isn't a bad as it seems, perhaps, as practically there _will_ be some sensitivity in semver, etc. to the client
- use our novel metadata in `#/jupyterlite` to pop all/some extensions/plugins from the client extension list
  - we may not need/want to support a single `npm` package offering both client and "server" plugins...
    - but if we did, i _guess_ we could do it by pkg/plugin ID, as... what's a file? e.g.
        ```yaml
       "name": "@org/pkg",
       "jupyterlab": {
          "extension": true
       },
       "jupyterlite": {
          "extension": true # for simplicity/parity, drop all, functionally equivalent to...
          "extension": ["@org/pkg"] # which is a special case of...
          "extension": ["@org/pkg:plugin-a", "@org/pkg:plugin-c"] # drop just one/some
       },
        ```

jtpio:
> create a cookiecutter as a fork of the [upstream cookiecutter](https://github.com/jupyterlab/extension-cookiecutter-ts)

Ongoing update happening in https://github.com/jupyterlite/serverlite-cookiecutter-ts/pull/1

jtpio:
First release of `jupyterlite-p5-kernel` is out: https://pypi.org/project/jupyterlite-p5-kernel/

Need to check how we can test it on RTD for this PR.

jtpio:
Nice, looking good on the RTD preview:

image

I guess we could package `jupyterlite-p5-kernel` for conda and add it to the list of `federated_extensions`. Or leave it out of the main repo for now and only include it on the demo repo (once `0.1.0a10` is out): https://github.com/jupyterlite/demo

bollwyvl:
>  we could package jupyterlite-p5-kernel for conda 

We can also update that part of the build chain to work with pip... and/or list it expliclity (since we know it doesn't have any dependencies).

jtpio:
> We can also update that part of the build chain to work with pip... and/or list it expliclity (since we know it doesn't have any dependencies).

Right, it looks like this is already supported actually: 

https://github.com/jupyterlite/jupyterlite/blob/81261504fc28b5107a89e5571c81cc643bc880c4/py/jupyterlite/src/jupyterlite/addons/federated_extensions.py#L129-L171

jtpio:
Looks like the wheel is nicely picked up:

image

jtpio:
I think this should be good to go now. It's nice that we can reuse the whole existing federated extension mechanism. Maybe the lab / serverlite naming might be confusing with server extensions going to `share/jupyter/labextensions`, but this is something we could address later if needed.

Will cut a new alpha release right after so we can start testing with other kernels like [`xeus-lua`](https://github.com/jupyter-xeus/xeus-lua).

cc @bollwyvl if you want to have a quick look

jtpio:
Let's merge and iterate. We can always follow-up later and cut patch releases.

PR #355, opened 2021-10-01
removed not needed methods from kernel interface

DerThorsten:
## References
As discussed in #348, xeus based kernel extension should be a federated server extension as https://github.com/jupyterlite/p5-kernel .
In this PR are the changes needed for the xeus based kernels 

## Code changes
the kernel interface should  only have the `handleMessage` function. 
The functions as `kernelInfoRequest` are not needed in the interace of the kernel.
In Xeus, the handling of the individual requests (like kernelInfoRequest) is done on the C++ side and it would be very cumbersome to implement `kernelInfoRequest` in js/ts  for a xeus based kernel.
Furthermore nothing in the whole jupyterlite code base seem to rely on `kernelInfoRequest` etc beeing in the interface. Therefore this PR just removes them

## User-facing changes
 - 

Kernel interface is changed


jtpio:
Thanks!

Indeed it makes sense to remove these methods for better flexibility for the kernel authors.

jtpio (approved):
Thank you!

PR #396, opened 2021-10-15
Better handle text files

### LINKED ISSUE #395: Fix handling of text files
## Description

This is an issue to track the following TODO:

https://github.com/jupyterlite/jupyterlite/blob/299d6c11e4926f76cba6d7eeb666a57b0bed1424/packages/contents/src/contents.ts#L367-L368

## Reproduce

The example `card.vue` opens in base64, although it's a text file:

image

This is also the case for `.js` files for example.

## Expected behavior

Text files should open with the text editor.

jtpio:
## References

Fixes #395

## Code changes

Improve the handling of text files fetched from the server.

## User-facing changes

Some files should now open in a text readable format (for example `.js`)

None

### DIFF
--- a/packages/contents/src/contents.ts
+++ b/packages/contents/src/contents.ts
@@ -27,25 +27,13 @@ const N_CHECKPOINTS = 5;
 /**
  * A list of mime types of text files
  */
-const TEXT_MIME_TYPES = new Set([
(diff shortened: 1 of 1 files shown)


jtpio:
This should already improve things.

We might want to revisit the approach or add extra mime types later.

PR #397, opened 2021-10-15
Add `@jupyterlab/javascript-extension`

jtpio:
## References

Since it was missing from the lab and retro builds.

## Code changes

Add the `javascript-extension`.

## User-facing changes

In Lab and Retro:

https://user-images.githubusercontent.com/591645/137548167-221ca797-7368-4677-911c-1f954e2ead28.mp4

None


bollwyvl:
Cool, this is clutch.

Perhaps instead of nickle-and-diming ourselves, we could capture the upstream extensions we definitely _don't_ want, and automate bringing in everything else...

PR #404, opened 2021-10-24
Handle 404s when getting contents

### LINKED ISSUE #370: Save as event will always trigger overwrite
## Description

When I try to save a file as a new one, jupyterlite will always prompt that "the file is already exists" even though it does not exist.
image

## Reproduce

File > Save Notebook As

jtpio:
## References

Fixes #370 

## Code changes

Slight rework of the server contents manager to:

- not implement the frontend `Contents.IManager`
- return 404 on missing file, so it can correctly be handled on the frontend
- make some methods private (fetching server files)

## User-facing changes

Should fix the "Save As" as described in #370.

The interface of the server contents manager changes (narrower).

### DIFF
--- a/packages/contents/package.json
+++ b/packages/contents/package.json
@@ -45,7 +45,6 @@
     "@jupyterlab/nbformat": "~3.2.0",
     "@jupyterlab/services": "~6.2.0",
     "@lumino/coreutils": "^1.8.0",
-    "@lumino/signaling": "^1.7.0",
     "localforage": "^1.9.0"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


jtpio:
Looks like there is still an issue when "Saving as" for a file has whitepaces in the name.

jtpio:
This seems to be fine on the preview:

https://user-images.githubusercontent.com/591645/140034429-5ac51356-67d7-4960-aa19-41b8bd1c7a69.mp4

jtpio:
Let's get it in. Will make a new release after to ship some of the other recent fixes as well.

PR #417, opened 2021-11-18
Add `pdf-extension`

jtpio:
## References

Add `pdf-extension` to the lab app.

## Code changes

Dependency update.

## User-facing changes

Users can open PDF files on the server:

image

None


bollwyvl:
This is very good, especially for docs sites.

It does raise the question again of whether/how much we can automate the
inclusion of upstream extensions... I guess we pretty much want everything
from jupyterlab/staging/package.json that isn't shadowed by a lite package
or explicitly ignored.

jtpio:
Agree. This was also the case recently with the `javascript-extension`.

Probably we can indeed rework how we pull dependencies from upstream, using an exclude list instead. Although there are not that many new packages added to JupyterLab so these kinds of updates should not happen that often.

bollwyvl (approved):
LGTM

PR #419, opened 2021-11-21
Self-hosting MathJax 

bollwyvl:
## References

- includes/blocked by #310
- continues towards a baseline for #45

### CLI build time
- adds a build-time MathJax addon which will bundle mathjax assets if 
  - provided by `--mathjax-dir` 
  - installed next to `jupyter-server-mathjax` (obeys `--ignore-sys-prefix`)

### our build/docs
- includes `--mathjax-dir`, because we use `--ignore-sys-prefix`

## User-facing changes

- even fewer CDN requests!

- `jupyter-server-mathjax` only ships the default configuration... if a user has that package installed and configured `mathjaxConfig` some other way, it won't find the assets
  - this is verified in `check`, but only for `lite`-managed assets (not custom CDN)


bollwyvl:
This looks ok on RTD.

jtpio:
Thanks @bollwyvl!

Mind updating that PR now that #310 has been merged?

bollwyvl:
Hm, I didn't add any docs for this... probably needs (automation of) it...

jtpio (approved):
Thanks!

jtpio:
> Hm, I didn't add any docs for this... probably needs (automation of) it...

Would you like to do this as part of this PR? Otherwise we can track that separately.

bollwyvl:
> Would you like to do this as part of this PR? Otherwise we can track that separately.

Working on tests/docs presently...

bollwyvl:
Woof, that went a bit deeper than expected, but so it goes. No further changes planned!

bollwyvl:
I don't know what to do about how long RTD takes...

jtpio:
> I don't know what to do about how long RTD takes...

Do you think it takes longer to build on RTD? The logs show a build time similar to the version on `main`:

image

bollwyvl:
Guess it just seemed longer last time. There are hundreds of files in the
mathjax distribution, of course... guess that's the price for durability!
Everything seems fine now.

jtpio:
Thanks!

PR #421, opened 2021-11-23
Version piplite with the `bump-version` script

### LINKED ISSUE #420: Version `piplite` with the `bump-version` script
As part of the automated release process, we use a script to bump the version of the Python and JS packages: https://github.com/jupyterlite/jupyterlite/blob/main/scripts/bump-version.py

As a follow-up to https://github.com/jupyterlite/jupyterlite/pull/310, we can update the logic to make sure we update the hardcoded `0.1.0ax` version numbers.

There is already a script that replaces the version in the source file:

https://github.com/jupyterlite/jupyterlite/blob/503682820eafee4d23ee0603caf8f51a4c4ae259/scripts/bump-version.py#L45-L50

So we can probably update it so it to take into account the new locations.

jtpio:
## References

Fixes #420 

## Code changes

Update the `bump-version.py` script to version the packages, as a follow-up to #310.

## User-facing changes

So the following code snippet:

```python
import piplite
piplite.__version__
```

Gives a consistent version.

None

### DIFF
--- a/scripts/bump-version.py
+++ b/scripts/bump-version.py
@@ -25,6 +25,7 @@ PYOLITE_PACKAGE = ROOT / "packages" / "pyolite-kernel"
 PYOLITE_PACKAGE_JSON = PYOLITE_PACKAGE / "package.json"
 PYOLITE_KERNEL_SOURCE = PYOLITE_PACKAGE / "src" / "kernel.ts"
 PYOLITE_INIT_PY = PYOLITE_PACKAGE / "py" / "pyolite" / "pyolite" / "__init__.py"
+PIPITE_INIT_PY = PYOLITE_PACKAGE / "py" / "piplite" / "piplite" / "__init__.py"
 
 
 def replace_in_file(path, pattern, replacement):
@@ -50,10 +51,14 @@ def postbump():
     replace_in_file(
         PYOLITE_INIT_PY, "__version__ = (.*)", f'__version__ = "{py_version}"'
     )
+    replace_in_file(
+        PIPITE_INIT_PY, "__version__ = (.*)", f'__version__ = "{py_version}"'
+    )
 
     # bump pyolite version
     pyolite_json = json.loads(PYOLITE_PACKAGE_JSON.read_text(**ENC))
     pyolite_json["pyolite"]["packages"]["py/pyolite"] = py_version
+    pyolite_json["pyolite"]["packages"]["py/piplite"] = py_version
     PYOLITE_PACKAGE_JSON.write_text(json.dumps(pyolite_json), **ENC)
     run(f"yarn prettier --write {PYOLITE_PACKAGE_JSON}")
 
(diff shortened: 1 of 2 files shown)


jtpio:
Looks good testing locally with the releaser assets:

image

scripts/bump-version.py
```diff
@@ -23,8 +23,9 @@
 
 PYOLITE_PACKAGE = ROOT / "packages" / "pyolite-kernel"
 PYOLITE_PACKAGE_JSON = PYOLITE_PACKAGE / "package.json"
-PYOLITE_KERNEL_SOURCE = PYOLITE_PACKAGE / "src" / "kernel.ts"
+PYPI_SOURCE = PYOLITE_PACKAGE / "src" / "_pypi.ts"
```
bollwyvl, line 26: hmmm... this file gets generated.

scripts/bump-version.py
```diff
@@ -23,8 +23,9 @@
...
 PYOLITE_INIT_PY = PYOLITE_PACKAGE / "py" / "pyolite" / "pyolite" / "__init__.py"
+PIPITE_INIT_PY = PYOLITE_PACKAGE / "py" / "piplite" / "piplite" / "__init__.py"
```
bollwyvl, line 28: `PIPLITE`?

scripts/bump-version.py
bollwyvl: The real source of truth migh tbe: https://github.com/jupyterlite/jupyterlite/blob/main/packages/pyolite-kernel/package.json#L69 (unless that's covered above)

scripts/bump-version.py
```diff
@@ -43,17 +44,26 @@ def postbump():
...
         PYOLITE_INIT_PY, "__version__ = (.*)", f'__version__ = "{py_version}"'
     )
+    replace_in_file(
+        PIPITE_INIT_PY, "__version__ = (.*)", f'__version__ = "{py_version}"'
+    )
 
     # bump pyolite version
     pyolite_json = json.loads(PYOLITE_PACKAGE_JSON.read_text(**ENC))
     pyolite_json["pyolite"]["packages"]["py/pyolite"] = py_version
+    pyolite_json["pyolite"]["packages"]["py/piplite"] = py_version
```
bollwyvl, line 66: right, is covered here

jtpio:
Also since this string search / replace is a bit brittle, it should probably just fail it if fails to do the replace. So the check release CI check would be able to catch future changes.

scripts/bump-version.py
jtpio: Maybe this one is not needed actually, gonna double check.

bollwyvl: This gets regenerated by `doit build:js:piplite:wheels`, so if that happens after this runs...

jtpio:
This seems to be looking good now. Let's merge and I'll make a release later today.

PR #424, opened 2021-11-27
Fix false positive for non-prebuilt extensions in wheels, MathJax check

### LINKED ISSUE #422: pre-built extension wheels with non-extension package.json
## Description

When trying to include a wheel (and probably a conda package) in it, extra `package.json` (e.g. vendored ones inside extensions) can be detected, and the build will fail.

## Reproduce

- In a pre-built extension, create and ship a random `package.json` in an unexpected place
- see the `jupyter lite build` fail on `pre_build:federated_extensions:extract`
  - e.g. `
Two different tasks can't have a common target.'/home/runner/work/ipydrawio/ipydrawio/build/demo/lab/extensions/@deathbeds/ipydrawio-webpack/static/dio/CITATION.cff' is a target for pre_build:federated_extensions:extract:wheel:@deathbeds/ipydrawio-webpack/static/dio and pre_build:federated_extensions:extract:wheel:@deathbeds/ipydrawio-webpack.
`

## Expected behavior

- Does not fail

## Context

- JupyterLite version: 0.1.0a16
- Operating System and version: linux
- Browser and version: n/a

## Design ideas

Before extracting a whole extension from a wheel/conda package, validate that `package.json` contains both `#/jupyterlab/_build/load` and that the file it points to actually exists.

bollwyvl:
## References

- fixes #422
- additionally addresses problems in new checks from #419
- both noticed while trying to adopt some of these new features in https://github.com/deathbeds/ipydrawio/pull/74

## Code changes

- [x] checks `package.json` files to see if they are actually federated extensions in wheels and conda packages
- [ ] fix path for mathjax check (was getting extra `MathJax.js` appended)

## User-facing changes

- extra `package.json` will not trigger extra copies
- custom MathJax URLs will pass `check`

- n/a

### DIFF
--- a/py/jupyterlite/src/jupyterlite/addons/federated_extensions.py
+++ b/py/jupyterlite/src/jupyterlite/addons/federated_extensions.py
@@ -35,7 +35,7 @@ class FederatedExtensionAddon(BaseAddon):
                 *root.glob(f"*/{PACKAGE_JSON}"),
                 *root.glob(f"@*/*/{PACKAGE_JSON}"),
             ]
-            if self.is_prebuilt(json.loads(p).read_text(**UTF8))
+            if self.is_prebuilt(json.loads(p.read_text(**UTF8)))
         ]
 
     @property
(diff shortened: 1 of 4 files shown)


jtpio (approved):
Thanks!

PR #426, opened 2021-12-01
Bump Python packages with `tbump`

jtpio:
## References

Follow-up to #421, so the bump is less brittle.

## Code changes

- [x] Bump `pyolite` and `piplite` with `tbump`
- [x] Fixes `package.json` files not being formatted by prettier
- [x] Reformat existing `packages.json` with prettier

Keeping `bump2version` for now to handle `patch`, `minor`, ... specs easily, and derive the other versions from the main `jupyterlite` version.

## User-facing changes

None

None

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,7 +1,6 @@
 **/_output
 **/lib
 **/node_modules
-**/package.json
 **/static
 
 .yarn-packages
--- a/.prettierignore
+++ b/.prettierignore
@@ -21,3 +21,6 @@ _pypi.ts
 
 # The changelog is updated by the releaser
 CHANGELOG.md
+
+# UI tests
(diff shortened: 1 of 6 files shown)


bollwyvl:
Hooray, i guess?

While i like some `flit`... i do wish it supported reading a `VERSION` file like the bad old days of `setup.cfg`... just feels icky updating code.

jtpio:
Right, `tbump` is quite agnostic which is good.

Hoping to drop `bump2version` some time soon to keep only one way for bumping Python packages, once https://github.com/jupyter-server/jupyter_releaser/issues/213 or similar is available.

jtpio:
Looking good when checking the dist files locally.

Issue #168, opened 2021-06-22
Reorganize the examples directory


jtpio:
Now might be a good time to reorder notebooks in the examples directory

_Originally posted by @martinRenou in https://github.com/jtpio/jupyterlite/issues/156#issuecomment-863861029_

bollwyvl:
At present, we have a lot of vertical space, but the default view is rather narrow (we could default it to larger, perhaps, but might need workspaces #139). Also, since we're thinking of renaming pyolite, I propose we just go with the canonical file extension, (e.g. `js`, `py`) a top-level group, and an (optional) specific like:

```
- _GALLERY_.ipynb
- js - basics.ipynb
- p5 - basics.ipynb
- py - basics.ipynb
- py - widgets.ipynb
- py - widgets - bqplot.ipynb
- py - widgets - leaflet.ipynb
```

The idea on the `_GALLERY_`: might need #139 (or we can fudge it), but having a default opening page that showed pictures of all the notebooks would be super helpful: if you don't know what `ipyleaflet` is, a picture of a map would get the point across quickly. for now, we could hand-jam it, but eventually (#6) this entire file could be generated, including screenshots.

psychemedia:
A quick sketch of a possible new structure (the data dir should check relative file loading and also make the same data file more sensibly available across kernel subdir demos), but I ran into issues with the `jupyter lite build` not behaving as expected (`ipywidgets` fail to render?). I'd quite like to make sure any PR on reorganised examples meshes with an updated `jupyterlite-demo` walkthrough that I'll do alongside, but as it stands, my local demo fails on the `ipywidget` containing notebooks. I also need to sign off for the w/e now, so will have to return to this next week...

![image](https://user-images.githubusercontent.com/82988/124273512-04327800-db38-11eb-9a88-c63b4079ecc6.png)

jtpio:
> I ran into issues with the `jupyter lite build` not behaving as expected (`ipywidgets` fail to render?)

Right, looks like this is tracked in https://github.com/jtpio/jupyterlite/issues/209

jtpio:
Maybe we could start offloading some of these dependencies and examples to https://github.com/jupyterlite/demo, to avoid increasing the build time on the ReadTheDocs previews?

[...]

PR #432, opened 2021-12-21
[DOCS] Adding introductory documentation to the launch buttons

choldgraf:
## References

In preparation for adding a retrolab tutorial to the top row of try.jupyter.org, I thought it'd be helpful to add two quick introductory notebooks for the Lab / Retrolab buttons on the JupyterLite docs. This way, users see some content when they click the button, and can have a few pointers in the right direction. This may be a short-term solution until we can flesh out some more introductory content that is fitting for try.jupyter.org, but what do people think about this as a first step?

- https://github.com/jupyter/jupyter.github.io/issues/513
- https://github.com/jupyterhub/team-compass/issues/463

## Documentation changes

- There are now two lightweight "introductory" notebooks that are meant to guide people on their first steps with the interfaces.
- The buttons in the top left of the docs now open an `intro` notebook when JupyterLite is loaded.


bollwyvl:
This is good content.

Given the size breakdown on that other issue, I don't think we'd necessarily want to _always_ point directly at a particular python notebook with its 30mb load as the very first introduction. If it was a notebook, a JS kernel notebook would be _far_ lighter weight, and is more likely to stay in this repo, and could still make some pretties with e.g. vega. However, more likely, we'd want to rework the prose into steps of a tour, and start that off.

Over on https://ipydrawio.readthedocs.io and https://pidgy.readthedocs.io i've played around with some other approaches to flexibly launching the demo with some URL-based configuring... perhaps we consider some alternate forms, including offering all of the examples as potential starting points.

More broadly: even if we _did_ bring the capability into `try.jupyter.org`, this project's `main` is not ready to be the destination, even if the released package could build a site, deployed on `try.jupyter.org/lite/`, for example, that was. We do what we can to keep it working here, but you know...

choldgraf:
Gotcha - makes sense to me. Do you think we should close this PR? Or rework the content somehow? I suspect that many people will land on the jupyterlite docs and just click one of the try buttons in the top left, since that's a high eyeballs spot, and might be confused at the lack of context.

I didn't realize size was a concern (WASM and jupyterlite continues to work in mysterious ways to me haha). Maybe it should just be a markdown file?

jtpio:
Thanks @choldgraf!

> Do you think we should close this PR? Or rework the content somehow? 

Probably we can still keep the `intro.ipynb` and `intro-retro.ipynb` as they would show up in the file browser at the top-level.

Another idea would be to open `examples/README.md` by default with links to these notebooks. And iterate a bit on the `README.md` to make it more like a proper gallery of examples (could be done at a later stage).

examples/README.md
```diff
@@ -1,4 +1,8 @@
 # JupyterLite Examples
 
 This folder contains example notebooks and files demoing what is possible to do with
-JupyterLite.
+JupyterLite. To get started, there is a short introduction in [the `intro.ipynb` notebook](intro.ipynb).
```
jtpio, line 4: Pushing a commit to fix the lint failure.

jtpio: Done in https://github.com/jupyterlite/jupyterlite/pull/432/commits/6a054a618bbe59bccf89eb9964216e130e51dba0

choldgraf:
> Another idea would be to open examples/README.md by default with links to these notebooks. And iterate a bit on the README.md to make it more like a proper gallery of examples (could be done at a later stage).

This sounds like a nice idea - let me know if you'd like this in the current PR or not.

jtpio:
@choldgraf if you would like to do it in this PR that would be great, thanks!

choldgraf:
ok - can you give me an idea of what kind of content you think is best there? Something like a few paragraphs about JupyterLite in general with pointers to the intro notebooks? Anything else?

jtpio:
That would already be a good introduction yes.

There was also some previous discussions about this in https://github.com/jupyterlite/jupyterlite/issues/168 and related issues.

jtpio:
Just rebased onto `main` to grab the latest updates.

Testing on RTD, opening a notebook by default doesn't feel too heavy at first. Although it's true some dependencies are pulled, and the first cell run still takes some time but it's a bit faster now thanks to https://github.com/jupyterlite/jupyterlite/pull/433.

jtpio:
However there seems to be a `micropip` related issue when opening the notebook with the new Pyodide 0.19 and lite `0.1.0a19`:

image

But this does not seem to be a problem when testing on https://jupyterlite.github.io/demo/, which doesn't pre-cache the wheels:

image

jtpio:
> However there seems to be a micropip related issue when opening the notebook with the new Pyodide 0.19 and lite 0.1.0a19:

Ah nevermind that, this was most likely a caching issue. Trying again in a private window seems to work fine.

docs/_templates/launch.html
```diff
@@ -13,7 +13,7 @@
 </a>
 
 <a
-  href="{{ pathto('_static/retro/index.html', 1) }}"
+  href="{{ pathto('_static/retro/', 1) }}?path=notebooks/intro-retro.ipynb"
```
jtpio, line 16: Looks like removing the `index.html` here creates an infinite redirect loop?

https://user-images.githubusercontent.com/591645/149344871-68ecb47f-2b17-4d43-a8e0-fa2a6e367966.mp4

bollwyvl: yeah, different hosts react differently with respect to `index.html`, trailing slashes, etc. and `index.html` is the only one that's going to _work_ work. That's why the rediraffe stuff in the first place for "pretty" urls.

bollwyvl:
> Format intro notebooks

I've been wondering: we should probably add a no-op `nbconvert --to notebook` to our linter, as the "one line of JSON" is _even less_ ergonomic than a regular ipynb :rofl:

docs/_templates/launch.html
jtpio:
```suggestion
  href="{{ pathto('_static/retro/index.html', 1) }}?path=notebooks/intro-retro.ipynb"
```

docs/_templates/launch.html
```diff
@@ -13,7 +13,7 @@
 </a>
 
 <a
-  href="{{ pathto('_static/retro/index.html', 1) }}"
+  href="{{ pathto('_static/retro/index.html', 1) }}?path=notebooks/intro-retro.ipynb"
```
jtpio, line 16: The format is slightly different for retro:

```suggestion
  href="{{ pathto('_static/retro/notebooks/index.html', 1) }}?path=intro-retro.ipynb"
```

jtpio:
Since this is good content, I would be in favor of getting this in.

We can keep the demo link in the project description of the repo (and similar links) as it is now without the `?path=`:

image

cc @bollwyvl since you had a comment about loading times above

choldgraf:
@jtpio I can probably put a quick cycle into this (say, 30 minutes) if I have a really clear idea of what new content needs to be in here, but probably don't have time to do more complex iterations. Let me know if you'd prefer to just merge it in, or if I should make some changes first

jtpio:
Thanks @choldgraf.

> Another idea would be to open examples/README.md by default with links to these notebooks. And iterate a bit on the README.md to make it more like a proper gallery of examples (could be done at a later stage).

Actually the `README` opens with the text editor by default for now and not the markdown preview, which makes opening that file with `?path=` less interesting I believe. Although we could override the preferred factory for markdown files to open with the markdown preview on the demo ReadTheDocs deployment.

Which is why we could just keep the current content as it is with this PR, and iterate later if needed.

jtpio (approved):
Thanks!

choldgraf:
thanks for the iteration and suggestions all :-)

Issue #259, opened 2021-07-14
Tab completion does not work with Pyolite kernel


drdarshan:
## Description
First of all - a big Thank You for this amazing project! It's magical when things like NumPy, Pandas and MPL "just work" without having to install anything. 
  
In "vanilla" Jupyter with Python kernel and IPython, hitting the tab character gives me completions at the given point. This seems to not work with Pyolite. More precisely, I tried this:
```
[1]: import sys
[ ]: sys.<TAB>
```
I expected to see everything under `sys` but nothing was displayed. Is tab completion expected to work with Jupyterlite/Pyolite? 

## Expected behavior
Expected to see a overlay with completions like this:
<img width="428" alt="image" src="https://user-images.githubusercontent.com/875518/125668120-0e30d7f9-8927-42b0-8a48-c0ae87de5423.png">

## Context

- Operating System and version: Mac OS 11.4
- Browser and version: Firefox 89 and Chrome 91
- JupyterLite version: 0.1.0-alpha-5

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

jtpio:
Thanks @drdarshan for the report.

Normally this should work since #142. Maybe something changed recently.

cc @martinRenou @madhur-tandon who have been doing related work

martinRenou:
This is expected to work. It seems that `sys.<tab>` does not complete while `from sys import <tab>` completes. This is quite odd, there might be something wrong in our usage of IPython's completion logic.

![completion](https://user-images.githubusercontent.com/21197331/125758705-2b63bb58-bcde-4964-9ed2-e4c8b61fe743.png)

qqdaiyu55:
Same with @martinRenou . `sys.<tab>` does not work

qqdaiyu55:
This issue hurts debugging experience
@martinRenou  @jtpio Any update for this issue thanks

I'm not familiar with auto-completion implementation, I may take a look when I got time.

qqdaiyu55:
We can fix by disable JEDI...
![image](https://user-images.githubusercontent.com/9846706/149466850-1b99edd7-8a81-45be-a5f8-3b425dcf591f.png)

qqdaiyu55:
Related issue:
https://github.com/ipython/ipython/issues/12887

drdarshan:
[...]

PR #448, opened 2022-01-14
Disable jedi to fix tab completion issue

qqdaiyu55:
Related issue: https://github.com/jupyterlite/jupyterlite/issues/259


jtpio:
Thanks!

cc @martinRenou who might know if this is related to a change upstream

martinRenou:
As said by @qqdaiyu55 in #259, this could be related: ipython/ipython#12887

I know that people have been complaining in the past that the completion does not behave properly in IPython when jedi is installed. I believe it's not jedi's fault, but the usage of jedi in IPython. Anyway I guess disabling jedi is good 👍🏽

martinRenou:
> I know that people have been complaining in the past that the completion does not behave properly in IPython when jedi is installed

See https://github.com/ipython/ipython/issues/11530

martinRenou:
Thanks @qqdaiyu55!

Issue #440, opened 2022-01-08
Notebooks are loaded from another instance if multiple instances are hosted on the same domain


leukipp:
## Description

Hosting multiple JupyterLite instances on the same domain (e.g. GitHub pages) leads to a weird behavior for equally named files.
Completely different .ipynb files from instance A will be loaded when you afterwards visit instance B.

## Reproduce

1. Go to project [https://leukipp.github.io/jupyterlite/lab/](https://leukipp.github.io/jupyterlite/lab/), forked from the demo repository.
2. Open `python.ipynb`, make some changes and save.
3. Go to another project [https://leukipp.github.io/xmastree/lab/](https://leukipp.github.io/xmastree/lab/), which also contains a file named `python.ipynb`.
4.  Open `python.ipynb`, the content from the previous visited notebook is loaded.

## Expected behavior

Notebook states should be loaded correctly, regardless of the previously visited instances.

## Context

- JupyterLite version: v0.1.0a18

## Possible Solutions

- Use different indexedDB database names per JupyterLite root path (currently always `JupyterLite Storage` is used) 
- Use relative filepaths inside indexedDB database (e.g. /jupyterlite/lab/... or /xmastree/lab/...)

bollwyvl:
I feel like in many cases (and probably the default), letting user-created data persist between different apps, deployments on the same domain is probably fine... we generally try to err on the side of giving the _end user_ more power unless it specifically conflicts with the ability to create a stable system. Certainly, if two apps are deployed that have different extensions that would make some notebooks (for example) not work at all would require doing something directly about this.

> different indexedDB database names per JupyterLite

A few options:
- make `appName` drive the [`STORAGE_NAME`](https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/contents.ts#L16)
- have a new key, e.g. `localStorageName`
- some combination of the two (prefer `localStorageName`, fall back to `appName`)

That last option sounds good, as it would keep the existing behavior, by default, allow changing the `appName` to get some isolated content (and tabs, etc), and allowing multiple apps to share the same contents.

> relative filepaths inside indexedDB

Relative paths are pain: a lot of our complexity in e.g. `config-utils.js` that i'd _love_ to get rid of is around trying to _avoid_ the apps knowing anything about relative paths... it's just very hard to predict how they will be resolved in the face of trailing slashes, etc. on different hosts.

leukipp:
[...]

PR #449, opened 2022-01-15
Allow contents and settings storage names to be configured

bollwyvl:
## References

- partially addresses #440

## Code changes

- allows the local storage names to changed independently for settings and contents, falling back to the default of `JupyterLite Storage`

## User-facing changes

- if a site does not configure a `*StorageName`, no change in behavior
- if a site is depoyed, used by users and then re-deployed with a new value, all the existing settings and contents storages will apparently disappear

- n/a


bollwyvl:
Also `pathspec` stopped coming up... don't know what that means...

bollwyvl:
I've verified this works locally... but don't think we want to, for example, change the storage key on the docs site.

jtpio:
Thanks!

> but don't think we want to, for example, change the storage key on the docs site.

Indeed, we can keep the default one as it is now.

packages/server-extension/src/index.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+import { PageConfig } from '@jupyterlab/coreutils';
```
jtpio, line 4: Looks like `@jupyterlab/coreutils` should then be added to the `dependencies`:

https://github.com/jupyterlite/jupyterlite/blob/21f072b3b57bc67973bf36312b6a4ddffafcfede/packages/server-extension/package.json#L40-L46

jtpio: At some point it would be great to be able to reuse some of the JupyterLab integrity scripts to ensure dependencies are in sync and catch this for us.

jtpio: thanks!

jtpio (approved):
Looks good, thanks!

Issue #284, opened 2021-07-21
FASTA example is failing


bollwyvl:
## Description

The fasta document viewer had been failing for a while, but as of #283, the mime renderer is also failing.

## Reproduce

- open the browser console
- open the [RTD build for #283](https://jupyterlite--283.org.readthedocs.build/en/283/_static/lab/index.html) 
- double click `fasta-example.fasta`
- see:
```
Uncaught TypeError: t.on is not a function
    resetSeqs 681.80b3439aec1db6dfd35f.js:2
    e 681.80b3439aec1db6dfd35f.js:2
    resetSeqs 681.80b3439aec1db6dfd35f.js:2
    i 681.80b3439aec1db6dfd35f.js:2
    _loadSeqs 681.80b3439aec1db6dfd35f.js:2
    initialize 681.80b3439aec1db6dfd35f.js:2
    a 681.80b3439aec1db6dfd35f.js:2
    i 681.80b3439aec1db6dfd35f.js:2
    i 681.80b3439aec1db6dfd35f.js:2
    l 568.761dbc3966ac7bd30a96.js:1
    createRenderer 568.761dbc3966ac7bd30a96.js:1
    createNewWidget mimedocument.js:204
    createNew default.js:349
    createWidget widgetmanager.js:62
    _createOrOpenDocument manager.js:465
    open manager.js:281
    openOrReveal manager.js:305
    handleOpen listing.js:862
    evtDblClick listing.js:943
    handleEvent listing.js:562
    onAfterAttach listing.js:606
    processMessage index.es6.js:1387
    c index.es6.js:442
    e index.es6.js:178
    attachWidget index.es6.js:2773
    insertWidget index.es6.js:2673
    addWidget index.es6.js:2642
    set useFuzzyFilter browser.js:146
    activate index.js:138
    promise callback*activate/< index.js:116
    promise callback*activate index.js:104
    promise index.es6.js:162
    promise callback*76219/a</e.prototype.activatePlugin index.es6.js:161
    i index.es6.js:281
    start index.es6.js:280
    60880 index.js:488
    60880 index.js:494
    Webpack 6
681.80b3439aec1db6dfd35f.js:2:238290
 ```

## Expected behavior

- See little colored squares and no error

## Options

- remove the example in the short term
- investigate upstream

jtpio:
Thanks for catching this :+1: 

This looks like an issue upstream, as it seems to be failing with this gist:

[...]

PR #453, opened 2022-01-18
Update to `jupyterlab-fasta==3.2.0` in docs app

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/284

## Code changes

Dependency update for the RTD demo app.

## User-facing changes

Fasta example should be working

None


jtpio:
Should be back now:

image

Issue #398, opened 2021-10-18
Share a link to a file from the Notebook interface


jtpio:
### Problem

#384 added a way to share a link to multiple files in lab:

![image](https://user-images.githubusercontent.com/591645/137783800-2af23de8-5052-4da5-966b-5858b8968db0.png)

"Copy Shareable Link" is also available from the retro file browser: 

![image](https://user-images.githubusercontent.com/591645/137784125-10150d04-5ce9-4f95-b568-cc21c07f5abe.png)

But it generates a link that redirects to `/lab`. Ideally the link generated from the RetroLab interfaces should redirect to `/retro`.

### Proposed Solution

Probably it would be enough to iterate on the existing plugin for now:

https://github.com/jupyterlite/jupyterlite/blob/52944e9cd21785291d0975a63f6723038efa9525/packages/application-extension/src/index.tsx#L445-L494

This is where `lab` is hardcoded in the URL: https://github.com/jupyterlite/jupyterlite/blob/52944e9cd21785291d0975a63f6723038efa9525/packages/application-extension/src/index.tsx#L475

### Additional context

As mentioned in related issues and PRs, it would be great to have a more advanced way for sharing files / workspaces (some of it could be done upstream).

bollwyvl:
Some discussion over on https://github.com/jupyterlite/jupyterlite/pull/457#issuecomment-1017079021

bollwyvl:
> This is where `lab` is hardcoded in the URL: 

In retro, does `appUrl` point at `/retro`?

jtpio:
> In retro, does `appUrl` point at `/retro`?

I think it's still `/lab` at the moment:

![image](https://user-images.githubusercontent.com/591645/152402354-1c407014-d237-4bc2-afa9-b770a05ebb82.png)

bollwyvl:
Hm... well it looks like upstream in lab, it drives `workspaces` (which we don't have) and `settings` (which we do... but implement quite differently)... and is _not_ used for the share link. So i guess `app` means "the host web server app," not "this in-browser app".

Also, semi-related: 
https://github.com/jupyterlab/retrolab/issues/344

Issue #382, opened 2021-10-11
Copy Shareable Link


jtpio:
### Problem

The "Copy Shareable Link" command is available but does not redirect to the document in the link:

https://user-images.githubusercontent.com/591645/136756438-54d6307a-451f-4f37-b7ce-677fedf9af16.mp4

### Proposed Solution

There could be a plugin similar to the one in lab: https://github.com/jupyterlab/jupyterlab/blob/e6bf58139ac6ceeac2ea00d8f22dcbd8e09f1a1b/packages/filebrowser-extension/src/index.ts#L489-L525

And use query string parameters to specify the path to the file (`?path=example.ipynb`).

We can also implement opening in the simple interface when the url is `/doc/tree`.

### Additional context

Follow-up to #380.

PR #457, opened 2022-01-20
[wip] Use appUrl and file name for shareable link

### LINKED ISSUE #450: "Copy Shareable Link" issue with browser go-back function
## Description

Following a link created by the "Copy Shareable Link" command (#382) breaks the window history back functionality.

## Reproduce

1. Go to [https://jupyterlite.readthedocs.io/en/latest/_static/lab?path=intro.ipynb](https://jupyterlite.readthedocs.io/en/latest/_static/lab?path=intro.ipynb).
2. Press the back button, the tab `intro.ipynb` is loaded again.

## Expected behavior

If you follow the link above and then hit the back button, you should be taken back to this GitHub page.

## Context

- JupyterLite version: v0.1.0-alpha.19

bollwyvl:
## References

- fixes #450
- follow-on to #382
- might address #398

## Code changes

- uses `{appUrl}/index.html` (where `appUrl` is usually `lab`) for the full link

## User-facing changes

- the back button should work after redirecting (should be able to test from the rtd preview)

- n/a

## Future work

We might want to open up the API to support making the shareable params configurable. For example, jupyter-video chat offers the `?jvc=room` feature, and it would make sense to be able to share that room state as well. 

This also initially supported the command palette, but I've rolled this back, as I didn't realize it only applied to the selection from the file tree, not the main area. This would probably be a useful feature in its own right, but would have a different name, e.g. _Copy Workspace Link_ or something along those lines, and is probably out of scope

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -919,6 +919,7 @@ class L:
     ALL_ESLINT = _clean_paths(
         P.PACKAGES.rglob("*/src/**/*.js"),
         P.PACKAGES.rglob("*/src/**/*.ts"),
+        P.PACKAGES.rglob("*/src/**/*.tsx"),
     )
     ALL_JSON = _clean_paths(
         P.PACKAGE_JSONS,
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -518,7 +518,7 @@ const shareFile: JupyterFrontEndPlugin<void> = {
           return;
         }
 
-        const url = new URL(URLExt.join(PageConfig.getBaseUrl(), 'lab'));
+        const url = new URL(URLExt.join(PageConfig.getBaseUrl(), 'lab', 'index.html'));
         const models = toArray(
           filter(widget.selectedItems(), item => item.type !== 'directory')
         );
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -498,13 +498,11 @@ const opener: JupyterFrontEndPlugin<void> = {
 const shareFile: JupyterFrontEndPlugin<void> = {
   id: '@jupyterlite/application-extension:share-file',
   requires: [IFileBrowserFactory, ITranslator],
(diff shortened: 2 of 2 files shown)


bollwyvl:
Hmmm... it looks like this is enabled in the retro tree, but points back to `/lab/`. It appears to work the same way in bog-standard retrolab. 

@jtpio any thoughts on what the correct behavior should be here? I don't think it would block this, especially if it needs to be fixed upstream. I could imagine a few options (non-exclusive):
- disable the command altogether
- add a button (classic didn't use much right click, so I'm wagering folk won't expect stuff there)
- disable the command if multiple items are picked
- allow picking multiple items, open the tree, and then open all the tabs... weird, but seems like it would logically fit the idea

Anyhow, seems like RTD is having A Day... will take another look tomorrow.

bollwyvl:
Ok, one should be able to back-button from this link:

https://jupyterlite--457.org.readthedocs.build/en/457/_static/lab/index.html?path=javascript.ipynb&path=python.ipynb

leukipp:
Related to #450 I found that `router.navigate(...);` is used

https://github.com/jupyterlite/jupyterlite/blob/7b322c80125d51b393721ca0a28bf2bf432484e5/packages/application-extension/src/index.tsx#L479

[which does](https://github.com/jupyterlab/jupyterlab/blob/ea1529b6683adb692b1c9908f04d0a2cead44a9f/packages/application/src/router.ts#L80) a `history.pushState({}, '', url);`.

The history stack contains then the old url (with _path_ parameter) and the new url (without _path_  parameter).  Going back leads to the loop behavior.

Some options involving history manipulation:
- delete current history entry before navigate
- using [history.replaceState](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState)
- ...

jtpio:
Planning to take a look later when we start updating to the JupyterLab 4 and Notebook 7 packages: https://github.com/jupyterlite/jupyterlite/issues/826

Issue #459, opened 2022-01-25
Can't Interrupt Cell Execution in Python Kernel


brendan0powers:
## Description

It seems like interrupting cell execution isn't implemented. If a cell is busy, and I click the "stop" button, nothing happens.
If a cell is running a loop or awaiting some resource, I need to restart the kernel to get control back.

I'm not super familiar with how this works in IPython, but I wonder if the pyodide implementation makes this tricky?
I might have some time to help implement a fix, so if you have any suggestions on where to start, let me know!

## Reproduce

Run a cell with this code:
```python
while True:
    pass
```

Click the interrupt button in the toolbar, and nothing happens.

## Expected behavior

A keyboard interrupt exception should be thrown, and the kernel should become idle.

## Context

- JupyterLite version: 0.1.0-alpha.20
- Operating System and version: MacOs 10.15.7
- Browser and version: Chrome 97.0.4692.71

bollwyvl:
Having a go at it upstream on the [pyodide console](https://pyodide.org/en/stable/console.html) suggests we can't really do much about it...

bollwyvl:
Oof, on the contrary, we can, but it requires a WebWorker (which we use, hooray) and `SharedArrayBuffer` (which we haven't had to yet, hmmm):

https://pyodide.org/en/latest/usage/keyboard-interrupts.html

We probably need to hook this  down in the [kernel](https://github.com/jupyterlite/jupyterlite/blob/main/packages/pyolite-kernel/src/kernel.ts)  or [worker](https://github.com/jupyterlite/jupyterlite/blob/main/packages/pyolite-kernel/src/worker.ts) .

brendan0powers:
I was digging a little deeper on this, and it does seem like there is a way to implement this, although it has limitations.
https://pyodide.org/en/stable/usage/keyboard-interrupts.html
Basically, we'd use a SharedArrayBuffer to signal a keyboard interrupt from the main thread. It would look something like this.

1) Wait for the kernel to become ready
2) In the main thread, create an SharedArrayBuffer of size, 1. Pass is to the worker via postMessage()
    * Browser support for SharedArrayBuffer is new, and certain HTTP headers need to be set, so we'd need a way to fail here and notify the user somehow, perhaps in the log?
3) In the worker, call pyodide.setInterruptBuffer()
4) Add a new route for kernel interrupts to the server, route the message to the kernel in the same way as a kernel restart request.
[...]

PR #460, opened 2022-01-28
[WIP] Initial support for the interrupt button

brendan0powers:
## References

#459 

## Code changes

* Implemented interrupt route handler
* Added interrupt API to the kernel interface
* Added interrupt support to the pyolite kernel, using the method described here: https://pyodide.org/en/stable/usage/keyboard-interrupts.html

## User-facing changes

No user visible changes. The only effect should be the interrupt button working as epxected.

None

## Remaining Issues
* The current implementation fails to interrupt the kernel if it is awaiting a promise. More research is needed to figure out what's going on.
* The current implementation only works in recent Chrome, Firefox, or Safari browsers, and requires the correct HTTP headers be set. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements See #459 for a more detailed discussion.
* When interrupt is not available the interrupt button does nothing, and there is no feedback to the user. Should we display an error dialog, or gray the button out?
* What do we do about the Javascript kernel?


jtpio:
Thanks @brendan0powers for starting this!

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -10,6 +10,13 @@ import worker from './worker?raw';
 
 import { PIPLITE_WHEEL } from './_pypi';
 
+declare global {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  interface Window {
+    crossOriginIsolated: boolean | undefined;
```
bollwyvl, line 16: can we find a more canonical source for this type? or at least document where this magic variable comes from. At the very least, we'd want to move it out of the implementation, and probably in a dedicated `.d.ts` somewhere...

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -85,6 +93,31 @@ export class PyoliteKernel extends BaseKernel implements IKernel {
...
+    if (this._interruptBuffer !== null) {
+      this._interruptBuffer[0] = 2; //Send SIGINT
+    }
+  }
+
+  private setupInterruptBuffer(): void {
+    if (!window.crossOriginIsolated) {
+      console.warn(
+        'JupyterLite is not running in a cross origin isolated context. The interrupt button will not function. Details: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements'
+      );
```
bollwyvl, line 109: I'd probably use a tick-delimited string and break it up over multiple lines) so it can pass the linter... and it (along with many other constants) should probably be made internationalizable (but I wouldn't worry about it on this PR).

One approach might be to link to our some of our own, perhaps-not-yet-written documentation for this, as we can put it in a little more context.

Again, the average user is not going to open up the console to see this, so we may want to put this someplace else as well, e.g. the in-app kernel-focused log console... and potentially open it here.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -412,6 +419,10 @@ self.onmessage = async (event: MessageEvent): Promise<void> => {
       results = commClose(messageContent);
       break;
 
+    case 'set-interrupt-buffer':
```
bollwyvl, line 422: hm... i don't know if the intent of this API was to closely map to the kernel messaging spec. Ideally, there would be some way to model this on top of jupyter primitives instead of adding a new message type.

packages/pyolite-kernel/src/kernel.ts
brendan0powers: I'll move this to the declarations.d.ts for now. Types for crossOriginIsolated were added in typescript 4.4. https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029

I'm happy to find another way to do this as well, if you have any suggestions.

packages/pyolite-kernel/src/worker.ts
brendan0powers: This is pretty low level, and i'm not sure it's well suited for the comms API.  Especially since I'm not sure what would happen if you passed a SharedArrayBuffer to python and back. I can certainly test that though, if you think we should.

brendan0powers:
@bollwyvl Do you have any thoughts on how we could make the CORS errors more visible to the user? My preferred approach would be to gray out the stop button, remove it entirely, or show a dialog. Putting the error message in the kernel log would be better than the browser console, but I'm still not sure users would notice it.

Also, I'm not sure how much freedom we have to alter the Jupyter UI.

bollwyvl:
> CORS errors more visible to the user

yeah, there's a whole class of issues in these deployment scenarios that are... hard to describe in user-relatable terms. As you say, the ideal is removing things that don't work...

> gray out the stop button

...It _may_ be possible... some command control surfaces (e.g. command palette, menu items) support an `isEnabled` callback... but i don't know about buttons. Also, a big chunk of stuff around buttons is landing upstream in JupyterLab 3.3... and the _perhaps_ might still be time to sneak in some more features.

> how much freedom we have 

ha, we basically have infinite freedom: it's just a question of how long we want to maintain something that's divergent from upstream.

packages/pyolite-kernel/src/kernel.ts
bollwyvl: Looks like [upstream](https://github.com/jupyterlab/jupyterlab/pull/11594) is on:

```
 "typescript": "~4.5.2",
```

This would probably be separate PR, and we should have a look at what other goodies we can get any new from static analysis tooling.

packages/pyolite-kernel/src/worker.ts
bollwyvl: Right, there are some limits to what will work (and across what boundaries, see #463) and not everything is going to map perfectly, but as the first "custom" message on this pipe, it's worth taking a look. If we _could_ use these buffers directly, that might be _great_ for e.g. #195.

Perhaps we need to further differentiate the types of messages... it would be lovely for us to be able to: first distinguish between:
- "well-known" kernel message types from JupyterLab
  - ... and be able to use more typescript types from upstream, and maybe (ab)use totality checks in the `switch`
- "custom" jupyterlite ones
  - there might be a class of ones that _all_ lite kernels might need, e.g. _what is the root of the lite site_
  - per-kernel ones...

brendan0powers: I'm not sure where to go from here. Do you have any concrete suggestions on an approach?

brendan0powers:
@bollwyvl I'm waiting on https://github.com/pyodide/pyodide/pull/2142, but in the meantime, I'd like to push this forward. Do you have an idea of what you'd like to see in order for this PR to be merged?

packages/pyolite-kernel/src/declarations.d.ts
```diff
@@ -14,3 +14,12 @@ declare let _pipliteUrls: string[];
 declare let _disablePyPIFallback: boolean;
 declare let pyodide: any;
 declare let loadPyodide: any;
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+interface WindowOrWorkerGlobalScope {
```
bollwyvl, line 19: We synced the typescript version with the upstream, so you should be able to remove the extra typing.

packages/pyolite-kernel/src/kernel.ts
bollwyvl: Hm... it's still rather a deep dive to figure out how this might be effectively set. Some ideas:
- add something to the `docs` (maybe in `deploying.md`) about how one might set the appropriate headers
- update [serve.py](https://github.com/jupyterlite/jupyterlite/blob/main/py/jupyterlite/src/jupyterlite/addons/serve.py#L72) to set the appropriate headers so we can evaluate _something_ happening on e.g. binder
  - probably put it behind a flag, e.g. `--enable-shared-buffer`
  - if this _won't_ work over HTTP (even/especially on `localhost`), we can figure out some way to make it https with e.g. `trustme`

bollwyvl:
Perhaps this will help: https://github.com/pyodide/pyodide/pull/2332/files

packages/pyolite-kernel/src/kernel.ts
brendan0powers: I'll add some documentation to deployment.md, and look into adding a flag as you suggest. Logging to the log console seems like a good idea. Should the log be printed on kernel startup, or when the stop button is pressed?

bollwyvl:
If throwing an "info" is below the status bar threshold, maybe that at
start up and then a warning?

You may also want to see if anything is different with the pyodide 0.20
pr... I haven't been tracking what actually got merged before/after that
alpha.

brendan0powers:
> Perhaps this will help: https://github.com/pyodide/pyodide/pull/2332/files

It's great to see these patches making their way upstream, unfortunately, it doesn't change the way interrupts work.

brendan0powers:
After fiddling with a few different ways of displaying an error, I've settled on simply printing to the notebook when the interrupt button is clicked.
Interrupt Button Error

This has the advantage of notifying the user that something is wrong immediately upon clicking the button. The user may not be in a position to resolve the issue, but they will at least have a useful error message to forward to the maintainer of site. 

While implementing this, a few things came up:
* Clicking the interrupt button while the kernel is starting up will cause kernel startup to fail. This silently breaks the session, and the user cannot execute code. I've worked around this by adding a new `_kernelStarted` promise, as well as a new type of event to go with it. This adds a second new message type, which is less than ideal.
* I had to modify the message for setting the interrupt buffer so that it did not return a reply from the worker, otherwise it would cause the first cell to finish executing early in some cases.

Like you mentioned earlier, it seems like we might need a more elegant way to handle the "out of band" messages. I don't think using comms for this would be appropriate, since some of these actions need to happen before the kernel is fully ready.

Oh, also I tested setting the headers on localhost, and it works. I must have done something incorrectly with my previous tests. This means you don't need special Chrome flags or SSL to test, just the headers. I wonder if it would make sense to set the by default when using the jupyterlite CLI tool?

bollwyvl:
There is now a service worker available as of #655, which might enable more of this.

horw:
Hi, I haven't found any related information about interrupts, but it seems like interrupts aren't working now. Are there any docs available about interrupts?

krassowski:
@brendan0powers would you be interested in finishing up this PR given the recent inclusion of `SharedArrayBuffer` kernel communication added in https://github.com/jupyterlite/jupyterlite/pull/1640, or would you like someone else to pick it up?

Issue #461, opened 2022-01-30
Two feature requests: cache private python pkg, set pre-run python codes when open a notebook


qqdaiyu55:
**Feature Request 1**
Allow local file path in `piplite_urls`, so I can cache my private python package.

**Feature Request 2**
Run python codes defined in config everytime I opened a new notebook.
This can be really good for cutomization, like pre-install some packages, ignore uncessary warnings, ...

If you thinks these requests are resonable, I can help file a PR.

qqdaiyu55:
ok I saw we already support for local path
```
    def resolve_one_wheel(self, path_or_url):
        """download a single wheel, and copy to the cache"""
        local_path = None
        will_fetch = False
```

bollwyvl:
I've thrown out some thoughts on this in a [number](https://github.com/jupyterlite/jupyterlite/issues/438#issuecomment-1010060484) of [places](https://github.com/jupyterlite/jupyterlite/issues/458#issuecomment-1022532513): I don't think pre-running code is a feature that should go into the jupyterlite core, for the same reasons it's not a feature in other clients, and trying to write code in JSON is pretty not fun... but we're pretty much stuck with JSON as a config language.

For kernels, if an approach were to work, it would ideally match the upstream kernel's behavior, and might not work the same way for different kernels. A quick scan shows:

- for ipykernel
  -  [IPython startup files](https://ipython.readthedocs.io/en/stable/interactive/tutorial.html?#startup-files)...
  - ... will honor the `PYTHONSTARTUP` environment variable  
  - ... but doesn't have any command line options
- for JS (which should probably get to where it reuses/mirrors `ijavascript`)
  - doesn't have any config files...
  - ... or enviornment variables...
  - ... but accepts [`--startup-script`](https://github.com/n-riesco/ijavascript#ijskernel-ijavascript-kernel) path...

But even then I'm uncomfortable adding these mechanisms, as this happens _before_ the user gets access to the system. By keeping things in-script (or -notebook), it raises the chances that they will work correctly in multiple settings, and when they fail, it is easy for the user to see why.

qqdaiyu55:
@bollwyvl  Thanks for reply!
> it would ideally match the upstream kernel's behavior a la #97, and might not work the same way for different kernels. A quick scan shows:

Yes, I should make the option name more specific, like `prerunPythonCodes`.

> for ipykernel, IPython startup files...

[...]

PR #462, opened 2022-01-30
Add prerunCodes option to define prerun codes during notebook initialization

qqdaiyu55:
Related issue: #461


jtpio:
Thanks @qqdaiyu55 for looking into it.

I am closing this PR because the relevant code has now been moved to the Pyodide kernel repository: https://github.com/jupyterlite/pyodide-kernel

Also as discussed in https://github.com/jupyterlite/jupyterlite/issues/461, a way forward could be to:

- develop a third-party JupyterLab extension to allow running code when a notebook is opened
- support the IPython config (could be investigated in https://github.com/jupyterlite/pyodide-kernel)

This would have the benefit of not working with the Pyodide kernel only, but could also be used more broadly.

Thanks again!

PR #472, opened 2022-02-08
Share `build` directory for both `lab` and `retro` apps

### LINKED ISSUE #118: Deduplicate the lab and retro assets

jtpio:
## References

This should help fix #118 

And help make new 

## Code changes

Hoist the lab and retro `build` directories to the top-level, so they share the common assets.


jtpio:
Looks like a `~5MB` improvement on the wheel:

image

app/lab/index.html
```diff
@@ -7,7 +7,7 @@
     <link
       id="jupyter-lite-main"
       rel="preload"
-      href="./build/bundle.js"
+      href="../build/bundle-lab.js"
```
bollwyvl, line 10: Still getting my head aroudn stuff, but is there any way `bundle-<app>.js` could end up in a more atomically-removable place, e.g.:
- in `./lab/bundle.js`
- in `../build/lab/bundle.js`

app/lab/package.json
```diff
@@ -184,6 +184,7 @@
     "whatwg-fetch": "^3.0.0"
   },
   "jupyterlab": {
+    "name": "lab",
```
bollwyvl, line 187: Perhaps we should we put it in `jupyterlite`? Doesn't look like it appears [upstream](https://github.com/jupyterlab/jupyterlab/blob/master/builder/metadata_schema.json)....

app/retro/package.json
```diff
@@ -165,7 +165,7 @@
     "whatwg-fetch": "^3.0.0"
   },
   "jupyterlab": {
-    "name": "JupyterLite Retro",
+    "name": "retro",
```
bollwyvl, line 168: as above

app/lab/index.html
bollwyvl: i wonder if this is also good time to add cache-busting prefix to the name of this file... it will of course complicate a lot of things, as we won't _know_ the name of the file until it builds. perhaps we can get webpack to always emit something (e.g. stats) that we can track in `dodo` (but not ship).

app/lab/package.json
jtpio: It's not in the schema upstream, but can be found in `dev_mode/package.json`: https://github.com/jupyterlab/jupyterlab/blob/fc73254325aaf445678c411d53e736d3027f81a2/dev_mode/package.json#L209

app/lab/index.html
bollwyvl: or maybe we can somehow template `bundle.js?x=<sha256sum>` into the html files...

jtpio: > or maybe we can somehow template `bundle.js?x=<sha256sum>` into the html files...

Agree we should start looking into that at some point. But maybe this can be deferred to a follow-up PR?

bollwyvl: sure!

jtpio:
cc @bollwyvl CI all green, this should be good to go now

examples/jupyter_lite_config.json
```diff
@@ -10,7 +10,7 @@
       "https://github.com/conda-forge/releases/releases/download/noarch/ipyvue-1.5.0-pyhd3deb0d_0.tar.bz2/ipyvue-1.5.0-pyhd3deb0d_0.tar.bz2",
       "https://github.com/conda-forge/releases/releases/download/noarch/ipyvuetify-1.8.1-pyhd8ed1ab_0.tar.bz2/ipyvuetify-1.8.1-pyhd8ed1ab_0.tar.bz2",
       "https://github.com/conda-forge/releases/releases/download/noarch/jupyterlab-drawio-0.9.0-pyhd8ed1ab_0.tar.bz2/jupyterlab-drawio-0.9.0-pyhd8ed1ab_0.tar.bz2",
-      "https://github.com/conda-forge/releases/releases/download/noarch/jupyterlab-fasta-3.2.0-pyhd8ed1ab_0.tar.bz2/jupyterlab-fasta-3.2.0-pyhd8ed1ab_0.tar.bz2",
+      "https://github.com/conda-forge/releases/releases/download/noarch/jupyterlab-fasta-3.1.1-pyhd8ed1ab_0.tar.bz2/jupyterlab-fasta-3.1.1-pyhd8ed1ab_0.tar.bz2",
```
jtpio, line 13: Ah looks like this change needs to be reverted.

docs/configuring.md
```diff
@@ -327,14 +327,14 @@ The Theme Manager expects to be able to load theme CSS/font assets from
...
-cp -r @*/themes/* ../build/themes/
+cd $YOUR_JUPYTERLITE/extensions
+mkdir -p ../../build/themes
+cp -r @*/*/themes/* ../../build/themes/
+cp -r @*/themes/* ../../build/themes/
 # To also ensure these are available for JupyterLite Retro:
-mkdir -p ../../retro/build/themes
-cp -r @*/*/themes/* ../../retro/build/themes/
-cp -r @*/themes/* ../../retro/build/themes/
+mkdir -p ../../build/themes
```
bollwyvl, line 335: is the second stanza still needed?

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -44,7 +44,7 @@ export class PyoliteKernel extends BaseKernel implements IKernel {
 
     const { origin } = window.location;
 
-    const pypi = URLExt.join(origin, PageConfig.getOption('appUrl'), 'build/pypi');
+    const pypi = URLExt.join(origin, PageConfig.getOption('baseUrl'), 'build/pypi');
```
bollwyvl, line 47: could be [`getBaseUrl`](https://github.com/jupyterlab/jupyterlab/blob/f786d539ab5210464d7bc02cda88999949335b6e/packages/coreutils/src/pageconfig.ts#L106)...

py/jupyterlite/src/jupyterlite/constants.py
```diff
@@ -87,7 +87,7 @@
 #: the Jupyter API route for Contents API
 API_CONTENTS = "api/contents"
 API_TRANSLATIONS = "api/translations"
-LAB_EXTENSIONS = "lab/extensions"
+LAB_EXTENSIONS = "extensions"
```
bollwyvl, line 90: awesome

packages/pyolite-kernel/src/kernel.ts
jtpio: ah right good catch!

docs/configuring.md
jtpio: Should be fixed in https://github.com/jupyterlite/jupyterlite/pull/472/commits/8b6694239c3d254a6a7891a21ca97628b9ad7f98

bollwyvl:
So... there are a _lot_ of examples where we're getting close, but not _exactly_ the same modules. For example: 
Screenshot from 2022-02-09 09-34-38

under the hood, these are _almost_ identical, except one depends on one extra file. This is _particular_ aggregious for `4523`, which is blueprint: it's 822k, and its sourcemap is 2mb, so shipping it twice really does kinda sting.

While we could probably special case some of these, I wonder if there are some more webpack config tricks we can apply (e.g. runtime chunks, etc) that would further stabilize the generated ids... and potentially improve build performance...

bollwyvl:
Also two copies of `moment` (and all of its locales)...

jtpio:
> This is particular aggregious for 4523, which is blueprint: it's 822k, and its sourcemap is 2mb, so shipping it twice really does kinda sting.

Thankfully Blueprint is going away in lab 4.0. And it looks like the source maps are indeed part of the prod assets, so maybe we could also strip them.

bollwyvl:
> Blueprint is going away in lab 4.0.

right, but there are _many_ examples like this... that's just the worst one.

> source maps are indeed part of the prod assets, so maybe we could also strip them.

i'm definitely _not arguing_ for dropping source maps: it makes the overall system all but impossible to debug.... i mean, i guess we could keep them in the `sdist` and strip them from the `whl`, but that's a lot of bother.

jtpio:
> Hm... i don't see what the issue is with the UI tests.

I think this is because of the `PageConfig.getBaseUrl()` recent change which is fixed in https://github.com/jupyterlite/jupyterlite/pull/474

When I was investigating the issue by running the UI tests locally we could see the `pypi` related error in the dev tools (see screenshot in https://github.com/jupyterlite/jupyterlite/pull/472#issue-1127547652)

PR #474, opened 2022-02-10
Build all apps in single webpack

bollwyvl:
## References

- progress towards #118
- built on #472

## Code changes

- [x] unifies `lab` and `retro` builds under a single run of webpack
- [x] total size of `.whl` is ~9mb
  - ... if we _were_ to strip the `.js.map` files size would be _~3.6mb_
     - ... but much harder to debug
  - ... perhaps make an option for deployment, e.g. `--no-source-maps`
    - ... could also work for labextensions
- [x] centralize listing of apps in e.g. `app/package.json`

## User-facing changes

- users should see better performance when switching between multiple apps
- deployers should see smaller package download and deployment sizes
- contributors should save a minute or two when rebuilding

- specifying a build of a single app (e.g. `app/build/lab`) will have a smaller overall impact, as it is now far more difficult to remove underlying dependencies (just the ~200kb applications)


bollwyvl:
hooray <9mb!

probably won't do any more this evening... feel free to steal/push!

jtpio:
Thanks!

Yeah thinking about https://github.com/jupyterlite/jupyterlite/pull/472#issuecomment-1033914621, instead of attempting to tweak the webpack config it started to look like it would be simpler to let webpack handle everything with a single build, instead of trying to manually merge the assets. 

> ... if we were to strip the .js.map files size would be ~3.6mb 

This would indeed be useful to have (can be follow-up), either opt-in or opt-out (to be decided).

jtpio:
>  centralize listing of apps in e.g. app/package.json

It's fine to leave them hardcoded for now.

jtpio (approved):
Thank you!

bollwyvl:
> It's fine to leave them hardcoded for now.

ha, just pushed with a fix on this, so that at least `doit` and `webpack` are on the same page. 

the python-side constant is still there: haven't touched that side of the house yet. If we normalize the data, though, we will be able to read it right out of the archive.

To get to the cache-busting (again as a follow-on), i think we can go a bit further, so that the `liteApps` would look like:

```yaml
"jupyterlite": {
  "liteApps": {
     "lab": {"pages": ["index.html"]},
     "retro": {"pages": ["edit/index.html", "notebooks/index.html", "consoles/index.html", "tree/index.html]}
  }
}
```

This would give us enough information to update all of the `bundle-js?x=[hash]` templates... and could be wrapped up with the changes to allow for stripping sourcemaps, as its related to "optimizing builds". I'll start that issue.

app/package.json
```diff
@@ -9,6 +9,37 @@
...
     "url": "https://github.com/jupyterlite/jupyterlite"
   },
+  "scripts": {
+    "clean": "rimraf build lab/build retro/build",
+    "build": "webpack",
+    "build:prod": "yarn clean && yarn build --mode=production",
+    "watch": "webpack --config webpack.config.watch.js"
+  },
+  "jupyterlite": {
+    "liteApps": [
```
jtpio, line 19: No strong opinion, but maybe just `apps` would be enough since this is already inside `jupyterlite`?

bollwyvl: [695ccda](https://github.com/jupyterlite/jupyterlite/pull/474/commits/695ccda5b5f73757619b35503f7f14b77af9a704)

Issue #386, opened 2021-10-12
Move `pyolite` to a different repository


jtpio:
### Problem

At the moment the Python kernel is part of the monorepo: 

- https://github.com/jupyterlite/jupyterlite/tree/main/packages/pyolite-kernel
- https://github.com/jupyterlite/jupyterlite/tree/main/packages/pyolite-kernel-extension

This has been convenient so far as it allowed for faster iterations until we implemented most of the Jupyter protocol (including comms for widgets).

However there are now a couple of reasons to start thinking moving it to a different repo:

- it is now growing in size and complexity
- it also depends on a couple of shims (`ipykernel`, `widgetsnbextension`...)
- all of this adds some weight to the `jupyterlite` dev workflow
- there might be alternative implementations for a Python kernel in the future (for example based on `xeus-python`)
- other JupyterLite distributions like the p5 notebook don't need it and have to disable the extension: https://github.com/jtpio/p5-notebook

With support for [federated server extensions](https://jupyterlite.readthedocs.io/en/latest/extensions.html), we can imagine having that kernel living outside of the main repo, for example in https://github.com/jupyterlite/pyolite-kernel or https://github.com/jupyterlite/ipylite-kernel.

### Proposed Solution

We move the kernel to https://github.com/jupyterlite/pyolite-kernel and make it a federated server extension.

### Additional context

This was also mentioned by @bollwyvl in https://github.com/jupyterlite/jupyterlite/issues/383#issuecomment-940112826

This might also simplify things like https://github.com/jupyterlite/jupyterlite/pull/310, but raises new issues:

#### config junk
- move the schema description of `pyodideUrls` and `micropipUrls` to a file pointed to by `package.json#/jupyterlite/schemaDir`
  - does this need to be disjoint from `jupyterlab/schemaDir`
- in `jupyterlite.schema.v0.json` (or maybe this is finally a `v1`?), add a new field, `extensionConfiguration`, which allows per-extension configuration (akin to `jupyter_(notebook|server)_config.json`

#### lite wheels
- serve from `lab/extensions/@jupyterlite/pyolite-extension/static/wheels` via webpack `assets`
  - need some CLI, e.g. `jupyter ipylite index` to generate the business for complex packages

#### CLI build stuff
  - move `entry_point` (hooray! _one_ thing is easy!)
  - add a way to extend `flags` and `aliases`?

jtpio:
[...]

PR #492, opened 2022-02-11
Have pyolite kernel `execute_request` does not honor `store_history` option

### LINKED ISSUE #478: Pyolite kernel `execute_request` does not honor `store_history` option
## Description

Running an `execute_request` against the pyolite kernel with `store_history: false` increments the execution counter. I believe that Jupyter kernels should not increase the execution counter when `store_history: false` is given, e.g.,

https://jupyter-client.readthedocs.io/en/latest/messaging.html#execute

I encountered this issue because I am trying to add silent pre-run code to [`replite`](https://github.com/jtpio/replite) in https://github.com/jtpio/replite/pull/24. Currently it's not possible to do that without having the first cell be something like `[2]` instead of `[1]`.

## Reproduce

```
kernel.requestExecute({ code: line,silent: true, store_history: false});
```
with the pyolite kernel causes the execution counter to be increased. This happens [here in the code](https://github.com/jupyterlite/jupyterlite/blob/ca7bd69f33164c02eac94ec5e3f850f2c2063c2e/packages/kernel/src/kernel.ts#L536).

## Expected behavior

For the execution counter not to be increased. I have a quick fix here:

https://github.com/jupyterlite/jupyterlite/compare/main...jobovy:honor-store_history

which I'm happy to PR. However, I'm not sure whether `store_history` shouldn't imply other additional behavior.

jobovy:
Fixes #478.

### DIFF
--- a/packages/kernel/src/kernel.ts
+++ b/packages/kernel/src/kernel.ts
@@ -533,7 +533,9 @@ export abstract class BaseKernel implements IKernel {
   private async _execute(msg: KernelMessage.IMessage): Promise<void> {
     const executeMsg = msg as KernelMessage.IExecuteRequestMsg;
     const content = executeMsg.content;
-    this._executionCount++;
+    if (content['store_history']) {
+      this._executionCount++;
+    }
 
     // TODO: handle differently
     this._parentHeader = executeMsg.header;
--- a/packages/kernel/src/kernel.ts
+++ b/packages/kernel/src/kernel.ts
@@ -542,7 +542,9 @@ export abstract class BaseKernel implements IKernel {
 
     this._executeInput(executeMsg);
 
-    this._history.push([0, 0, content.code]);
+    if (content['store_history']) {
+      this._history.push([0, 0, content.code]);
+    }
 
     const reply = await this.executeRequest(executeMsg.content);
     const message = KernelMessage.createMessage<KernelMessage.IExecuteReplyMsg>({


packages/kernel/src/kernel.ts
```diff
@@ -533,14 +533,18 @@ export abstract class BaseKernel implements IKernel {
   private async _execute(msg: KernelMessage.IMessage): Promise<void> {
     const executeMsg = msg as KernelMessage.IExecuteRequestMsg;
     const content = executeMsg.content;
-    this._executionCount++;
+    if (content['store_history']) {
```
jtpio, line 536: Since `content` is well typed, something like the following should also work:

```suggestion
    if (content.store_history) {
```

packages/kernel/src/kernel.ts
```diff
@@ -533,14 +533,18 @@ export abstract class BaseKernel implements IKernel {
...
+      this._executionCount++;
+    }
 
     // TODO: handle differently
     this._parentHeader = executeMsg.header;
 
     this._executeInput(executeMsg);
 
-    this._history.push([0, 0, content.code]);
+    if (content['store_history']) {
```
jtpio, line 545: Similar comment as above:

```suggestion
    if (content.store_history) {
```

packages/kernel/src/kernel.ts
```diff
@@ -533,14 +533,18 @@ export abstract class BaseKernel implements IKernel {
   private async _execute(msg: KernelMessage.IMessage): Promise<void> {
     const executeMsg = msg as KernelMessage.IExecuteRequestMsg;
     const content = executeMsg.content;
-    this._executionCount++;
+    if (content.store_history) {
```
bollwyvl, line 536: [Upstream](https://github.com/jupyter/jupyter_client/blob/main/jupyter_client/client.py#L559-L565), it looks like we may want to:

- check if `content.store_history !== false` (as it should be true by default)
- check if `content.silent` is true, which would override it

jobovy: I had assumed that the documented behavior was enforced somewhere, but testing it it does indeed seem to not be the case (i.e., running with just `silent: true` does not lead to `store_history: false`). Would be simple to include here, but are you suggesting that this really needs to be fixed in another component (like the `jupyter_client`)? I don't know enough about how all of the pieces of the jupyter framework work together to know where to actually fix this (took me long enough to figure out that the execution counter was increased here!).

bollwyvl: > documented behavior 

all depends on which one you read, really, and some of it is _only_ in the reference implementation. 

>  fixed in another component 

for purposes of jupyterlite, the only component we get to reuse is [ipython](https://github.com/ipython/ipython/search?l=Python&q=execution_count): we have to do everything else ourselves in this repo, as `ipykernel` and `jupyter_client` do all of the zmq stuff (which doesn't work in the browser). so really, this behavior is probably _already_ covered in ipython, and we just need to wire it up... except it's on the "wrong" side of the kernel wall. Putting it in our base `kernel` implementation is fine, though, if it means other kernel authors don't have to think about it for it to work "right" for different use cases.

>  don't know enough about how all of the pieces of the jupyter framework

No worries, there's rather a lot to keep straight when we're shipping client/server/kernel in the same repo! in the Bad Old Days, this is what ipython was like... we're just getting to put all the shiny pieces back together... even here, we want to get to a multi-repo setup (e.g. #386), but for now this is what we have.

ideally, we'd figure out some way to wire up [jupyter_kernel_test](https://github.com/jupyter/jupyter_kernel_test) and write some conformance tests against our kernels: for now, just pointing to how our stuff quacks like the upstream is sufficient and helpful: liberal links to (versioned) places in the reference implementations would be great... it all gets minimized anyway.

jtpio: > we just need to wire it up... except it's on the "wrong" side of the kernel wall. Putting it in our base kernel implementation is fine, though, if it means other kernel authors don't have to think about it for it to work "right" for different use cases.

Yeah I also thought we were tracking the execution count on Python side for the pyolite kernel.

Probably this is fine for now to add the check in the base kernel class for now since this is where the tracking is done. We can open an issue to track this.

For reference JupyterLab default to `store_history: true` here: https://github.com/jupyterlab/jupyterlab/blob/c5df985383964b027e475e667d69a7e8facf42bb/packages/services/src/kernel/default.ts#L710

jtpio (approved):
Thanks!

Issue #148, opened 2021-06-16
Provide a single executable cell


benabel:
Hi dear jupyterians, 

I'm a french teacher and I really like this project because it's quite difficult for students to install python on their own machines.

First of all the full notebook frontend would be very useful for practicing on projects.

But sometimes we just need a single cell to show a quick example or ask a question. Is there's any component in the jupyterstack that could be ported as something as simple as a web component as it is done on `wc-component`.

https://github.com/vanillawc/wc-code

Thanks for sharing this as an open source project.

jtpio:
Thanks @benabel.

JupyterLab Code Consoles should already work in JupyterLite, and might be a good fit for this use case?

https://user-images.githubusercontent.com/591645/122247036-5b451580-cec7-11eb-8a3a-4ff25a537fcd.mp4

Ideally this would also be supported in the Retro interface (still to be implemented) to have a single console / cell on the page:  https://github.com/jupyterlab/retrolab/issues/49

benabel:
Thanks for your answer,, console seems to do the job if it is alone one the page. 

Is there any way to provide initial content in the console, some lines of code to be executed  and or modified by the user?

jtpio:
> Is there any way to provide initial content in the console, some lines of code to be executed and or modified by the user?

The code console is more of a temporary and stateless component. It sounds like a notebook would be quite appropriate in that case?

psychemedia:
Just as an example of a related user experience, [SageMathCell](https://sagecell.sagemath.org/) allows you to run an arbitrary bit of code against a CoCalc kernel and display the result.

![image](https://user-images.githubusercontent.com/82988/122429023-5a28ec80-cf8a-11eb-9046-32fb3d26fdf0.png)

bollwyvl:
Right. It's a good goal. Once we have #118 squared away, we'll be able to do all kinds of custom apps (without substantially over-inflating the final archive (or having to do multiple releases), with a file tree more like:

```
app/
  index.html    # will redirect to... something
  static/       # where all the build stuff will end up
  lab/         
    index.html  # won't be required
    extensions/
  retro/
    index.html
  editor/
    index.html
  cell/
    index.html
  term/
    index.html
  book/
    index.html  # will need lots of work!
  # .../        # basically anything from https://github.com/jupyterlab/jupyterlab/tree/master/examples
```

[...]

PR #498, opened 2022-02-13
Add a new `repl` app

jtpio:
## References

#148 

https://github.com/jtpio/replite

Fixes https://github.com/jtpio/replite/issues/20

This PR adds a new `repl` app:

repl

## Code changes

- [x] Add a new `console` app
- [x] Add a new `@jupyterlite/console-extension` package
- [x] Bring plugins from https://github.com/jtpio/replite
- [x] Remove unneeded plugins in the console index script
- [x] Document the new app
- [x] Document how to embed
- [x] Check new size for the assets (start measuring in a follow-up?)
- [x] Add simple UI test to cover the new app
- [x] Decide on the name: `console`, `cell`, `repl`, `replite`?

## User-facing changes

Users are now able to use a `console` app.

For example on the RTD preview: https://jupyterlite--498.org.readthedocs.build/en/498/_static/repl/index.html?toolbar=1&kernel=javascript

None (this is additive)

#### Provide `IURLParamManager`

consider `provide`ing this as a  `IURLParamManager` which can have additional params added that modify how things work at different lifecycle hooks:

```
urlParamManager.registerParam('code', {
  onAppStarted: async () => console.log('whatever')
  onWidgetAdded: (widget) => await ....
});
```


app/lab/package.json
```diff
@@ -78,18 +78,18 @@
...
-    "@lumino/coreutils": "~1.8.2",
-    "@lumino/disposable": "~1.7.2",
-    "@lumino/domutils": "~1.5.2",
-    "@lumino/dragdrop": "~1.10.2",
-    "@lumino/messaging": "~1.7.2",
-    "@lumino/properties": "~1.5.2",
-    "@lumino/signaling": "~1.7.2",
-    "@lumino/virtualdom": "~1.11.2",
-    "@lumino/widgets": "~1.26.2",
+    "@lumino/algorithm": "~1.9.1",
```
bollwyvl, line 81: :+1:

app/package.json
```diff
@@ -10,15 +10,16 @@
     "url": "https://github.com/jupyterlite/jupyterlite"
   },
   "scripts": {
-    "clean": "rimraf build lab/build retro/build",
+    "clean": "rimraf build console/build lab/build retro/build",
```
bollwyvl, line 13: maybe we need a `lerna run clean`

packages/application/src/singleWidgetApp.ts
```diff
@@ -0,0 +1,155 @@
...
+import { IRenderMime } from '@jupyterlab/rendermime-interfaces';
+
+import { ISingleWidgetShell, SingleWidgetShell } from './singleWidgetShell';
+
+/**
+ * App is the main application class. It is instantiated once and shared.
+ */
+export class SingleWidgetApp extends JupyterFrontEnd<ISingleWidgetShell> {
+  /**
+   * Construct a new NotebookApp object.
```
bollwyvl, line 21: update for new class name

packages/application/src/singleWidgetApp.ts
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @param options The instantiation options for an application.
+   */
+  constructor(options: NotebookApp.IOptions = { shell: new SingleWidgetShell() }) {
```
bollwyvl, line 25: `NotebookApp.IOptions` accurate?

packages/application/src/singleWidgetApp.ts
```diff
@@ -0,0 +1,155 @@
...
+    mods.forEach(mod => {
+      this.registerPluginModule(mod);
+    });
+  }
+}
+
+/**
+ * A namespace for App statics.
+ */
+export namespace NotebookApp {
```
bollwyvl, line 127: `SingleWidgetApp`

packages/application/src/singleWidgetShell.ts
```diff
@@ -0,0 +1,120 @@
...
+
+import { ISignal, Signal } from '@lumino/signaling';
+
+import { Panel, Widget, PanelLayout } from '@lumino/widgets';
+
+/**
+ * The single widget application shell token.
+ */
+export const ISingleWidgetShell = new Token<ISingleWidgetShell>(
+  '@jupyterlite/:INotebookShell'
```
bollwyvl, line 20: `@jupyterlite/application:ISingleWidgetShell`

packages/console-extension/src/index.ts
```diff
@@ -0,0 +1,236 @@
...
+    tracker.widgetAdded.connect((_, console) => {
+      const { toolbar } = console;
+
+      console.toolbar.addItem('run', runButton);
+      console.toolbar.addItem('restart', restartButton);
+      console.toolbar.addItem('clear', clearButton);
+
+      toolbar.addItem('spacer', Toolbar.createSpacerItem());
+
+      const wrapper = new Panel();
```
bollwyvl, line 98: i :heart: branding, but would consider:
- toning it down to just an icon + tooltip
- making it a separate plugin which can be independently disabled

packages/console-extension/src/index.ts
```diff
@@ -0,0 +1,236 @@
...
+    tracker: IConsoleTracker | null,
+    themeManager: IThemeManager | null
+  ) => {
+    if (!tracker) {
+      return;
+    }
+    const { commands } = app;
+
+    const search = window.location.search;
+    const urlParams = new URLSearchParams(search);
```
bollwyvl, line 142: As a follow-on, consider `provide`ing this as a  `IURLParamManager` which can have additional params added that modify how things work at different lifecycle hooks:

```
urlParamManager.registerParam('code', {
  onAppStarted: async () => console.log('whatever')
  onWidgetAdded: (widget) => await ....
});
```

packages/console-extension/src/index.ts
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * The default JupyterLab application status provider.
+ */
+const status: JupyterFrontEndPlugin<ILabStatus> = {
+  id: '@jupyterlite/console-extension:status',
+  autoStart: true,
+  provides: ILabStatus,
+  activate: (app: JupyterFrontEnd) => {
+    if (!(app instanceof SingleWidgetApp)) {
+      throw new Error(`${status.id} must be activated in SingleWidgetApp.`);
```
bollwyvl, line 182: maybe hoist this from both places to a function

yarn.lock
```diff
@@ -1691,7 +1691,7 @@
     y-websocket "^1.3.15"
     yjs "^13.5.17"
 
-"@jupyterlab/docregistry@^3.2.9":
+"@jupyterlab/docregistry@^3.2.9", "@jupyterlab/docregistry@~3.2.9":
```
bollwyvl, line 1694: should probably align these

bollwyvl:
> Decide on the name: console, cell, repl, replite?

I guess I like `repl` as the _app_ name, to distinguish it from the console component of retro and lab.

packages/application/src/singleWidgetApp.ts
jtpio: Good catch, fixed in https://github.com/jupyterlite/jupyterlite/pull/498/commits/b98e9a8283e467d93891a225954323dd64fca980

app/package.json
jtpio: Or `rimraf build **/build`?

Otherwise that would probably mean adding back the `clean` scripts in each of the apps.

jtpio: Done in https://github.com/jupyterlite/jupyterlite/pull/498/commits/880847fd9b750319457b25495148aa00bb989928

packages/console-extension/src/index.ts
jtpio: Sounds good, adding to the follow-up list.

Maybe this can even be part of `@jupyterlite/application-extension` so it's always available in lite apps.

jtpio:
> I guess I like `repl` as the _app_ name, to distinguish it from the console component of retro and lab.

Right `repl` sounds good. And this would allow us (or a third-party extension) to switch the widget to something else than the JupyterLab Code Console if we want, for example to something like in https://github.com/jupyterlite/jupyterlite/issues/148#issuecomment-863434415:

mockup

packages/console-extension/src/index.ts
jtpio: > toning it down to just an icon + tooltip

done in https://github.com/jupyterlite/jupyterlite/pull/498/commits/01a5564622413e4fb2f3ba1ffc26657dc3ab9a2e

bollwyvl: looks good. i feel like perhaps if `?kernel` is not specified, then the kernel picker should probably still be shown, as the once you pick it with the dialog, there's no way to change it. Perhaps collapsing the kernel text down to just the kernel status icon might be a good way.

jtpio: > if ?kernel is not specified, then the kernel picker should probably still be shown

Yes that's the default behavior. 

Then we could indeed as the toolbar widget to select the kernel. Otherwise it's still possible to open the command palette and choose "Change Kernel" (although harder to do on mobile)

bollwyvl: > the default behavior.

Right, what i mean is the one that's in the menu bar in classic that says, e.g. `Python 3 ( )`. I think we could collapse it down to just being the kernel icon and the name in a mouseover, next to the status icon...

> open the command palette 

yeah, i'd basically assume nobody is going to find _any_ keyboard shortcut, unless it is labeled in the tooltip of a button, or otherwise discoverable. indeed, having a _show keyboard shortcuts_ button it probably a _really_ good idea as a default...

really, the whole embeddable toolbar needs some of settings-data-driven love, as i could see it enabling a host of nice things, but maybe not all the time for every use case.

as an aside, not to fix right now: a lot of modals/popups are just going to look _bad_ when embedded due to some hard-coded design decisions based on having the whole screen...

yarn.lock
jtpio: I think we have a couple of `@jupyterlab/` dependencies like this.

Maybe we can do one more pass on the `yarn.lock` as a follow-up.

jtpio:
This should be good to go for now, at least for a first version that we can iterate on.

docs/applications/repl.md
```diff
@@ -0,0 +1,76 @@
...
+
+![image](https://user-images.githubusercontent.com/591645/153935929-23a5d380-363e-490b-aabd-f0a780140588.png)
+
+## Embedding the REPL on another website
+
+You can embed the REPL on any website with the following code snippet:
+
+```html
+<iframe
+  src="https://jupyterlite.rtfd.io/en/latest/_static/lab/index.html"
```
bollwyvl, line 13: all point at `lab` instead of `repl`...

bollwyvl: also somewhat uneasy about repping the docs URL directly as a thing to be used hot... perhaps an activated version of the docs, once available would be less concerning...

bollwyvl:
Seems to be working pretty well: provided it actually prevents them from being loaded, we can probably tune the number of loaded `federated_extensions` _way_ down in `examples/repl/jupyter-lite.json`: no point _yet_ in any widgets, drawio, kernelspy, vue, etc. though maybe https://github.com/jupyter-widgets/ipywidgets/pull/3004 will land :rofl:

bollwyvl (approved):

docs/applications/repl.md
jtpio: Good point, we can mention folks should use a custom deployment if possible, and maybe we can point to `jupyterlite.rtfd.io/repl` or `jupyterlite.github.io/demo/repl` (a bit more stable) as an example.

jtpio: Updated in https://github.com/jupyterlite/jupyterlite/pull/498/commits/9aaeca6bdb675465e9a9b94c40795f52bc919292

jtpio:
Some extensions should now be disabled for the `repl` app.

Issue #390, opened 2021-10-13
Adding new server routes in a serverlite extension


jtpio:
### Problem

For now the server implementation doesn't allow adding new routes: https://github.com/jupyterlite/jupyterlite/tree/main/packages/server/src

### Proposed Solution

There should be an extension point so serverlite plugins can add new routes.

One way would be to just expose the router to the public interface:

```typescript
app.server.router.get(
  '/voila/render/(.+)',
  async (req: Router.IRequest, path: string) => {
    // handle the request
  }
);
```

That would be for requests going through `ServerConnection.ISettings`. In the case of voila that would be a bit more complicated and probably require a new app / endpoint that would also leave on the server.

### Additional context

For reference the default routes are defined here:

https://github.com/jupyterlite/jupyterlite/blob/6fdcbbe38616fced8ca59ee5b0d1ce6530765c4f/packages/server/src/server.ts#L52-L218

bollwyvl:
Ha, thought it already did that! We'll want new routes for sure!

As might be guessed, i'd see value in (optionally?) allowing registration
of a (couple) routes with an openapi spec.

I don't understand the voila issue, but that's probably okay.

bollwyvl:
Ha, i thought we had already done this!

jtpio:
Still one "major" change to do before 0.1.0 beta since this is probably going to break the API.

Issue #383, opened 2021-10-11
0.1.0 Release Plan


jtpio:
This is an issue for the `0.1.0` release planning.

We still have the MVP board for the things we wanted to tackle: https://github.com/jupyterlite/jupyterlite/projects/1 (although this was mostly before we have the `jupyterlite` CLI)

Having this in an issue makes it more visible and it can also be pinned.

## Dates

There will be a talk at PyData Global 2021 about JupyterLite: https://pydata.org/global2021/schedule/presentation/113/jupyterlite-jupyter-webassembly-python/

Maybe we can aim for a `0.1.0b0` just before the event, somewhere around Wednesday 27 October. We could then stay in beta for a little while, or just proceed with rc and then final one and two weeks after.

## Issues

- https://github.com/jupyterlite/jupyterlite/pull/371
- https://github.com/jupyterlite/jupyterlite/pull/310
- https://github.com/jupyterlite/jupyterlite/issues/382
- A look at the issues [tagged with bug](https://github.com/jupyterlite/jupyterlite/issues?q=is%3Aissue+is%3Aopen+label%3Abug)

## Nice to have

- Not strictly related to the main `jupyterlite` repo, but it would be great to have Xeus kernels packaged and release:
  - https://github.com/jupyterlite/xeus-kernel for the common package other xeus kernels can depend on
  - https://github.com/DerThorsten/jupyterlite_xeus_lua (to be moved to the `jupyterlite` org later) for a xeus-based Lua kernel
  - https://github.com/DerThorsten/xeus-wren following the same approach for a different language

jtpio:
The first beta is out: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.0b0

We can wait for https://github.com/jupyterlite/jupyterlite/issues/511 and make a new beta release with the JupyterLab 3.3 packages (and other beta releases before if needed).

And we can postpone https://github.com/jupyterlite/jupyterlite/issues/386 to `0.2.0`, so there is also time to iterate on the web worker interface (https://github.com/jupyterlite/jupyterlite/issues/463).

jtpio:
@bollwyvl I was thinking about pushing `0.1.0` soon so we have a "proper" release out. Apart from https://github.com/jupyterlite/jupyterlite/pull/525 it's mostly bug fixes and docs remaining, which could go to `0.1.x` as well.

[...]

Issue #393, opened 2021-10-15
Restructure the documentation


jtpio:
### Problem

There is now quite a bit of documentation mainly split into two categories:

- User Guide
- Developer Guide

![image](https://user-images.githubusercontent.com/591645/137454643-892154c3-d596-4653-abca-a55eb29ca8d7.png)

The developer guide is however longer than the user guide:

![image](https://user-images.githubusercontent.com/591645/137454672-301c4933-6f81-4ee7-bb39-de1a1cd40adf.png)

### Suggested Improvement

TLJH has some "How-to Guides" 

![image](https://user-images.githubusercontent.com/591645/137455409-3d4b5522-0e24-47a3-82ea-84deb4862289.png)

Maybe we could move some of the sections in the developer guide to the top-level, so they are more visible? Since we often get questions from folks who would like to deploy their own on GitHub Pages, for example: https://github.com/jupyterlite/jupyterlite/issues/318

With top-level categories:

- User Guide
  - For users of https://jupyterlite.rtfd.io/en/latest/try/lab or other already deployed lite web site
- Deploying
- Configuring
- Advanced:
  - CLI
  - API
  - Architecture
- Contributing
- Changelog

Also, the user guide could be consolidated by https://github.com/jupyterlite/jupyterlite/issues/31.

jtpio:
With #310 now in (and soon #419), it would be great to have hands-on and practical guides for various use cases.

For example following the 4 types of documentation: https://documentation.divio.com/

jtpio:
Some ideas for how-to guides (this list might be updated later):

- How to override keyboard shortcuts, for example to use <kbd>Enter</kbd> to execute code in the console

PR #506, opened 2022-02-15
Expose the Router to be able to add new routes from third-party plugins

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/390

Closing in on some potential `0.1.0` API breaking changes so we can soon start releasing `0.1.0` final (#383).

## Code changes

Expose the main `Router` instance so server extensions can add new routes.

## User-facing changes

None

- [x] Remove the `JupyterServer` and `LiteServiceManager` classes
- [x] Routes are added by the server plugins directly
- [x] Remove `@jupyterlite/server-extension:server` plugin


bollwyvl:
This is a good first step: breaking up the `_addRoutes` into separate plugins (perhaps co-located with their managers) that use the API would a) make it much clearer how one might use these APIs effectively and b) make it far easier to reason about what is/isn't loaded in case someone wants to overload it.

jtpio:
More code cleaning and route splitting in https://github.com/jupyterlite/jupyterlite/pull/506/commits/e518e4b4322ea0a5ccf03a676ecddb46066f40cf.

This should be better now, and allow for third-party server extensions to easily add new routes, or replace core plugins just like for JupyterLab (via `disabledExtensions` or providing custom ones).

bollwyvl:
This already reads much better than the monolithic `_addRoutes`. 

The architecture this PR currently presents is a bit different from the upstream `jupyter_server`, in that there, one can replace just, e.g. `ContentsManager`, and implement its API, but _not_ have to redefine the routes, which are really the contract with the frontend. Further, the routes are kinda untyped (basically `Request<string>`) which might make it harder still to practically consider replacing them.

So anyhow: the other way (just for consideration) is a pattern where all of the routes stay in the files where they have been moved, but are separated into `thingRoutesPlugin`: e.g. the `contentsPlugin` provides `IContents`, to be consumed by `contentsRoutesPlugin`, which actually adds the routes.

jtpio:
Good point. Especially since the frontends (at least both lab and retro) expect these routes to be functional.

It makes sense to be able to just swap the implementation of the different services, while still keeping the default routes. I'll push a change to split that into the routes plugins.

jtpio:
> It makes sense to be able to just swap the implementation of the different services, while still keeping the default routes. I'll push a change to split that into the routes plugins.

Done in https://github.com/jupyterlite/jupyterlite/pull/506/commits/b2918d34d3f494bf0916d195f69f10a81eb9c5e4

jtpio:
A follow-up would be to add a sample plugin the serverlite cookiecutter to show how to add new routes: https://github.com/jupyterlite/serverlite-cookiecutter-ts

And mention that in the docs: https://jupyterlite.readthedocs.io/en/latest/extensions.html (could be part of https://github.com/jupyterlite/jupyterlite/issues/393)

bollwyvl:
>  just swap the implementation of the different services

Yeah, this feels good: now the routes are written in terms of interfaces, not concrete implementations e.g. `IContents` vs `Contents`.

The only thing that still smells kinda off for a "final" API (i mean, heck, it's still `0.1.0`) is that the HTTP responses are just typed with `string | null`... but we probably don't need to solve everything at this point... perhaps in some future state where we're generating typings from the OpenAPI spec would help.

> sample plugin the serverlite cookiecutter

yeah, very good. i think overloading any of the services isn't ready for someone to do without a thorough reading of the code ... though, anywhere where we can make the default services more composable than the upstream winner-takes-all, is good.  However, completely replacing e.g. contents would be a good long-form tutorial, e.g. integrating with an existing API, using multiple domains to improve security, etc.

bollwyvl (approved):
might as well move forward: we won't know until we really start trying to implement some non-trivial stuff downstream

jtpio:
Alright let's get this in.

PR #515, opened 2022-02-20
[wip] improve selective inclusion with `--apps`, add `--no-sourcemaps`

### LINKED ISSUE #475: Add more build/deployment size optimization options
### References

- continues #118
- follow-on to #474

### Problem

For some use cases, the as-deployed size of the site is an important metric. The baseline size of a site on disk is still quite large, and could be optimized in a few ways.

### Proposed Solution

Offer some complementary approaches, and document the trade-offs:

- offer `no_sourcemaps` which will strip out sourcemaps (even from labextensions)
  - this _slashes_ the as-deployed size by a large margin
    - but makes the site almost undebuggable
- improve the `apps` config option (perhaps coupled with a negative `prune_apps`) which does a better job of cleaning up per-app assets
  - this will only save a modest 250k until additional core `@jupyterlab` and `@jupyterlite` modules are federated 
  - it _may_ be possible to further tree-shake the assets only needed by a pruned app, but this might be fragile

### Additional context

As of #474, while the data-to-first-pixels is still around 3.5M (uncompressed), some significant progress has been made to the overall size of the application:

| artifact                  | as of #474 | without sourcemaps |
|---------------------------|-----------:|-------------------:|
| as-downloaded `whl` file  | 8.9M       | 3.5M               |
| as-deployed empty site    | 35M        | 12M                |
| as-deployed docs app      | 57M        | 36M                |

bollwyvl:
## References

- fixes #475

## Code changes

- [x] add `--no-sourcemaps`
- [x] if provided, apps not included in `--apps` will be removed from `build/{app}/bundle.js(.map)`
- [ ] remove `build/<chunk>.<hash>.js` files ununsed by remaining apps
  - _this may be a bridge too far... parsing js is kind of a mugs game, but we pretty much know what we need_
- [x] remove hard-coded list of apps python-side (only appears in string)
  - reading it from the (potentially large) app bundle doesn't seem particularly useful
- [x] clean up some tests, messages

## User-facing changes

- n/a

- probably nothing: it's unlikely the `.apps` key was being used anywhere (`repl` didn't get added to it, for example)


bollwyvl:
A quick benchmark, a couple different runs as tarballs:

```
7.8M 0-full.tar.gz
7.6M 1-repl.tar.gz
7.5M 2-repl-no-unused-shared.tar.gz
2.3M 3-repl-no-unused-shared-maps.tar.gz
```

bollwyvl:
I can dig further into the win pypy thing if there's nothing else... most likely, would just skip the offending test, as it's been flakier of late...

jtpio (approved):
Thanks!

PR #516, opened 2022-02-21
Fix markdown table in the docs

taigaozawa:
I fixed a markdown table whose layout was broken in the docs.


jtpio:
Running script "['yarn run prettier:fix']" on behalf of "jtpio"

jtpio (approved):
Thanks!

For reference this is what `latest` looks like:

image

With the fix:

image

PR #521, opened 2022-02-24
Update CI test matrix

bollwyvl:
## References

- last several merges to `main` failing reliably on `linux pypy-3.7`
 
## Code changes

- [x] bump top `python-version` to `3.10`
  - continues to use `3.9` for docs build/test
- [x] under pypy
  - [ ] ~~bump to `pypy-3.8`~~
    - can't yet because `cloudpickle` uses `_pickle`
      - fixed by https://github.com/cloudpipe/cloudpickle/pull/461
  - [x] drops windows altogether
    - funny story, it wasn't running there anyway, apparently 
  - [x] skips archive tests?
    - just do this on CI with an env var: might be a GHA issue
  - [x] doesn't run tests under coverage
    - it's super slow, might improve in the future
- [x] bump to nodejs version to latest LTS 16
- [x] add filesystem cache for webpack (saves ~40s) 

## User-facing changes

- n/a

- n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -184,7 +184,7 @@ jobs:
             build/pytest
 
   docs:
-    needs: [build, lint-js-test]
+    needs: [build]
     runs-on: ubuntu-latest
     env:
       DOCS_IN_CI: 1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
         uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.9
+          python-version: '3.10'
       - name: Setup pip (base)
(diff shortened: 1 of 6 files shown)


jtpio (approved):
Thanks!

.github/workflows/build.yml
```diff
@@ -167,7 +184,7 @@ jobs:
             build/pytest
 
   docs:
-    needs: [build, lint-js-test]
+    needs: [build]
```
jtpio, line 187: Looks like typedoc is built and uploaded in the lint job:

https://github.com/jupyterlite/jupyterlite/blob/989395371072ff32b7d407af5b05b37f77c504fd/.github/workflows/build.yml#L83-L89

Which makes the `docs` check flaky as these artifacts might not be available when the docs job runs.

bollwyvl: ah, good catch.

Issue #511, opened 2022-02-18
Update to the JupyterLab 3.3 packages


jtpio:
JupyterLab 3.3 is almost there, and we should consider updating to it soon.

There is a corresponding PR on RetroLab: https://github.com/jupyterlab/retrolab/pull/351

We'll probably want to wait for the 3.3 final release to land before making the update here.

PR #528, opened 2022-03-03
Update to JupyterLab 3.3.0 and RetroLab 0.3.20

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/511

## Code changes

- [x] Update to the latest `@jupyterlab` and `@retrolab` packages
- [x] Fix missing plugins for `retro`: toolbar registry
- [x] Update reference snapshots
- [x] Fix UI tests timeout: disabled toolbar items

## User-facing changes

Some changes coming from: https://github.com/jupyterlab/jupyterlab/releases/tag/v3.3.0


jtpio:
Looks like the UI tests timeout might be relevant, some toolbar items seem to be disabled:

image

jtpio:
> Looks like the UI tests timeout might be relevant, some toolbar items seem to be disabled:

Looks like lab has some logic to trigger a `app.commands.notifyCommandChanged();`:

https://github.com/jupyterlab/jupyterlab/blob/2f648bca377ad6074344f3eae21e3cee90a25363/packages/application-extension/src/index.tsx#L529-L533

We disable the corresponding plugin here:

https://github.com/jupyterlite/jupyterlite/blob/8a7590ed38f0fd0ed949e4a64aded956facf6a9f/app/lab/index.template.js#L23

packages/application-extension/src/index.tsx
```diff
@@ -428,6 +428,22 @@ const liteLogo: JupyterFrontEndPlugin<void> = {
...
+ * A plugin to trigger a refresh of the commands when the shell layout changes.
+ */
+const notifyCommands: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlite/application-extension:notify-commands',
+  autoStart: true,
+  optional: [ILabShell],
+  activate: (app: JupyterFrontEnd, labShell: ILabShell | null) => {
+    if (labShell) {
+      labShell.layoutModified.connect(() => {
+        app.commands.notifyCommandChanged();
```
jtpio, line 441: Not really happy about this, as this is hinting for some race conditions upstream in lab as to how the toolbar items are now populated (based on the settings).

This might also be because the lite settings endpoint sends the response very quickly (faster server roundtrip), which removes a delay JupyterLab might expect to find.

jtpio: cc @fcollonval who might know more

jtpio:
> Looks like lab has some logic to trigger a `app.commands.notifyCommandChanged();`:

https://github.com/jupyterlite/jupyterlite/pull/528/commits/874062f823f960b4a4681e635f230b77081bf2a1 seems to be fixing it in the UI tests. also on the RTD preview:

image

jtpio:
@bollwyvl probably this can still go in even with the additional plugins added in https://github.com/jupyterlite/jupyterlite/commit/874062f823f960b4a4681e635f230b77081bf2a1.

So it does not block https://github.com/jupyterlite/jupyterlite/pull/525.

Happy to make a `0.1.0b1` release with this when it lands. So we can also update the try.jupyter.org deployment (https://github.com/jupyter/jupyter.github.io/pull/682).

packages/application-extension/src/index.tsx
fcollonval: I saw that line a while ago and don't like it neither. I don't have any clue why it was made. But it definitely hide explicit call at more appropriate places.
For the bug you are seeing, it points towards the toolbar being build before the notebook being set as the current widget.

fcollonval: It may be more appropriate to trigger the notification when the current widget changes rather than the broader layout change signal.

jtpio: Looks like this code is 4 years old in the lab code base, so yeah something that's been around for a while.

bollwyvl:
I don't think anybody's downstreams are broken waiting for #525, but feel like we want to start kicking the tires on this.

package.json
```diff
@@ -77,9 +81,5 @@
     "typedoc-plugin-markdown": "~3.11.13",
     "typescript": "~4.5.2",
     "yarn-deduplicate": "^3.1.0"
-  },
-  "resolutions": {
-    "json-schema": "^0.4.0",
-    "validator": "^13.7.0"
```
bollwyvl, line 83: hooray

jtpio: I think these have been moved above in the file and not removed (but maybe they can be removed).

bollwyvl: ah, nm.

also, might we normalize some of the `~` and `^` usage? or is the distinction between app code (where `~` is correct) and library code (where `^` makes more sense).

bollwyvl (approved):
RTD looks good, might as well let it rip

jtpio:
Nice. Starting a new release now.

PR #532, opened 2022-03-04
Update content creating method in `_getServerContents`

### LINKED ISSUE #530: Error on opening medium size binary file
## Description

 In `jupyterlite`,  I got the following error message `Maximum call stack size exceeded` on opening a ~500kb binary file. It works well with the smaller files (~100kb). The error stack pointed to this call: https://github.com/jupyterlite/jupyterlite/blob/03da209a8b82242a92c85075ddfb5425337f0621/packages/contents/src/contents.ts#L594-L596

## Reproduce

Build a `jupyterlite` site with some binary files then open it from the file browser.

## Expected behavior

The files should be opened correctly. 

## Context

- JupyterLite version: latest

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

trungleduc:
## References
Fixes #530 

## Code changes

The current implementation of content base64-encoding can not handle files with medium size as described in #530.  This PR introduces a more robust way to handle this task. 

## User-facing changes
- Before:
before

- After:
after3

N/A

### DIFF
--- a/packages/contents/src/contents.ts
+++ b/packages/contents/src/contents.ts
@@ -589,11 +589,14 @@ export class Contents implements IContents {
             mimetype: mimetype || 'text/plain',
           };
         } else {
+          const byteToString = (data: string, byte: number) =>
+            data + String.fromCharCode(byte);
+          const content = btoa(
+            new Uint8Array(await response.arrayBuffer()).reduce(byteToString, '')
+          );
           model = {
             ...model,
-            content: btoa(
-              String.fromCharCode(...new Uint8Array(await response.arrayBuffer()))
-            ),
+            content,
             format: 'base64',
             mimetype: mimetype || 'octet/stream',
           };


bollwyvl:
Nice. Perhaps hoist the new anonymous function to a helper so it's not
getting re-created every time? I mean, it's very short, so it probably gets
optimized away anyway, but can't hurt....

packages/contents/src/contents.ts
```diff
@@ -589,11 +589,14 @@ export class Contents implements IContents {
             mimetype: mimetype || 'text/plain',
           };
         } else {
+          const byteToString = (data: string, byte: number) =>
```
bollwyvl, line 592: Sorry, was on my phone earlier: i meant hoist _outside_ the function (or even the class) so that we're not defining a fresh function/scope every time we use this. BUT: turns out we _already_ define a bunch of anonymous functions all over this file, so it's unfair to ask for micro-optimizations for new contributions. I've raised #53, so let's not worry about it here!

bollwyvl:
This is great, will try out on RTD and probably merge.

Do you have any sense of how big a file had (in kb/mb)  to be before/after this change to crap out? It would probably be useful to document this someplace, though I'm sure it will change between browsers, or even different versions of the same browser.

bollwyvl:
Hm... I still had some issues _uploading_ binary files, but this was also an issue on master. Will issue up, but otherwise LGTM.

Issue #315, opened 2021-09-01
Normalize and make Content frontends and backends extensible


bollwyvl:
### Problem

We currently offer a union of files, backed by two storage _backends_, in the following order:

- pre-baked contents, as indexed in a file tree that roughly mirrors the Jupyter Contents REST API
- in-browser contents, managed by `localForage`

...with one storage _frontend_:

- a mock REST API, exposing Jupyter Contents

While this makes it _possible_ to deliver the Lab file management UI, such as it is, it is not very ergonomic for:

- accessing from kernels
- providing alternate storage implementations e.g.
  - WebDAV (a la sharepoint, etc)
  - sqlite
  - git
  - fossil-scm
  - some rando REST API

### Proposed Solution

- formalize and define an API for the existing backends
  - these would become two separate server extensions (#104)
- separate the mock REST API frontend from the provider
- add an emscripten-compatible `FS` for the overall contents

### Additional context

- some upstream thoughts about `FS`: https://github.com/pyodide/pyodide/issues/1715#issuecomment-900122795
- this may have impacts on #310
  - packages are files, too!

oeway:
FYI: I have been also looking into the storage options, and what I am using right now is using BrowserFS inside a ServiceWorker so we can manage files using varies backends (e.g. indexeddb), and the service worker can work as a in-browser file server. This allows all unified file access via HTTP, and for pyodide kernel, we can do XMLHTTPRequest in synchronous mode and create file-like object that works with large files.

Here is an demo shows how it work in a custom deployment of jupyterlite (available at https://jupyter.imjoy.io ):
![elfinder-storage-demo](https://user-images.githubusercontent.com/478667/146198453-a90d2317-e217-4cb1-aee5-840d9c06a43a.gif)

Here is the notebook if you want to try it out: https://github.com/imjoy-team/jupyter.imjoy.io/blob/master/docs/files/elfinder-demo.ipynb

jtpio:
Thanks for sharing this @oeway it looks cool!

bollwyvl:
I've updated the issue description with a link to the emscripten filesystem built for the (maybe) sleeping starboard: of note, there is a link to the (even more dormant) BrowserFS example at the top, which has even better typing info, which is probably the most important to get started. 

[...]

PR #547, opened 2022-03-11
Add localforage memory fallback

### LINKED ISSUE #535: Enable storage fallback when locastorage and IndexedDB aren't available
### Problem

Currently, when both `localstorage` _and_ `IndexedDB` are not available (due to iframe settings or browser limitations), even the `repl` app is basically unusable, as a number of things rely on being able to update _something_ (usually settings).

Out of the box, there is no solution to this, but `localforage` _does_ have an opt-in, non-persistent storage mechanism which would be reasonable for a number of use cases, but relies on side-effects of the `localforage` global instance. However, "drops all data without warning" doesn't seem like a very good default, and should either not be enabled by default, or throw a bunch of warnings.

### Proposed Solution

- [ ] hoist the `localforage` singleton to a service
- [ ] adopt [`localforage-memoryStorageDriver`](https://github.com/localForage/localForage-memoryStorageDriver)
  - [ ] add a server plugin that installs that driver
  - [ ] add an opt-in `PageConfig` setting, e.g. `"enableMemoryStorageFallback": true`
- [ ] ideally fix the potential race condition for enabling memory storage 
  - [ ] #525 did this with optional dependencies, but this wouldn't scale to other drivers 
  - [ ] could also consider making all of the storage-related things promises, so there's enough time for all the extensions to update

### Additional context

This was partially demonstrated [over on #525](https://github.com/jupyterlite/jupyterlite/pull/525/files#diff-af39f1d59798653f5550af7684059a99f64aeb3258f8738fbf9d371b26aac08aR33) but was kind of unrelated to that.

bollwyvl:
## References

- fixes #535
- expands on #525

## Code changes

- [x] extracts the localforage singleton to a plugin
- [x] adds a plugin for memory storage
- [x] adds a `enableMemoryStorage` config value to enable memory storage
  - shipping default `disabledPlugins` seems... bad  
  - was enabled by default on #525, but can-lose-data-by-default is not great
- [x] adds `ready` promises to settings and contents
  - this allows enough time for extra drivers to be installed
- [x] add `(settings|contents)StorageDrivers` config value
  - if left as the default (`null`), will fall back to the "best available" (apparently `IndexedDB`)
- [x] disable persistence on `repl` in docs app 
- [x] update configuring docs

## User-facing changes

- iframe-embedded apps, e.g. `repl` can be configured without extra `sandbox` permissions and still work
- while the as-distributed `repl` will have persistent settings...
  - on _our_ docs site, settings will _not_ persist between page loads
    - this appears to impact theme/scrollbars/font size... kind of annoying to lose, but ultimately `repl` is more about _publisher_ control, anyway, and we need to keep this feature under test/demonstration

- adds ready promise to `IContents` and `ISettings`
- shortened some `(Settings|Contents).Options` members (remove the `settings` and `contents` prefixes)

## Test Procedure

- Go to the RTD preview
- See lab
  - in the _Application_ tab in dev tools, see the IndexedDB
- in the same tab, go to the repl app
  - delete the `JupyterLite` IndexedDB
- choose a different theme through `Ctrl+Shift+C`modal
- reload the repl, see that the theme reverts to the default
- see that no new IndexedDB is created

## Future Work

- there are a number of other localforage drivers
  - some of these might be interesting to explore as extensions
    - these could add the drivers and subclass the existing settings and contents, and overload the `createDefault*Storage` methods to create interesting effects
  - ...but probably don't need to go in core


app/jupyterlite.schema.v0.json
```diff
@@ -105,11 +105,24 @@
           "default": "./build",
           "$ref": "#/definitions/non-trailing-slash-uri"
         },
+        "enableMemoryStorage": {
+          "description": "enable memory storage fallback which will can lose data if the page is reloaded",
```
jtpio, line 109:
```suggestion
          "description": "enable memory storage fallback which can lose data if the page is reloaded",
```

docs/configuring.md
```diff
@@ -181,6 +186,45 @@ JupyterLite instances hosted there. To create separate settings stores, change t
...
+### Volatile Memory Storage
+
+Many extensions and features require the ability to at least _think_ they are saving and
+loading contents and settings. If a user's data cannot be stored due to browser security
+settings, a JupyterLite app will generally fail to fully initialize: while this might be
+frustrating, losing a user's unique data creation is _even more_ frustating.
+
+```{warning}
+If persistence is **entirely** handled outside of JupyterLite, e.g. in an embedded
+[`repl`](./applications/repl.md) it is possible to disable all persistence, assuring
```
jtpio, line 222: The `repl` app was already not persisting data to the browser since it's based on the ephemeral lab code console.

But agree switching to an in-memory storage for this app will help to eventually allow for other widgets (for example https://github.com/jupyterlite/jupyterlite/issues/148) :+1:

packages/localforage/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@jupyterlite/localforage",
```
jtpio, line 2: Do we want to follow the name of the underlying library, or use something a bit more generic like `@jupyterlite/persistence`?

docs/configuring.md
bollwyvl: > not persisting data to the browser

it _would_ persist settings

packages/localforage/package.json
bollwyvl: But yeah, you know me: in theory i wish it could be generic! I think the larger refactor of #315 can address some more of these issues, but trying to abstract our storage layer at this point feels... premature. This gets us to a "doesn't fail by default unless you have insecure iframe settings," which seems like progress, and allows for at least integrating with more localforage options.

It's kinda like codemirror: it's really good at what it does (e.g. fallback) but has a lot of global state in The Singleton `localforage`. Even with `sharedPackages`, etc. weird stuff can still happen, so i wanted it to be very explicit. 

If one wants a new storage driver, there are _way_ more localforage examples to look at and emulate than pretty much our _one example_ of contents and settings persistence, so encouraging hooking that by example is the best I figured we can do (even if i didn't write the driver).

As present, one would have to pretty much replace the `IContents` and `ISettings` providers wholesale to change the behavior much, otherwise (e.g. use a new or existing REST server), and as such those _don't_ have `localforage` anywhere on their boundaries, and only the routes consume _those_, so i think we're covered. Even then, maybe a localForage API is useful in its own right.

docs/configuring.md
jtpio: > it would persist settings

Ah right sure, for example themes and other user settings.

bollwyvl: right: the joke on #525 was, i tried to open an iframe to `repl` from inside a notebook (hosted on the same site), and by default a lot of stuff was broken... the memory fallback enables a way to require slightly fewer sandbox exceptions to work "properly,"

jtpio (approved):
Thanks!

PR #554, opened 2022-03-15
Update to JupyterLab 3.3.2 packages

jtpio:
## References

Small update to the latest JupyterLab packages.

## Code changes

Dependency update.

## User-facing changes

Some bug fixes.

None


bollwyvl:
Mmmm. Do want, if only for how much quieter the console is... we're down to just the `leaflet` warning caused by https://github.com/QuantStack/leaflet-splitmap/issues/7

bollwyvl (approved):
LGTM

Issue #553, opened 2022-03-15
HookWebpackError: ENOENT: no such file or directory, scandir


danielemoroni:
## Description

I did a fresh download and install, and I get the error in the title.

## Reproduce

I literally followed the contributing guide. After downloading mambaforge:
1. git clone https://github.com/jupyterlite/jupyterlite
2. mamba env update --file .binder/environment.yml
3. mamba activate jupyterlite-dev
4. python -m pip install -r requirements-docs.txt -r requirements-lint.txt
5. doit

and I get the error

> .  build:js:app
yarn run v1.22.17
$ C:\Users\DanieleMoroni\Documents\dev\jupyterlite\node_modules\.bin\lerna run build:prod --scope @jupyterlite/app
lerna notice cli v4.0.0
lerna info versioning independent
lerna notice filter including "@jupyterlite/app"
lerna info filter [ '@jupyterlite/app' ]
lerna info Executing command in 1 package: "yarn run build:prod"
lerna ERR! yarn run build:prod exited 2 in '@jupyterlite/app'
lerna ERR! yarn run build:prod stdout:
$ yarn clean && yarn build --mode=production
$ rimraf build **/build
$ webpack --mode=production
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run build:prod stderr:
[webpack-cli] ←[31mHookWebpackError: ENOENT: no such file or directory, scandir
    at makeWebpackError (C:\Users\DanieleMoroni\Documents\dev\jupyterlite\node_modules\webpack\lib\HookWebpackError.js:48:9)
    at C:\Users\DanieleMoroni\Documents\dev\jupyterlite\node_modules\webpack\lib\Compilation.js:3059:12
    at eval (eval at create (C:\Users\DanieleMoroni\Documents\dev\jupyterlite\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:23:1)
    at fn (C:\Users\DanieleMoroni\Documents\dev\jupyterlite\node_modules\webpack\lib\Compilation.js:480:17)
    at _next9 (eval at create (C:\Users\DanieleMoroni\Documents\dev\jupyterlite\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:21:1)
[...]

PR #557, opened 2022-03-16
WebPack tweaks for fonts, licenses, Windows

### LINKED ISSUE #550: ipympl buttons not properly rendered
## Description

When displaying a matplitlib figure using `ipympl`, the buttons are not properly rendered:
image

## Reproduce

Open and execute [this notebook](https://jupyterlite.readthedocs.io/en/latest/_static/lab?path=pyolite%2Fipympl.ipynb).

bollwyvl:
## References

- includes #554
- fixes #550 (apparently)
- works around #553

## Code changes

- replaces `file-loader` and `url-loader` usage with `asset/resource`
- adds `NO_WEBPACK_LICENSES`, doesn't use that plugin
  - defaults to active on windows
- adds a windows dev job 
  - mamba, based on the binder env
  
## User-facing changes

- font-awesome should work

- n/a


bollwyvl:
Welp, windows is rather slow, but we knew that.

If we wanted to further optimize, we could move it to a less-frequently-run job (e.g. only on merges to `main` and/or changes to dodo/yarn.lock/env.yml), but i think simple beats fast in this case.

app/webpack.config.js
```diff
@@ -222,6 +228,26 @@ class CompileSchemasPlugin {
   }
 }
 
+/**
+ * A helper for filtering deprecated webpack loaders, to be replaced with assets
```
jtpio, line 232: Maybe the upstream base webpack config should be updated to assets too at some point.

jtpio (approved):
Thanks!!

PR #562, opened 2022-03-17
Clarify RTC config setting

psychemedia:
Docs currently assume knowledge of where config settings are expected to reside. Clarify where settings changes to support RTC need to be made.

## Code changes

None.

Changes to docs only.


jtpio (approved):
Looks good on RTD, thanks!

image

Issue #569, opened 2022-03-19
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain" 


liudonghua123:
## Description

I tried to execute `jupyter lite init` and `jupyter lite serve`, then I opened `http://127.0.0.1:8000/index.html` in the latest chrome 99 browser. The page is blank and I check the console log and found `ailed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.` around `config-utils.js:1 `.

**_If I use `serve` of npm to serve the static files. No such problems._**

So it's maybe a simple python web server MIME type config error. I will try to check the code and fix it.

![image](https://user-images.githubusercontent.com/2276718/159116542-e9c56926-af06-4cac-87e0-69fee3fbfc91.png)

## Reproduce

1. `jupyter lite init`
2. `jupyter lite serve`
3. Open `http://127.0.0.1:8000/index.html` in the latest chrome 99 browser

## Context

- JupyterLite version: 0.1.0b3
- Operating System and version: Microsoft Windows [Version 10.0.22000.556]
- Browser and version: Chrome 99

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

liudonghua123:
I find the actural serve used is `tornado`, and I also tried `python -m http.server --directory _output 5000`, the error is the same. So both `tornado` and `http.server` should config the mime type mappings directly.

The code need to update maybe the following.

https://github.com/jupyterlite/jupyterlite/blob/5889b98f0f3df0317d87a518a3e9294c9f0edf0a/py/jupyterlite/src/jupyterlite/addons/serve.py#L57-L68

```python
# a demonstation code from https://gist.github.com/HaiyangXu/ec88cbdce3cdbac7b8d5?permalink_comment_id=3275090#gistcomment-3275090
import http.server
import socketserver

PORT = 8080

class HttpRequestHandler(http.server.SimpleHTTPRequestHandler):
    extensions_map = {
        '': 'application/octet-stream',
        '.manifest': 'text/cache-manifest',
        '.html': 'text/html',
        '.png': 'image/png',
        '.jpg': 'image/jpg',
        '.svg':	'image/svg+xml',
        '.css':	'text/css',
        '.js':'application/x-javascript',
        '.wasm': 'application/wasm',
        '.json': 'application/json',
[...]

PR #570, opened 2022-03-19
Update serve.py

liudonghua123:
try to fix https://github.com/jupyterlite/jupyterlite/issues/569


bollwyvl:
Implemented with configurable file types in #670

PR #583, opened 2022-03-29
ensure addons can be configured, more docs

bollwyvl:
## References

- discovered during https://github.com/robots-from-jupyter/robotkernel/issues/72#issuecomment-1082295709

## Code changes

- [x] ensures a `BaseAddon` also uses the `manager` as `parent`, in order to load config
  - [x] test
- [x] moves some docs from the manager class to a dedicated docs page 

## User-facing changes

- n/a

- n/a


docs/extensions.md
```diff
@@ -111,7 +111,7 @@ python -m pip install myliteextension
 In most cases, this would mean adding the extension to a `requirements.txt` file:
 
 ```
-jupyterlite==0.1.0a12
+jupyterlite==0.1.0b4
```
bollwyvl, line 114: do we have this under lint right now?

py/jupyterlite/pyproject.toml
```diff
@@ -91,18 +91,18 @@ all = [
 ]
 
 [project.entry-points."jupyterlite.addon.v0"]
-static = "jupyterlite.addons.static:StaticAddon"
-federated_extensions = "jupyterlite.addons.federated_extensions:FederatedExtensionAddon"
-settings = "jupyterlite.addons.settings:SettingsAddon"
+archive = "jupyterlite.addons.archive:ArchiveAddon"
```
bollwyvl, line 94: sorted

py/jupyterlite/src/jupyterlite/addons/base.py
```diff
@@ -33,6 +33,11 @@ class BaseAddon(LoggingConfigurable):
 
     manager: LiteManager = Instance(LiteManager)
 
+    def __init__(self, manager, *args, **kwargs):
+        kwargs["parent"] = manager
+        kwargs["manager"] = manager
+        super().__init__(*args, **kwargs)
+
```
bollwyvl, line 40: @datakurre in the near term you can  just apply this to your addon, if you need to...

docs/extensions.md
jtpio: Do you mean bumping the version automatically in this file?

If so then it's not the case (but could be handled).

Otherwise we can just drop the version here:

```suggestion
jupyterlite
```

bollwyvl: Well, i'm more about version _linting_ than bumping. More so that most stuff, I think a version pin is going to be clutch.

So i'll update it to be "correct" (without a version) but add a note about _YOU SHOULD ALWAYS PIN THE VERSION_

jtpio (approved):
Thanks!

PR #600, opened 2022-04-13
Patch WebRTC collaboration (y-webrtc) to use buffering 

### LINKED ISSUE #598: Collaborative WebRTC disconnects with big enough update
## Description
 
Collaborative WebRTC silently disconnects with too big display data being transferred.

## Reproduce

```python
from IPython.display import display
payload = "a" * 1024 * 256
display({"text/html": f"OK<!--{payload}-->"}, raw=True)
```

y-webrtc logging enabled with
```js
// enable logging for all modules
localStorage.log = 'true'
```

webrtc-disconnects

## Expected behavior

WebRTC would not disconnect silently. Either support large payloads or gracefully degrades.

## Context

RobotKernel injects execution logs into notebook by displaying HTML links with large inline data URIs. Not beautiful, but has been working too well so far:

https://robotkernel.readthedocs.io/en/latest/_/lab/index.html?path=Example.ipynb&room=demo

/cc @bollwyvl

datakurre:
Fixes #598 (WebRTC disconnect when loading large notebook or executing cell with large display payload)

### DIFF
--- a/app/webpack.config.js
+++ b/app/webpack.config.js
@@ -301,14 +301,9 @@ module.exports = [
           loader: 'string-replace-loader',
           options: {
             search: 'simple-peer/simplepeer.min.js',
-            replace: [
-              '..',
-              '..',
-              '..',
-              'app',
-              'vendor',
-              'SimplePeerExtended.js'
-            ].join(path.sep),
+            replace: ['..', '..', '..', 'app', 'vendor', 'SimplePeerExtended.js']
+              .join(path.sep)
+              .replace(/\\/g, '\\\\'),
           },
         },
       ],
--- a/app/webpack.config.js
+++ b/app/webpack.config.js
@@ -295,6 +295,22 @@ module.exports = [
           test: /\.(jpe?g|png|gif|ico|eot|ttf|map|woff2?)(\?v=\d+\.\d+\.\d+)?$/i,
           type: 'asset/resource',
         },
+        // Fix WebRTC buffered transmission: https://github.com/yjs/y-webrtc/pull/25
+        {
+          test: /y-webrtc\.js$/,
+          loader: 'string-replace-loader',
+          options: {
+            search: 'simple-peer/simplepeer.min.js',
+            replace: [
+              '..',
+              '..',
+              '..',
+              'app',
+              'vendor',
+              'SimplePeerExtended.js'
+            ].join(path.sep),
+          },
+        },
       ],
     },
     optimization: {
--- a/app/webpack.config.js
+++ b/app/webpack.config.js
@@ -301,7 +301,7 @@ module.exports = [
           loader: 'string-replace-loader',
           options: {
             search: 'simple-peer/simplepeer.min.js',
-            replace: ['..', '..', '..', 'app', 'vendor', 'SimplePeerExtended.js']
+            replace: ['./', '..', '..', '..', 'app', 'vendor', 'SimplePeerExtended.js']
               .join(path.sep)
               .replace(/\\/g, '\\\\'),
           },
--- a/.prettierignore
+++ b/.prettierignore
@@ -26,3 +26,6 @@ CHANGELOG.md
 
 # UI tests
 ui-tests/ui-tests-app
+
+# vendored
+app/vendor/int64-buffer.min.js
--- a/app/package.json
+++ b/app/package.json
@@ -30,6 +30,7 @@
     "mini-css-extract-plugin": "^2.5.3",
     "raw-loader": "^4.0.2",
     "rimraf": "^3.0.2",
+    "string-replace-loader": "^3.1.0",
     "style-loader": "^3.3.1",
     "svg-url-loader": "^7.1.1",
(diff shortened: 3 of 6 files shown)


bollwyvl:
I'm interested to see this running on RTD, but again: i don't think we want to carry this patch at this point, but rather would want this as an alternative implementation, e.g as a labextension.

It looks like we would need a [dependency-injection provider](https://github.com/jupyterlite/jupyterlite/blob/34921cbb4cb5216d6149ef5b7cef1edb131d687f/packages/application-extension/src/index.tsx#L284
) that listened to `?room` and could read off the same settings as `jupyter-config-data`.

datakurre:
@bollwyvl You can try this out at RTD with some crazy previously failing display calls like

```python
from IPython.display import display
payload = "a" * 1024 * 256 * 40
display({"text/html": f"OK<!--{payload}-->"}, raw=True)   
```

datakurre:
The issue / pull at y-webrtc has been inactive for a year, so this patch should probably be carried to where ever the collaboration is refactored.

bollwyvl:
> need a dependency-injection provider

but we might also need to do some stuff to the webpack setup so the stock provider doesn't get bundled...

app/webpack.config.js
```diff
@@ -295,6 +295,22 @@ module.exports = [
...
+            replace: path.join(
+              '..',
+              '..',
+              '..',
+              'app',
+              'vendor',
+              'SimplePeerExtended.js'
+            ),
+          },
+        },
```
datakurre, line 313: Initially tried absolute path to `SimplePeerExtended.js` here, but could not make it work on Windows. This ugly relative path (assuming node_modules location) might just work.

datakurre: Still broken. Is it really that `path.join()` does not join parts with `path.sep` on Windows :flushed:

datakurre: Or maybe `string-replace-loader` breaks Windows `path.sep` on replace.

datakurre:
> i don't think we want to carry this patch at this point, but rather would want this as an alternative implementation, e.g as a labextension.

Do you mean making collaboration an optional labextension, or just this buffered WebRTC a labextension? Because without this patch, collaboration is not only broken for robotkernel, but also for any notebook saved with too much data. Loading any too big notebook with collaboration will disconnect WebRTC without this buffering.

bollwyvl:
> making collaboration an optional labextension

Right: if it's broken as-is, and the upstream is not healthy, perhaps we need to take the yjs stack out of the lite core build for the time being so that a patched one can be provided by a labextension.

This "@jupyterlite/docprovider-webrtc," distributed on PyPI as a standard labextension, could then be used in the RTD demo... and would work for "real" lab, as well, so might be more appropriate anyway.

datakurre:
> This "@jupyterlite/docprovider-webrtc," distributed on PyPI as a standard labextension, could then be used in the RTD demo... and would work for "real" lab, as well, so might be more appropriate anyway.

That makes sense.

I patched https://robotkernel.readthedocs.io/en/latest/_/lab/index.html to be able to demo collaboration with robot suites (up to widgets...).

Should drawio collaboration be real-time yet? Diagrams get synchronized on save, but need to be reopened to see the changes.

bollwyvl:
> demo collaboration with robot suites

Cool!

> up to widgets...

I think getting to that fully working is part of the carrot of upgrading to the upcoming ipywidgets 8 stuff? the widget graph usually isn't being serialized to the notebook format fast enough to work well, and i think they had to do something new...

> drawio collaboration be real-time yet?

I don't know the state of it at the moment: on [a fork](https://github.com/deathbeds/ipydrawio) i did some stuff to enable the collaborative stuff in lite, but haven't looked into it much. as that fork more directly consumes the upstream UI: the `17` line of the drawio ui has been doing a _lot_ of work around collaborative editing: tying more directly into the jupyter implementation that might be very powerful indeed.

datakurre:
```
[webpack-cli] ModuleNotFoundError: Module not found: Error: Can't resolve '..\..\..\app\vendor\SimplePeerExtended.js' in 'D:\a\jupyterlite\jupyterlite\node_modules\y-webrtc\src'
Did you mean './..\..\..\app\vendor\SimplePeerExtended.js'?
```

Sounds close. Even we don't merge this, I must try one more time to see if that suggestion works.

bollwyvl:
Thanks for sticking with it. It's a vicious cycle: with few sustained contributions from windows users, it's hard to keep the machine running while building crazy new things... but without keeping the machine running,  we _definitely_ won't get any contributions from windows users.

datakurre:
Now I've seen a Webpack error message that actually helped to solve the issue!

bollwyvl:
Fired this up over here:

https://github.com/jupyterlite/jupyterlab-webrtc-docprovider

jtpio:
Closing as fixed by https://github.com/jupyterlite/jupyterlite/pull/615.

Thanks @datakurre @bollwyvl!

PR #611, opened 2022-04-21
Switch to `standby: 'never'` for the server polling to ensure kernel specs are ready

### LINKED ISSUE #610: No kernel when the browser tab doesn't have the focus
## Description

The REPL application sometimes displays an empty list of kernels on startup:

image

## Reproduce

1. Go to https://jupyterlite.rtfd.io/en/latest/try/repl or https://jupyterlite.github.io/demo/repl/
2. Quickly open a new browser tab
3. Wait a couple of seconds
4. Go back to the JupyterLite tab

https://user-images.githubusercontent.com/591645/164416198-1d21b010-18ec-4db0-9777-b6a1b2545634.mp4

## Expected behavior

The kernels should be available on startup even if the tab doesn't have the focus.

## Context

This was initially reported in https://github.com/sympy/sympy.github.com/pull/174

jtpio:
## References

Fixes #610 

## Code changes

Switch to `standby: 'never'` when instantiating the `ServiceManager`.

There is no real server polling anyway since the server runs in the browser, so we should be able to use this default.

Various other cleanups.

## User-facing changes

Users opening JupyterLite in a new background tab (for example via Ctrl+click) should still be able to see the list of kernels when switching to the tab later on. 

None

### DIFF
--- a/app/bootstrap.js
+++ b/app/bootstrap.js
@@ -19,6 +19,7 @@
  * assumed that the page HTML includes a script tag with the id
  * `jupyter-config-data` containing the configuration as valid JSON.
  */
+
 let _CONFIG_DATA = null;
 function getOption(name) {
   if (_CONFIG_DATA === null) {
@@ -88,5 +89,5 @@ void (async function bootstrap() {
   // Now that all federated containers are initialized with the main
   // container, we can import the main function.
   let main = (await import('./index.js')).main;
-  window.addEventListener('load', main);
+  void main();
 })();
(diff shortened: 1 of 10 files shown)


jtpio:
This is not only affecting the REPL app, but other apps as well: https://github.com/jupyterlite/jupyterlite/issues/610#issuecomment-1105073113.

Gonna dig a bit more.

jtpio:
This is now looking good on RTD:

https://user-images.githubusercontent.com/591645/166207131-ccc99d20-39ad-438b-9923-3698e0d0d843.mp4

jtpio:
CI now passing with the fix from https://github.com/jupyter-server/jupyter_releaser/pull/308.

jtpio:
Making a new release with that one and other recent fixes.

PR #615, opened 2022-04-22
[wip] Remove code for WebRTC collaboration, use jupyterlab-webrtc-docprovider

### LINKED ISSUE #598: Collaborative WebRTC disconnects with big enough update

bollwyvl:
## References

- fixes #598
- alternative to #600

## Code changes

- [x] makes default docprovider always fall back to the mock provider
- [x] add `jupyterlab-webrtc-docprovider` to binder env
- [x] resolve example extensions
- [x] remove dependency on (unpatched) `y-webrtc` 
- [x] update docs
- [x] note in browser console if `collaborative` is configured but the default, no-op provider is used 

## User-facing changes

- docs site user 
  - will see the "sharing" status bar icon
  - will be able to configure room names, etc. settings

- webrtc will no longer be provided by default


docs/rtc/index.md
```diff
@@ -1,8 +1,17 @@
 # Real Time Collaboration
 
-JupyterLite enables basic support for Real Time Collaboration over WebRTC.
+JupyterLite can be extended to enable basic support for Real Time Collaboration over
+WebRTC.
+
+```{note}
+Prior to version `0.1.0b5`, the WebRTC document provider was built directly into
```
jtpio, line 7: This will be the case in the next `0.1.0b6` release:

```suggestion
Prior to version `0.1.0b6`, the WebRTC document provider was built directly into
```

jtpio, line 7: This will be the case in the next `0.1.0b6` release:

```suggestion
Prior to version `0.1.0b6`, the WebRTC document provider was built directly into
```

jtpio:
Thanks, looking good!

This sound like a good way forward for now.

In the future, maybe we'll still need some RTC logic in core lite, since the upstream work is heading towards moving more things to the backend with `y-py`. Unless this can still be implemented separately as a serverlite extension.

jtpio:
> it would be tragic if things get substantially _more_ complicated around "move files to/from disk," in the name of a use case that really is rather niche, if very cool. Here's hoping not every extension author has to become an expert on yjs schema, etc.

Yeah that's my hope as well that this machinery will be available for those that need it, but still easy to disable if folks are not interested in RTC.

jtpio:
Looks good on RTD!

image

jtpio (approved):
Thanks!

PR #617, opened 2022-04-24
Rename default name from `Consolite` to `REPLite`

jtpio:
## References

Minor follow-up to https://github.com/jupyterlite/jupyterlite/pull/498

## Code changes

Rename the default name for the `REPL` application, mostly for consistency.

## User-facing changes

None

None


bollwyvl:
Hm... there's really no reason why we _couldn't_ have an xterm-based repl, too...

Issue #529, opened 2022-03-03
Re-add language packs to the RTD demo site


jtpio:
### Problem

At some point it was possible to try different display languages on the demo site: https://jupyterlite.rtfd.io/en/latest/try/lab

This does not seem to be the case anymore with the latest `main`:

![image](https://user-images.githubusercontent.com/591645/156592554-f555dd63-3418-4e0a-8669-2c9a63637a64.png)

### Suggested Improvement

The language packs are defined here: 

https://github.com/jupyterlite/jupyterlite/blob/8a7590ed38f0fd0ed949e4a64aded956facf6a9f/docs/environment.yml#L42-L43

And should be picked up by the build.

jtpio:
The latest `jupyterlite/demo` running with beta 0 has them:

![image](https://user-images.githubusercontent.com/591645/156593098-95ad87a3-1018-42ac-abe8-206e47641453.png)

So it's probably just an issue with the RTD demo site.

jtpio:
This is likely related to this check:

https://github.com/jupyterlite/jupyterlite/blob/a5cb58bd37dcfa2a32008c25661e0bc20c547a6d/py/jupyterlite/src/jupyterlite/addons/translation.py#L83

Since the demo site uses the flag:

https://github.com/jupyterlite/jupyterlite/blob/a5cb58bd37dcfa2a32008c25661e0bc20c547a6d/examples/jupyter_lite_config.json#L25

jtpio:
Maybe the simplest would be to add a flag to `jupyterlite.schema.v0.json` to ignore translation bundles or not.

jtpio:
> Although not ideal since `ignore_sys_prefix` implies the translation bundles should be ignored since they are under the prefix.

Or we rename `ignore_sys_prefix` to something more specific to federated extensions / mathjax.

jtpio:
> This is likely related to this check:

We probably want to keep this check or a similar one to avoid side-effects when building from a `jupyter_lite_config.json` with `ignore_sys_prefix: true`

PR #621, opened 2022-04-27
More granular way to ignore components from `sys.prefix`

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/529

## Code changes

- [x] Update the logic around `ignore_sys_prefix` so site builders can choose what to ignore more easily (more granular).
- [x] Update docs

## User-facing changes

Should fix some situations of translations being ignored when building the site with `ignore_sys_prefix`.

~The `ignore_sys_prefix` is a list of components to ignore instead of being a `boolean`~

Since `ignore_sys_prefix` still supports boolean values, previous configurations should still be compatible after this change.

### DIFF
--- a/py/jupyterlite/src/jupyterlite/addons/base.py
+++ b/py/jupyterlite/src/jupyterlite/addons/base.py
@@ -33,10 +33,7 @@ class BaseAddon(LoggingConfigurable):
 
     manager: LiteManager = Instance(LiteManager)
 
-    ignore_sys_prefix: bool = Bool(
-        False,
-        help="ignore components from sys.prefix",
-    ).tag(config=True)
+    ignore_sys_prefix: bool = Bool(False)
 
     def __init__(self, manager, *args, **kwargs):
         kwargs["parent"] = manager
--- a/py/jupyterlite/src/jupyterlite/addons/base.py
+++ b/py/jupyterlite/src/jupyterlite/addons/base.py
@@ -7,7 +7,7 @@ import time
 import warnings
 from pathlib import Path
 
-from traitlets import Instance
+from traitlets import Bool, Instance
 from traitlets.config import LoggingConfigurable
 
 from ..constants import (
@@ -32,6 +32,11 @@ class BaseAddon(LoggingConfigurable):
     """
 
     manager: LiteManager = Instance(LiteManager)
+
+    ignore_sys_prefix: bool = Bool(
+        False,
+        help="ignore components from sys.prefix",
+    ).tag(config=True)
 
     def __init__(self, manager, *args, **kwargs):
         kwargs["parent"] = manager
@@ -276,3 +281,6 @@ class BaseAddon(LoggingConfigurable):
                     is_ignored = True
                     break
         return is_ignored
+
+    def is_sys_prefix_ignored(self):
+        return self.ignore_sys_prefix
--- a/py/jupyterlite/src/jupyterlite/addons/federated_extensions.py
+++ b/py/jupyterlite/src/jupyterlite/addons/federated_extensions.py
@@ -57,7 +57,7 @@ class FederatedExtensionAddon(BaseAddon):
         """yield a doit task to copy each federated extension into the output_dir"""
         root = ENV_EXTENSIONS
 
-        if not manager.ignore_sys_prefix:
+        if not "federated_extensions" in manager.ignore_sys_prefix:
             for pkg_json in self.env_extensions(root):
                 yield from self.copy_one_extension(pkg_json)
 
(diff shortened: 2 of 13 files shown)


py/jupyterlite/src/jupyterlite/config.py
```diff
@@ -64,9 +64,10 @@ class LiteBuildConfig(LoggingConfigurable):
         config=True
     )
 
-    ignore_sys_prefix: bool = Bool(
-        False,
-        help="ignore lab components from sys.prefix, such as federated_extensions",
+    ignore_sys_prefix: _Tuple[str] = TypedTuple(
+        Unicode(),
+        help="list of components from sys.prefix to ignore."
+        "Values: ['federated_extensions', 'mathjax', 'translations']",
```
bollwyvl, line 70: Hm... this is probably in the same domain as `disable_addons`, and could be based on the entry_point key.

py/jupyterlite/src/jupyterlite/addons/federated_extensions.py
```diff
@@ -57,7 +57,7 @@ def pre_build(self, manager):
         """yield a doit task to copy each federated extension into the output_dir"""
         root = ENV_EXTENSIONS
 
-        if not manager.ignore_sys_prefix:
+        if not "federated_extensions" in manager.ignore_sys_prefix:
```
bollwyvl, line 60: maybe each addon hoists this key, and `BaseAddon` grows a `is_sys_prefix_ignored()`?

py/jupyterlite/src/jupyterlite/config.py
bollwyvl: also, perhaps it is perhaps a `Union(Boolean(), TypedTuple(Unicode()))`?

py/jupyterlite/src/jupyterlite/addons/federated_extensions.py
jtpio: Yeah this sounds more flexible. And allows for future or third-party addons that might want to use components from the sys prefix.

py/jupyterlite/src/jupyterlite/config.py
jtpio: The main motivation was to keep the existing `LiteBuildConfig` as the place for configuring the build.

https://github.com/jupyterlite/jupyterlite/pull/621/commits/05eb8aa221582192510310aadcdc12a629490972 takes one step towards moving that logic to the addons.

Maybe we could indeed keep the `ignore_sys_prefix` on `LiteBuildAddon` similar to `disable_addon`. And set the corresponding `ignore_sys_prefix` on each of the addons.

jtpio:
Translations are back on RTD:

image

jtpio:
> The `ignore_sys_prefix` is a list of components to ignore instead of being a `boolean`

Since `ignore_sys_prefix` still supports boolean values with the latest commits, previous configurations should still be compatible after this change.

py/jupyterlite/src/jupyterlite/manager.py
```diff
@@ -68,6 +68,9 @@ def _default_addons(self):
             self.log.debug(f"[lite] [addon] [{name}] load ...")
             try:
                 addon_inst = addon.load()(manager=self)
+                if self._is_sys_prefix_ignored(name):
+                    self.log.debug(f"[lite] [addon] ignore sys prefix for [{name}] ...")
+                    addon_inst.ignore_sys_prefix = True
```
bollwyvl, line 73: hm... it's not really part of the API... if provided to `kwargs`, though, traitlets would just drop unknown kwargs on the floor (for good or ill):

```py
addon_kwargs=dict(manager=self)
if self._is_sys_prefix_ignored(name):
    self.log.debug(f"[lite] [addon] ignore sys prefix for [{name}] ...")
    addon_kwargs.update(ignore_sys_prefix=True)
addon_inst = addon.load()(**addon_kwargs)
```

py/jupyterlite/src/jupyterlite/manager.py
```diff
@@ -131,3 +134,9 @@ def _delayed_gather():
                 yield task
 
         return _delayed_gather
+
+    def _is_sys_prefix_ignored(self, addon):
+        ignore = self.ignore_sys_prefix
+        all_ignored = isinstance(ignore, bool) and ignore == True
```
bollwyvl, line 140: maybe just:

```py
return addon in ignore if isinstance(ignore, tuple) else ignore
```

py/jupyterlite/src/jupyterlite/addons/base.py
```diff
@@ -33,6 +33,11 @@ class BaseAddon(LoggingConfigurable):
 
     manager: LiteManager = Instance(LiteManager)
 
+    ignore_sys_prefix: bool = Bool(
+        False,
+        help="ignore components from sys.prefix",
+    ).tag(config=True)
```
bollwyvl, line 39: hm... i wonder if this should even be configurable... it will get clobbered by the manager.

jtpio: Right we can keep `LiteBuildConfig` as the only place to configure this.

bollwyvl (commented):
LGTM

jtpio:
Thanks @bollwyvl!

Issue #628, opened 2022-05-03
Behavior when downloading files with unsaved edits


HighDiceRoller:
### Problem

When downloading a file that has unsaved edits, the unedited version of the file is downloaded. I've been surprised by this behavior on a few occasions; my natural expectation is that the file would be saved before downloading.

### Proposed Solution

A (non-exhaustive) list of possible alternative behaviors for this case:

1. Automatically save the file before downloading, with the context menu option being named something like "Save and Download".
2. Add separate context menu options for "Download Without Saving" and "Save and Download".
3. After selecting "Download", prompt the user whether they would like to save before downloading.
4. Automatically save the file before downloading if the "Autosave Documents" user setting is checked. (Actually, I'm not sure what this setting currently does?)
5. Add a new user setting to determine whether files are automatically saved before downloading.

If I had to pick one I'd go with option 4, but I welcome any other opinions even if they are for the status quo.

As always, thanks for this phenomenal project!

jtpio:
Thanks @HighDiceRoller.

> Automatically save the file before downloading, with the context menu option being named something like "Save and Download".

Since JupyterLite defines its own commands for downloading the content, probably the context could first be saved with `await context.save()` before triggering the download. 

The relevant code is here:

https://github.com/jupyterlite/jupyterlite/blob/7a7a8a92d98b5b3700ef8ad51c8e2306efcbea23/packages/application-extension/src/index.tsx#L204-L298

jtpio:
@HighDiceRoller would you like to give this a try and open a PR?

My guess would be that we could add `await context.save()` here:

https://github.com/jupyterlite/jupyterlite/blob/7a7a8a92d98b5b3700ef8ad51c8e2306efcbea23/packages/application-extension/src/index.tsx#L263

And here:

https://github.com/jupyterlite/jupyterlite/blob/7a7a8a92d98b5b3700ef8ad51c8e2306efcbea23/packages/application-extension/src/index.tsx#L289

Then we could check the behavior on the ReadTheDocs preview for the PR.

HighDiceRoller:
Gave it a try, see what you think.

PR #629, opened 2022-05-04
Save files before downloading

HighDiceRoller:
## References

Addresses #628 .

## Code changes

Automatically (and unconditionally) saves files before downloading.

## User-facing changes

This applies when the Download option is selected from the File menu or from the File Browser context menu.

None known.

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -287,7 +287,6 @@ const downloadPlugin: JupyterFrontEndPlugin<void> = {
             if (item.type === 'directory') {
               return;
             }
-            await contents.save(item.path);
             const content = await formatContent(item.path);
             downloadContent(content, item.name);
           });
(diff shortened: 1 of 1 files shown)


HighDiceRoller:
Looks like the File menu option works but not the File Browser context menu. I'll investigate further tomorrow.

packages/application-extension/src/index.tsx
```diff
@@ -286,6 +287,7 @@ const downloadPlugin: JupyterFrontEndPlugin<void> = {
             if (item.type === 'directory') {
               return;
             }
+            await contents.save(item.path);
```
jtpio, line 290: This corresponds to the following case:

image

And the UI tests are failing because the second parameter (content) is not specified and hence storing `null`.

jtpio: Maybe we can omit this case for now, and only save when using the docmanager command?

jtpio: That would mean:

```suggestion
```

HighDiceRoller: Ah that's what that means. Sure, let's go for that.

jtpio (approved):
Thanks!

Issue #637, opened 2022-05-11
Might have accidentally edited the official docs?? Really, really confused, sorry.


nathanielvirgo:
I would like to make a simple iPython notebook to share with students in a class, who might not have Python installed.

I found the documents at https://jupyterlite.readthedocs.io/en/latest/_static/retro/tree/ and thought "oh, I didn't know about this project, let's try it out." So I created a new file and put some teaching content in it with the intention of sharing the link with the students.

But now I'm really confused - it seems like somehow the file I created is in the same folder as the official documentation. Did I somehow accidentally edit the official docs? Or does the file only exist on my machine but just appears to be in the repository at readthedocs? If I share the link, will other people be able to access it?

The file I created is at https://jupyterlite.readthedocs.io/en/latest/_static/retro/notebooks/?path=Population%20dynamics.ipynb . Many apologies if I've accidentally vandalised the official docs - feel free to remove the file as I have a local copy.

If it is possible for random noobs to accidentally edit the official docs, I would suggest changing that. Either way, I strongly suggest making the documentation **much** clearer regarding how to make shareable links. At the moment it is not even clear to me whether this is a web-based service that anyone can use, or whether I'm supposed to set up my own web server.

jobovy:
You haven't edited the official docs. The file you've created is in your browser's cache, so it shows up for you, but not for others (I don't see it). So don't worry! While you can technically create a 'shareable link' to your file, it isn't actually shareable, because the notebook doesn't exist for anybody else.

If you want to share your own notebook(s) with other users, you have to build your own deployment with the notebook(s) that you want to share and then you could use the 'shareable link' feature to share links to them. The easiest way to do this is using the demo site repository here: https://github.com/jupyterlite/demo. The README there has detailed instructions on how to use the demo repository to build your own deployment with custom notebooks.

nathanielvirgo:
Thanks, that's very helpful.

I would still suggest some improvements to the documentation are in order, since I wasn't able to discover the page you linked to from the main demo site. Since I got there by searching how to share a notebook online, it was reasonable to think that creating a new file would create a shareable page and not just a local one in my browser's cache - that should be made clear to new users to avoid confusion.

Now I'm trying to follow the instructions you linked to. I guess there are reasons for this being the way it is, but it all seems incredibly weird. I have to make a pull request against the original repository in order to publish something locally on my own account? In any case it's telling me I need approval in order to do that, so no success so far.

nathanielvirgo:
[...]

PR #641, opened 2022-05-12
Restructure the documentation with how-to guides and tutorials

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/393
Closes https://github.com/jupyterlite/jupyterlite/issues/637

Opening this early without too much restructuring yet to get the ball rolling.

This is based on feedback from GitHub issues, Twitter threads and video / IRL conversations, and questions that came up after presentations.

## Code changes

Documentation changes.

Taking some inspiration from TLJH (https://tljh.jupyter.org) and the 4 kinds of documentation (https://documentation.divio.com).

Copying some points raised in https://github.com/jupyterlite/jupyterlite/issues/393#issuecomment-1118292000 here:

- [x]  How to override keyboard shortcuts, for example to use <kbd>Enter</kbd> to execute code in the console
- [x] How do I bundle more packages by default (link to the pyodide docs on adding new packages)
- [x] How to create an extension (link to JupyterLab, mention lite specific parts of the workflow)
- [x] How to create a new kernel
- [x] How to access content from a notebook
- [x] How to create a JupyterLite archive that can be used offline
- [x] How to embed a Python console on a website

We might not need to address them all at once, but it could still be useful to have the sections there even if empty for now, that we could iterate on.

The current "Developer Guide" could be kept almost as-is and renamed to "Reference Guide".

## User-facing changes

This will hopefully make it easier for some folks to deploy JupyterLite websites.

Documentation preview on RTD: https://jupyterlite--641.org.readthedocs.build/en/641/

None


jtpio:
>  How do I bundle more packages by default (link to the pyodide docs on adding new packages)

Given the good progress on https://github.com/jupyterlite/xeus-python-kernel, we might want to just keep that section empty for now until `xeus-python-kernel` is ready to be used.

jtpio:
Stopping here for now, but this can definitely be iterated on.

jtpio:
cc @bollwyvl if you would like to have a look

As mentioned above, this mostly consists in a split of the previous (big) sections into smaller guides, which will hopefully make it more accessible to some folks.

docs/howto/python/packages.md
```diff
@@ -0,0 +1,36 @@
...
+Some Python packages require a frontend extension to also be installed.
+
+See [](../../howto/configure/simple_extensions.md) for more information.
+
+See also how to [ship additional wheels](./wheels.md) at build time.
+```
+
+[micropip]: https://pyodide.org/en/latest/usage/loading-packages.html?#micropip
+
+## Bundling additional packages by default
```
jtpio, line 26: @martinRenou once it is ready for prime time, we can update the documentation to mention how to do this with the xeus python kernel.

Thinking this could go in a `0.2.0` release, also after we have moved the Pyolite kernel to a separate repo (https://github.com/jupyterlite/jupyterlite/issues/386)

docs/howto/index.md
```diff
@@ -0,0 +1,75 @@
+# How-To Guides
```
bollwyvl, line 1: May need either of: 
- a little text that clarifies that this is all during the build, e.g. _you can't do any of this at runtime_ that have access to the configuration and deployment
- another level of headers, and (a place for) some how-to guides for users of an existing site that only have access to the browser
  - getting files out of contents
  - clearing out local versions
  - etc.

jtpio: Good point. Added a note in https://github.com/jupyterlite/jupyterlite/pull/641/commits/95927241d5a76f7614b68a067d6c67b479e52b51.

> another level of headers, and (a place for) some how-to guides for users of an existing site that only have access to the browser 

We can do this in a follow-up PR, since there has been good progress on https://github.com/jupyterlite/jupyterlite/pull/655 and https://github.com/jupyterlite/xeus-python-kernel.

docs/howto/python/packages.md
jtpio: > I don't fully know what that looks like, but it seems like having both python kernels (and their downstream wrappers) being able to rely on some common configuration source would make a lot of sense.

Yeah I'm not sure whether it will be possible to unify the configuration for both kernels here. It's also not clear if or when xeus python and pyodide might converge at some point if they start sharing more common parts.

PR #654, opened 2022-05-28
Fix signature for pyplot.show in Matplotlib

joemarshall:
matplotlib.pyplot.show takes a single optional argument (block).

It is meaningless in jupyterlite, but some other libraries depend on being able to pass it in, e.g. mplfinance, so we should just take it and then ignore it.


jtpio (approved):
Thanks!

Issue #119, opened 2021-06-01
pandas.read_csv("iris.csv") FileNotFound


mklilley:
## Description

Pandas read_csv gave `FileNotFound` error on example file `iris.csv`. I'm hope this is not a silly user error and I'm sorry in advance if it is.

<img width="798" alt="Screenshot 2021-06-01 at 17 02 46" src="https://user-images.githubusercontent.com/7937044/120355480-3d8a8480-c2fb-11eb-900f-2d61f1d3217c.png">

## Reproduce

- Create new notebook in examples folder
- `import pandas as pd`
- `pd.read_csv("iris.csv")`

## Expected behavior

Load and display the content of `iris.csv` on screen

## Context

- Operating System and version:
- MacOS 11.2.3
- MacBook Air (M1, 2020)
- Apple M1 chip
- Browser and version: Brave Version 1.24.84 Chromium: 90.0.4430.93 (Official Build) (arm64)
- JupyterLite version: can't find version number but I did this today (2021/06/01) using JupyterLite for the first time

<details><summary>Browser Output</summary>
<pre>
Dispose worker for kernel 7f928317-a338-41fb-9c00-6c1d041909fd
default.js:1228 Connection lost, reconnecting in 0 seconds.
_reconnect @ default.js:1228
reconnect @ default.js:516
restart @ default.js:487
async function (async)
restart @ default.js:484
restartKernel @ sessioncontext.js:298
restart @ sessioncontext.js:758
async function (async)
restart @ sessioncontext.js:733
onClick @ toolbar.js:343
y.createElement.onMouseDown @ toolbar.js:402
Be @ react-dom.production.min.js:52
Ke @ react-dom.production.min.js:52
(anonymous) @ react-dom.production.min.js:53
xr @ react-dom.production.min.js:100
Cr @ react-dom.production.min.js:101
(anonymous) @ react-dom.production.min.js:113
De @ react-dom.production.min.js:292
(anonymous) @ react-dom.production.min.js:50
Lr @ react-dom.production.min.js:105
Zn @ react-dom.production.min.js:75
Jn @ react-dom.production.min.js:74
n.unstable_runWithPriority @ scheduler.production.min.js:18
Bl @ react-dom.production.min.js:122
Me @ react-dom.production.min.js:292
Xn @ react-dom.production.min.js:73
pyodide.asm.js:9 Python initialization complete
[...]

Issue #653, opened 2022-05-28
File -> download all


joemarshall:
It would be nice to have an easy way to download everything in the files. 

If you're building a starting notebook environment with a bunch of files in it, it is nice to be able to edit in jupyterlite, but it would be even nicer if you could then just save everything quickly into your build folder, rather than having to download each individually.

jtpio:
Thanks @joemarshall for the suggestion.

Would you expect this feature to download every file individually, or everything in an archive? Also wondering how this would work for files in nested folders.

joemarshall:
I think my dream is just a ‘download site archive’ feature, so I can just click file, download all, and I get a zip file of the whole thing including folders.

Joe

manics:
Can https://github.com/jupyterlab-contrib/jupyter-archive be made to work on JupyterLite?

bollwyvl:
> [jupyterlab-contrib/jupyter-archive](https://github.com/jupyterlab-contrib/jupyter-archive) be made to work on JupyterLite?

The UI of that extension could be reused directly.

It would, however, need an implementation of the [server handlers](https://github.com/jupyterlab-contrib/jupyter-archive/blob/master/jupyter_archive/handlers.py), as well as a way to build _useful_ archives, if only a subset of those: `libarchive` has a couple of ports to wasm, ~2mb, and there is some examples in the wild of [wrapping](https://github.com/gt-coar/janki/tree/main/js/jupyterlab-libarchive) in the past... it can handle _reasonably_ sized things, and work with most archives.

Where that would live is more interesting... it doesn't seem like this is a super-useful in-browser capability if you _do_ have a real server.... unless it _also_ knew how to handle `IDrive` stuff (#655). But focusing on light touch: perhaps a `@jupyterlite/lite-jupyter-archive` [extension](https://github.com/jupyterlite/serverlite-cookiecutter-ts) (`pip`-installed as `jupyterlite_archive`) could depend on a specific version ranges of `jupyter-archive` and `jupyterlite`, as the contents api is one of the things most likely to change in the near future.

pierrepo:
My students usually work on university workstations. When the practical is over, that would be very useful if they could just export all their notebooks / files within a single right click "Export current directory as a zip archive" (and then send themselves this archive by e-mail or copy it into a USB thumbdrive).
The cherry on the cake would be the possibility for them to re-import this archive in another instance of the same JupyterLite (at home for example) and extract all files automatically to complete their practical.

Issue #463, opened 2022-02-01
Web worker kernels


SylvainCorlay:
Following discussions with @jtpio and @DerThorsten.

There is a lot of code duplication in the handling of kernels that run in web workers or iframes. For example, in xeus kernels, we duplicated the implementation of IKernel for each xeus-based kernel ([here](https://github.com/jupyterlite/xeus-sqlite-kernel/blob/main/src/xeus_server_kernel.ts), [here](https://github.com/jupyterlite/xeus-lua-kernel/blob/main/src/xeus_server_kernel.ts), and [here](https://github.com/jupyterlite/xeus-wren-kernel/blob/main/src/xeus_server_kernel.ts)). It is also implemented in the pyolite kernel and in a similar way in the JS kernel (but with an IFrame).

I think that:

 - we should have a single implementation of IKernel for communicating with WebWorker-based kernels, as well as IFrame-based kernels.
 - we should probably promote the notion of WebWorker kernels as the preferred way of making secure JupyterLite kernels.

Once this is implemented in core JupyterLite, this will allow us to remove the `xeus_server_kernel.ts` file for each xeus kernel and reduce the amount of code for both the JS and the PyOdide kernel.

bollwyvl:
> making secure 

I think "more stable" or "less likely to wedge your entire UI when they fail," perhaps. 

To actually be discussing "security", the iframe approach actually _could_ be made to use more of the browser's built-in security infrastructure, as one could host kernel iframes on a separate domain. I don't think wrapping a webworker _in_ an iframe on another domain would provide any additional benefits, unless there are some advantages to further limiting the scope of what a user can do.

> single implementation of IKernel

this sounds good, provided we can get it to work with something like [comlink](https://github.com/GoogleChromeLabs/comlink) which supports both windows and webworkers.

bollwyvl:
Here's the "kernel-like" thing i did on `sqlite.js` with comlink:

- [the worker](https://github.com/nrbgt/janki/blob/gh-6-add-cards/js/jupyterlab-sqlite3/src/sqlite.worker.ts)
- [the wrapper](https://github.com/nrbgt/janki/blob/gh-6-add-cards/js/jupyterlab-sqlite3/src/sqlite.ts)

jtpio:
We should indeed consider switching the JS kernel to be backed by a Web Worker instead of an IFrame.

But still mention that extension authors can still have control on this if they want to.

bollwyvl:
[...]

PR #655, opened 2022-05-31
Implement a custom Emscripten File System which communicates with the JupyterLab Content Manager, giving file access to pyolite

martinRenou:
Fixes https://github.com/jupyterlite/jupyterlite/issues/665
Fixes https://github.com/jupyterlite/jupyterlite/issues/119

Progress towards https://github.com/jupyterlite/jupyterlite/issues/315

This PR implements a DriveFS custom file system which allows Emscripten kernels to have access to JupyterLab files. It uses a service worker to make any async task of accessing/deleting/renaming files into a synchronous task.


bollwyvl:
This seems like a fine starting point.

Note that, in case of emscripten-based kernels, we really should try to figure out a way to use the built-in `FS` capability, by mounting to a specific location... there are some links on #315. We gain a _lot_ from transparent read/write access to the same underlying system

bollwyvl:
> other drives

I haven't taken the time to look into other drives for some time... but yes, from the emscripten side, presumably we would need one `IDriveFS` implementation that could wrap any drive, each as another [mount point](https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.mount) into the emscripten in-memory system. 

It's unfortunate drives are such a separate thing: as a litmus test, do they work in RetroLab?

> not limited to IndexedDB 

Strictly speaking, it's _not_ limited to indexeddb, as there are a lot of other localforage implementations that we don't have to implement.

The whole point of #315 is to be as consistent as possible in the UI and implementation. However many backends/frontends we need, they are all accessible through the same API, that work in the different apps (e.g. full lab, retro). So a new kernel only needs to implement against one contents API (if that, as #463 might normalize that, too, by necessity), while a new backend doesn't have to mess with service/web worker boundaries.

martinRenou:
> presumably we would need one IDriveFS implementation that could wrap any drive

In the case of https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access the worker has no access to the files, unlike with indexeddb where any worker can access it. How would/Can an `IDriveFS` (in the emscripten world) work in that case?

Actually I see this JupyterLab drive API as being already the normalized FileSystem API for the JupyterLab world. JupyterLite does not do this at the moment, but it could technically use a custom drive backed by indexeddb.

It would be really nice if there was a way for the worker to mount such a custom JupyterLab drive. But it's unclear to me how this would work, as the worker lives in a different scope.

bollwyvl:
> How would/Can an IDriveFS (in the emscripten world) work in that case?

Right, it's tricky: and why it's not done/barely started yet.

On #463, I had started looking into using [comlink](https://github.com/GoogleChromeLabs/comlink), which does automatic proxies over worker boundaries.  Basically, instead of the API boundary being "this untyped `postMessage` pipe that kinda looks like jupyter kernel messaging" it would be, "this set of messages that can be awaited." Among them would of course be the kernel messages, but also "weird" stuff like interrupts (#460)... and contents messages that can efficiently ship large binary files.

> the JupyterLab world. 

Right: the question of #315 and #463 is: _what other robust worlds already exist_, e.g. emscripten, localforage, such that an implementer isn't starting over from scratch. And indeed, it doesn't seem like contents is an entirely solved problem: perhaps a yjs-on-wasm-FS would become viable, it might make sense to support, as it would be passing less content over the wire.

martinRenou:
Using comlink would indeed be very nice and make the codebase more readable

> and contents messages that can efficiently ship large binary files.

Do you think comlink is more efficient than `postMessage` for large binary files?

> perhaps a yjs-on-wasm-FS would become viable, it might make sense to support, as it would be passing less content over the wire.

Indeed. Though there is still this limitation with emscripten-fs that forces you to "sync" (with `syncfs`) the Python virtual filesystem if you want to open the files with the Python standard library, unless I missed something there?

Should I go ahead with trying to implement an approach that syncs the current drive with the Python virtual filesystem through `postMessage` calls (as proposed in the PR summary)? Although not ideal and not optimized, it could be a first version that works.

bollwyvl:
>  forces you to "sync" (with syncfs) the Python 

My understanding is that a mount point, once created, _is_ the system of record, instead of `MEMFS`, just like a unix mount point. So if you:
- uploaded a file to the file browser and `[*Path.glob("*")]` in the kernel you'd see the file listed
- wrote a file with `Path("foo.txt").write_text("foo")` it would show up, all with accurate modification times, etc.

> comlink is more efficient than postMessage for large binary files?

It's just sugar on postmessage, and one can override whatever they need, it's just moving away from _the huge `case` statement_... of course, it would _stay_ the huge `case` statement for kernel messages, but would probably be better typed. But crucially, we could add other things that _weren't_ kernel messages, and might have different call patterns.

As to performance: it would only actually transfer the file if the contents were requested.

>  go ahead with trying to implement an approach that syncs the current drive

I have been hesitant to do python-kernel-specific stuff at this low of a level: I really feel like solving it once at the webworker/emscripten level is really the right way to go. I tried a bit of `sync` stuff in the `emscripten-filesystem`, and didn't enjoy it very much.

martinRenou:
Thanks for your comments. I'm looking into comlink and emscripten-fs right now, see if I can come up with some kind of wrapper FS around the JupyterLab drives.

martinRenou:
I was able to use comlink for the kernel protocol 🎉 Now let's see if I can setup some drive/emscripten-fs interoperability

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -192,23 +202,20 @@ export class PyoliteKernel extends BaseKernel implements IKernel {
...
    * Handle an complete_request message
    *
    * @param msg The parent message.
    */
-  async completeRequest(
+  completeRequest(
     content: KernelMessage.ICompleteRequestMsg['content']
   ): Promise<KernelMessage.ICompleteReplyMsg['content']> {
-    return await this._sendRequestMessageToWorker('complete-request', content);
+    return this._workerKernel.complete(content, this.parent);
```
bollwyvl, line 218: i personally like leaving the `await` in, even if it's just returning, as it will give more precise error stacks.

martinRenou: Sure I'll add it back, thanks for the review.

I've been wondering if we could not even have the "workerKernel" be the one that implements the `IKernel` interface, because the interfaces become very similar now. This would need some restructure.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -22,6 +22,10 @@ let stderr_stream: any;
 // @ts-ignore: breaks typedoc
 let resolveInputReply: any;
 
+// TODO Import this locally
+declare var Comlink: any;
+importScripts("https://unpkg.com/comlink/dist/umd/comlink.js");
```
bollwyvl, line 27: yeah: `importScripts` has a a couple nast complications, and an unpkg URL is definitely a non-starter.

This has some of the cleanest comlink worker stuff i was able to muster:

https://github.com/nrbgt/janki/tree/gh-6-add-cards/js/jupyterlab-sqlite3

there's some webpack conf needed, but then it can just use `file-loader` (maybe `assets` now?) and and plain old `import`s.

martinRenou: Great thanks, I wasn't sure how to tackle this, I'll have a look at what you did. Webpack is a beast.

martinRenou:
One difficulty with the drive/emscripten-fs interoperability we might face is the fact that our custom emscripten-fs will be async for most (all) tasks. There isn't so much docs about custom emscripten FileSystem implementations and those I found were all synchronous.

martinRenou:
Thanks :) I was actually reading all those stuffs this morning when looking for docs implementing custom file systems.

> as for specific pyodide features

What do you see as being specific to pyodide? It looks to me that if we're able to have a emscripten-fs that can talk to a jupyterlab drive we could reuse the same implementation for all emscripten kernels?

packages/pyolite-kernel/src/worker.ts
```diff
@@ -26,6 +26,11 @@ let resolveInputReply: any;
 declare let Comlink: any;
 importScripts('https://unpkg.com/comlink/dist/umd/comlink.js');
 
+enum Mode {
+  file = 32768,
```
bollwyvl, line 30: it's worth indicating this is a decimal mapping of an octal, e.g. https://github.com/gzuidhof/starboard-python/blob/88327bb56f7e2f673cf1ea4c5ff1e0a68c50e288/src/worker/emscripten-fs.ts#L67

martinRenou:
The code is a mess, but I have an implementation for `readdir` that works 😅 

EMFS

bollwyvl:
Looks like this is going well... though the pain of not being able to use `import` properly is starting to catch up, as I'd think we don't want the drivefs to have to live forever inside the kernel.

Among the goals of #463 (which this PR is already starting, so here we go) would be: wrapper kernels (e.g. of `pyolite`: `pidgylite`, `robotlite`) would be only have to re-implement a small surface area to get new features like this (e.g. just the startup and kernel info) than the current required fork-and-patch.

- normalize the interface of the `IKernelWorker`
  - instead of the generic object
- import the concrete implementation in _another_ file that actually does the wrapping

So my initial play, loosely derived from that `sqlite` was something like this:

- `kernel.ts`
```ts
//  ...
import Comlink from 'comlink';
//  ...
export class PyoliteKernel extends BaseKernel implements IKernel {
  constructor(options: PyoliteKernel.IOptions) {
    super(options);
    this.initWorker(options); // no more string banging!
  }

  async initWorker(): Promise<void> {
    const this._rawWorker = new Worker(new URL('./worker-wrapper.js', import.meta.url), {
      type: 'module',
    });
    const this._remote = Comlink.wrap(this._worker);
    await this._remote.initialize(options);
    this._ready.resolve();
  }
  // ..
  private _worker: Worker;
  private _remote:  Comlink.Remote<IPyoliteKernelWorker>; 
}
```


- `worker-wrapper.ts`
```ts
// worker-comlink-wrapper.ts
import { expose } from 'comlink';
import { PyoliteKernelWorker } from './worker';
const worker = new PyoliteKernelWorker();
expose(worker);
```

- `worker.ts`
```ts
import { DriveFS } from '@jupyterlite/contents`;  // seems pretty legit?

export class PyoliteKernelWorker implements IPyoliteKernelWorker {
  // no constructor, if possible
  async initialize(options: PyoliteKernelWorker.IOptions): Promise<void> {
    this._options = options;
    await this.initRuntime(options);
    await this.initFS(options); // this is new!
    await this.initPackageManager(options);
    await this.initKernel(options);
    await this.initGlobals(options);
  }
}
```

If you like, I can try and take a crack at a sub-PR

martinRenou:
> though the pain of not being able to use `import` properly is starting to catch up, as I'd think we don't want the drivefs to have to live forever inside the kernel.

I totally agree, I was thinking the same thing. We've compiled xeus-python recently and made a JupyterLite kernel for it https://github.com/jupyterlite/xeus-python-kernel, I'd really want it to use the DriveFS too and there is no reason for the implementation to be different.

I'm actually tempted to remove the comlink usage from this PR and move it to a separate branch, as it's kind-of parallel to the DriveFS work. I am not sure we can use comlink for the DriveFS unless comlink can be used with a service worker for making synchronous tasks.

bollwyvl:
> parallel to the DriveFS work

but it _is_ using DriveFS from inside comlink worker at this time? I guess I haven't traced through which layers are doing what.

bollwyvl:
Created https://github.com/martinRenou/jupyterlite/pull/1: it breaks a lot, but might have some salveageable ideas.

I'll be (all but) offline over the weekend, but happy to keep tinkering with this: i'd really rather see us get something many folks can build on and use... but am semi-concerned about some of the browser limitations.

martinRenou:
Thanks! I'll also be offline over the weekend and on Monday, will look at it as soon as I'm back.

martinRenou:
From @bollwyvl in https://github.com/martinRenou/jupyterlite/pull/1:

> - based on https://github.com/jupyterlite/jupyterlite/pull/655
> - inching towards https://github.com/jupyterlite/jupyterlite/issues/463
> - splits up more of the comlink, FS, etc. stuff
> - removes use of buildScripts to create worker
> - removes lots of declares
> - adds (some) more typing
> - very WIP, many things broken
>   - os.listdir()
>   - drive in firefox
>   - doesn't even load in firefox private browsing (navigator.serviceworker undefined)

I'll try to find why `os.listdir()` does not work anymore.

I've been testing it on Chromium only for now, will try later on Firefox as well. Concerning the Firefox private browsing issue, this seems to be a known issue https://bugzilla.mozilla.org/show_bug.cgi?id=1320796, we probably want to give some meaningful feedback to the user in that case.

martinRenou:
My latest commit doesn't work well for now. It requires this fix from pyodide https://github.com/pyodide/pyodide/pull/2582 to be included.

bollwyvl:
> we probably want to give some meaningful feedback to the user in that case.

i feel like, beyond feedback, it will _have_ to be some kind of graceful degredation, not a hard fail telling you to use your browser a different way. Likely, this is wholly separate kernels, with the service-worker-powered one a subclass of the more limited ephemeral one.

At some point in the not-too-distant future, emscripten's `WasmFS` will give us the right tools for the job, even without a service worker.

martinRenou:
Making some progress, I have support for:
- creating directories
- listing files/directories
- renaming files
- removing directories

filesystem

There is a permission issue when trying to access files though, I need to investigate it.

Also there is an issue with the service worker, it seems to only work for the tab that installed the service worker, after that it does not intercept requests. I suspect it needs to claim new tabs when they are open.

martinRenou:
We can now read/write files! I took the https://github.com/gzuidhof/starboard-notebook repo as an inspiration for the implementation.

There are a couple of things I would like to fix/improve/check still, after that I think the PR is ready:
- [ ] Reading a file content adds quotes around the string value while it shouldn't
- [ ] The service worker works the first time it is installed, but refreshing the page breaks the communication (maybe the service worker does not claim new tabs, though it should)
- [ ] Check reading/writing binary data
- [ ] Move the "server" part (the one responsible for using the content manager) to its own plugin
- [ ] Remove the IContent exposition to the IKernel I added in the first commit, we don't need it
- [ ] Make use of the kernel "location" passed to the kernel and have the kernel do a `cd` to that location after mounting the drive

jtpio:
Nice progress!

>  Reading a file content adds quotes around the string value while it shouldn't

Random guess: maybe related to a `JSON.stringify` somewhere?

martinRenou:
Probably! I'll have a try, thanks for the hint

martinRenou:
Thanks, it was indeed the `JSON.stringify` that was adding the extra quote

martinRenou:
@jtpio @bollwyvl I rebased and marked the PR as ready for review. If you agree we can tackle the two remaining points I added above in separate PRs (and open issues to track that).

Here is a picture of loading an image I dragged/dropped in jupyterlite and loaded from a Notebook that is in a subdirectory:
file_read

packages/server-extension/src/index.ts
```diff
@@ -187,6 +193,177 @@ const contentsRoutesPlugin: JupyterLiteServerPlugin<void> = {
   },
 };
 
+/**
+ * A plugin installing the service worker for and handles communication with the Emscpriten file system.
+ */
+const serviceWorkerPlugin: JupyterLiteServerPlugin<void> = {
+  id: '@jupyterlite/server-extension:service-worker',
+  autoStart: true,
+  requires: [],
```
jtpio, line 202: Looks like this could be removed if it's empty:

```suggestion
```

packages/contents/src/drivefs.ts
```diff
@@ -0,0 +1,517 @@
...
+}
+
+export class DriveFSEmscriptenStreamOps implements IEmscriptenStreamOps {
+  private fs: DriveFS;
+
+  constructor(fs: DriveFS) {
+    this.fs = fs;
+  }
+
+  public open(stream: IEmscriptenStream): void {
```
jtpio, line 121: Maybe we could drop the `public` here (and below) for consistency with the rest of the code base:

```suggestion
  open(stream: IEmscriptenStream): void {
```

packages/server-extension/src/index.ts
```diff
@@ -187,6 +193,177 @@ const contentsRoutesPlugin: JupyterLiteServerPlugin<void> = {
...
+  activate: (app: JupyterLiteServer) => {
+    navigator.serviceWorker
```
jtpio, line 204: Probably we could handle `navigator.serviceWorker` not being available, which seems to be the case in a Firefox private window.

So the plugin does not fail at startup.

packages/server-extension/src/index.ts
jtpio: Also having the service worker logic in its own plugin sounds like a good idea :+1: 

Since it doesn't depend on other services and doesn't provide a token, it could easily be omitted or disabled (which could be useful to https://github.com/executablebooks/thebe-core/pull/3).

packages/contents/src/drivefs.ts
```diff
@@ -0,0 +1,517 @@
+export const DIR_MODE = 16895; // 040777
```
bollwyvl, line 1: a lot of this looks copied: probably needs some headers with URLs and license info from the sources.

packages/contents/src/drivefs.ts
```diff
@@ -0,0 +1,517 @@
...
+    throw new this.fs.FS.ErrnoError(this.fs.ERRNO_CODES['EPERM']);
+  }
+
+  public readlink(node: IEmscriptenFSNode): string {
+    throw new this.fs.FS.ErrnoError(this.fs.ERRNO_CODES['EPERM']);
+  }
+}
+
+/**
+ * Wrap serviceworker requests for an Emscripten-compatible syncronous API.
```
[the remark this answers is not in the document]
bollwyvl: can go either way: i almost feel like using the code-style `ServiceWorker` is more appropriate.

packages/pyolite-kernel/src/declarations.d.ts
```diff
@@ -7,10 +7,3 @@ declare module '!!file-loader*' {
   const res: string;
   return res;
 }
-
```
bollwyvl, line 10: these deletions feel great... i wonder if this whole file can be deleted now?

packages/pyolite-kernel/src/worker.ts
```diff
@@ -1,420 +1,575 @@
...
+  /**
+   * Accept the URLs from the host
+   **/
+  async initialize(options: IPyoliteWorkerKernel.IOptions): Promise<void> {
+    this._options = options;
+    await this.initRuntime(options);
+    await this.initPackageManager(options);
+    await this.initKernel(options);
+    await this.initGlobals(options);
+    await this.initFilesystem(options);
```
bollwyvl, line 150: Hm... I know i moved this down because i was breaking other stuff, but I wonder if we can move `initFilesystem` up to somwhere before `initKernel`: this would enable a _not_ not-invented-here way to do startup code for kernels, e.g. ipython profiles.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -1,420 +1,575 @@
...
+      PATH: this._pyodide._module.PATH,
+      ERRNO_CODES: ERRNO_CODES,
+      baseUrl,
     });
-  };
+    FS.mkdir('/drive');
+    FS.mount(driveFS, {}, '/drive');
+    FS.chdir('/drive');
+    this._driveFS = driveFS;
+    console.log('changing dir to ', options.location);
```
bollwyvl, line 221: would remove

packages/pyolite-kernel/src/declarations.d.ts
bollwyvl: we do still need it, for now, for the file-loader stuff... but `raw` can go... as can `buildScripts`... having a look...

bollwyvl:
Hm... been experimenting locally and on RTD on the other PR... on Linux with `Chromium 102.0.5005.61` it doesn't really seem like any of the drivefs stuff works in private or standard browsing... and firefox has some other issues.

martinRenou:
Thank you both for the reviews! I'll resolve the comments.

> Hm... been experimenting locally and on RTD on the other PR... on Linux with Chromium 102.0.5005.61 it doesn't really seem like any of the drivefs stuff works in private or standard browsing... and firefox has some other issues.

Indeed it fails to `chdir` to the kernel location at startup. I'm looking into it right now.

packages/server-extension/src/index.ts
```diff
@@ -187,6 +193,177 @@ const contentsRoutesPlugin: JupyterLiteServerPlugin<void> = {
...
+          );
+        },
+        (err) => {
+          // registration failed :(
+          console.log('ServiceWorker registration failed: ', err);
+        }
+      );
+
+    // Setup communication with service worker for the virtual fs
+    const broadcast = new BroadcastChannel('/api/drive.v1');
```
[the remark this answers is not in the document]
martinRenou: I agree, but it might be difficult for the `/services.js` to import such a string from another place?

martinRenou:
It should now be more robust when it comes to hard refreshing the page :)

Please give it a try on RTD! Also I made a proper ServiceWorker plugin as you suggested, and the pyolite kernel only mounts the custom drive if the service worker registration succeeded.

packages/contents/src/drivefs.ts
```diff
@@ -0,0 +1,517 @@
...
+ */
+export class ContentsAPI {
+  constructor(baseUrl: string, FS: any, ERRNO_CODES: any) {
```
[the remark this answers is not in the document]
martinRenou: I'll guess we'll want to contribute those types to `@types/emscripten` actually, it doesn't have the types we need yet https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/emscripten/index.d.ts

bollwyvl:
Getting much further! It is impressive

Observations from running RTD: 
- `open` and `read`ing text and binary looks good... 
  - reading a notebook always returns `[object Object]`
    - Screenshot from 2022-06-10 09-08-03
- when opening a second browser tab with `?room=` in the same (browser, incognito), the python kernel disappears from the launcher, without any console messages, and can't be selected from the kernel change menu
  - Screenshot from 2022-06-10 09-12-21
    - subsequent reloads of the browser also missing the kernel, no kernel message
    - after dev tools -> application -> storage -> clear site data, python kernel appears again
- if a file with spaces in the name is drag-and-dropped into the file browser, the parent directory (but not subdirectories) starts to appear empty to python
  - Screenshot from 2022-06-10 09-26-12
  -  it does fire some SW requests for _every_ file, however
  - removing the space from the filename makes things work again

martinRenou:
One nice side effect of the virtual file system implementation: completion on file names!
file_completion

martinRenou:
Thanks for the previous testing @bollwyvl 

I pushed some fixes for:
- issue when file names contain spaces
- issue when reading JSON files
- issue with the Python kernel not showing up

martinRenou:
In combination with #666 which fixes a CSV upload issue we can open CSV files with pandas:

csv

jtpio:
This is looking good, thanks Martin!!

Do you think we could also update the documentation here: https://jupyterlite.readthedocs.io/en/latest/howto/content/python.html
It would also be fine to do that as a follow-up.

app/services.js
```diff
@@ -0,0 +1,58 @@
+const broadcast = new BroadcastChannel('/api/drive.v1');
+
+self.addEventListener('install', (event) => {
+  console.log('Install JupyterLite service v1');
```
jtpio, line 4: Should we keep these `console.log()` here? (some site deployers want quiet apps)

packages/server-extension/src/index.ts
```diff
@@ -247,7 +423,7 @@ const kernelSpecRoutesPlugin: JupyterLiteServerPlugin<void> = {
   id: '@jupyterlite/server-extension:kernelspec-routes',
   autoStart: true,
   requires: [IKernelSpecs],
-  activate: (app: JupyterLiteServer, kernelspecs: IKernelSpecs) => {
+  activate: async (app: JupyterLiteServer, kernelspecs: IKernelSpecs) => {
```
jtpio, line 426: The diff shows the addition of `async`, but it looks like this is not needed?

jtpio (approved):
Thanks @martinRenou and @bollwyvl!!

This is solving a big usability issue and working well on the RTD preview :+1: 

We can still iterate in follow-up PRs and issues.

jtpio:
cc @bollwyvl if you want to take another look at it.

Happy to make a new release with this when it lands.

martinRenou:
I'd like to push a last commit so that it respects custom `drive`s

psychemedia:
This PR looks really exciting, not least because it seems to work with `pandas` csv file handling for local files at least.

I note that sqlite doesn't play nicely though:

```
import sqlite3

conn = sqlite3.connect('test.db')

conn.execute('''CREATE TABLE TEST
         (ID INT PRIMARY KEY     NOT NULL,
         COL           TEXT    NOT NULL);''')

conn.close()

#---
"""
---------------------------------------------------------------------------
OperationalError                          Traceback (most recent call last)
Input In [5], in <cell line: 3>()
      1 conn = sqlite3.connect('test.db')
----> 3 conn.execute('''CREATE TABLE TEST
      4          (ID INT PRIMARY KEY     NOT NULL,
      5          COL           TEXT    NOT NULL);''')
      7 conn.close()

OperationalError: disk I/O error
""'
```

martinRenou:
Thanks for testing it out! I'll have a look at this issue.

bollwyvl:
> respects custom `drive`s

This seems reasonable... is there a good one to add to the demo site? of the ones i know are maybe too complicated/opinionated:
- `jupyterlab-github` would be great (though proprietary), but the read-only UI acts kinda broken (even though it's working)
- `jupyterlab-filesystem-access` is maybe _not_ a good baseline, because of the only-works-in-chrome thing... 
- `jupyterlab-s3-browser` might always need a token
- `jupyterlab-hdf5` has a rather complicated extension

All of this makes me wonder if getting a working `libarchive` backend working would make sense for more than just #653: akin to `hdf5`, it could use the default browser to fetch the archive and then spawn a new idrive from that...

martinRenou:
I pushed another commit for drive support:

jupyterlab-github:
file_github

jupyterlab-filesystem-access:
file_local

packages/contents/src/drivefs.ts
```diff
@@ -384,29 +390,53 @@ export class ContentsAPI {
...
 
+  /**
+   * Normalize a Path by making it compliant for the content manager
+   *
+   * @param path: the path relatively to the Emscripten drive
+   */
+  normalizePath(path: string): string {
+    // Remove drive prefix
+    // The following check should always be true but for safety we check
+    if (path.startsWith('/drive')) {
```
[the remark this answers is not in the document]
martinRenou: Yeah definitely 👍🏽

packages/contents/src/drivefs.ts
```diff
@@ -384,29 +390,53 @@ export class ContentsAPI {
...
+      path = path.slice('/drive'.length);
+    }
+
+    if (this._driveName) {
+      path = `${this._driveName}:${path}`;
```
bollwyvl, line 432: perhaps hoisting the separator, and making it multiple characters, e.g. `DRIVE_SEP = '::'`

martinRenou:
> This seems reasonable... is there a good one to add to the demo site? of the ones i know are maybe too complicated/opinionated

I personally think the jupyterlab-filesystem-access one would be a good candidate, but indeed only works with Chrome. Which is no big deal, Firefox users will just not see it.

packages/contents/src/drivefs.ts
```diff
@@ -0,0 +1,551 @@
...
+  }
+
+  llseek(stream: IEmscriptenStream, offset: number, whence: number): number {
+    let position = offset;
+    if (whence === SEEK_CUR) {
+      position += stream.position;
+    } else if (whence === SEEK_END) {
+      if (this.fs.FS.isFile(stream.node.mode)) {
+        try {
+          position += stream.file!.data.length;
```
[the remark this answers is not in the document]
martinRenou: Sorry I don't understand what you are saying, can you elaborate on this?

psychemedia:
Good support for filesystem access would be much appreciated in edu settings; `jupyerlite` offers the promise of install free, which is a big plus, and being able to support local file system access in a simple way also provides a route to file persistence by individuals.

*(Note to self: check local filesystem access is fine with USB mounted directories.)*

martinRenou:
Thanks for the review again @bollwyvl. I resolved most of your comments, I have not understood one of them though.

If you don't mind we can try to tackle the issue @psychemedia reported in a follow-up PR.

psychemedia:
If I download a python file into memory and save it as a `.txt` file, I can click to open it as expected from the JupyterLab file browser.

If I save it as a `.py` file and then click to open it from the file browser, I see base64 encoded data.

```python
import pyodide

# Load in the contents of `ouseful_jupyterlite_utils/utils` package from a remote URL:

URL = "https://raw.githubusercontent.com/innovationOUtside/ouseful_jupyterlite_utils/main/ouseful_jupyterlite_utils/utils.py"
txt = pyodide.open_url(URL).read()

with open("test.txt", "w") as f:
    f.write(txt)
    f.close()
# Click to open file displays text as expected

with open("test.py", "w") as f:
    f.write(txt)
    f.close()
# Click to open file displays base64 encoded data
```

packages/contents/src/drivefs.ts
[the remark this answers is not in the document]
bollwyvl: yeah: this is the single warning that shows up in eslint the command line and in the review ui. 


Screenshot from 2022-06-14 20-06-01


i'm thinking we do one of:
- fail if there's no file (this function already throws a lot) 
- console.error as continue happily
- ignore the linter rule (would prefer not)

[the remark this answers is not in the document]
martinRenou: Thanks :) I went for the first proposition

jtpio:
Thanks all!

Let's move forward and get this in so we can iterate and improve things in follow-up issues.

jtpio:
Let's try to get https://github.com/jupyterlite/jupyterlite/pull/670 as well before making a new release.

martinRenou:
You made my day merging this, thanks!!

PR #657, opened 2022-06-02
Fix the kernelspecs response

jtpio:
## References

Follow the kernelspecs response as in Jupyter Server, which looks like the following:

```json
{
    "default": "python3",
    "kernelspecs":
    {
        "python3":
        {
            "name": "python3",
            "spec":
            {
                "argv":
                [
                    "PREFIX/bin/python",
                    "-m",
                    "ipykernel_launcher",
                    "-f",
                    "{connection_file}"
                ],
                "env":
                {},
                "display_name": "Python 3 (ipykernel)",
                "language": "python",
                "interrupt_mode": "signal",
                "metadata":
                {
                    "debugger": true
                }
            },
            "resources":
            {
                "logo-64x64": "/kernelspecs/python3/logo-64x64.png",
                "logo-32x32": "/kernelspecs/python3/logo-32x32.png"
            }
        }
    }
}
```

This helps address the issue reported in https://github.com/jupyterlite/echo-kernel/issues/5

## Code changes

Change the reponse of the `/api/kernelspecs` endpoint.

## User-facing changes

None

None, the response is still compatible with the existing kernels.

### DIFF
--- a/packages/server-extension/src/index.ts
+++ b/packages/server-extension/src/index.ts
@@ -254,18 +254,18 @@ const kernelSpecRoutesPlugin: JupyterLiteServerPlugin<void> = {
         return new Response(null);
       }
       // follow the same format as in Jupyter Server
-      const kernels: {
+      const allKernelSpecs: {
         [name: string]: {
           name: string;
-          spec: KernelSpec.ISpecModel;
-          resources: { [name: string]: string };
+          spec: KernelSpec.ISpecModel | undefined;
+          resources: { [name: string]: string } | undefined;
         };
       } = {};
       const allSpecs = specs.kernelspecs;
       Object.keys(allSpecs).forEach((name) => {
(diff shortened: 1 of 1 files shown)


jtpio:
This seems to be behaving correctly with https://github.com/jupyterlite/echo-kernel/pull/6 built and installed locally:

image

jtpio:
Another source of confusion is the top-level `resources` entry, not specified in https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs, but in the JupyterLab `KernelSpec.ISpecModel`:

https://github.com/jupyterlab/jupyterlab/blob/edabd4e6812a53cc325b80e224dfddf68cc29d70/packages/services/src/kernelspec/restapi.ts#L74

jtpio:
OK getting that one in since it will help with https://github.com/jupyterlite/echo-kernel/issues/5

PR #660, opened 2022-06-08
Simplify kernelspecs

jtpio:
## References

Follow-up to #657 

## Code changes

Remove redundant information when registering the JS and Python kernels.

## User-facing changes

None

None


jtpio:
Looking good on RTD:

image

Issue #665, opened 2022-06-11
Opening csv files


annawawrzynczak:
Opening a csv file with pandas doesn't work. I uploeaded the file in the same folder as the notebook yet still it cannot be found

jtpio:
Thanks @annawawrzynczak for reporting.

This will likely be fixed by https://github.com/jupyterlite/jupyterlite/pull/655

martinRenou:
JupyterLite seems to fail at uploading CSV files as well. The uploaded file is a base64 representation of the file.

![base64](https://user-images.githubusercontent.com/21197331/173299605-7a184b36-f89c-4e24-922f-a2021b9455ab.png)

martinRenou:
#666 and #655 will respectively fix the upload issue and allow the kernel to load the file

![csv](https://user-images.githubusercontent.com/21197331/173301875-4d2685b2-9726-426d-afc4-175bb59b5b1d.png)

bollwyvl:
See #91 for the long running discussion about this. In short: today it's
easier to download files from elsewhere on the Internet then inside the
same browser session.

You can also use ipywidgets:
https://ipywidgets.readthedocs.io/en/7.7.0/examples/Widget%20List.html#File-Upload

Longer term: #655 is the current effort towards unifying contents and the
kernel file system: we're still determining what the limits are of this
approach, but you can already try it out on the ReadTheDocs preview there.
The end goal is for all emscripten-based kernels to have near-transparent
read/write access to a consistent set of files.

PR #666, opened 2022-06-13
Fix CSV file upload

martinRenou:
## References

A partial fix for https://github.com/jupyterlite/jupyterlite/issues/665. This PR fixes the upload of CSV file by making it a known text type.

In combination with https://github.com/jupyterlite/jupyterlite/pull/655 we can open the CSV file with pandas:

csv

## Code changes

Make the CSV mimetype a known text type

## User-facing changes

They can now upload CSV files

None

### DIFF
--- a/packages/contents/src/tokens.ts
+++ b/packages/contents/src/tokens.ts
@@ -129,10 +129,14 @@ export interface IContents {
  * Commonly-used mimetypes
  */
 export namespace MIME {
-  export const JS = 'application/javascript';
   export const JSON = 'application/json';
   export const MANIFEST_JSON = 'application/manifest+json';
   export const PLAIN_TEXT = 'text/plain';
+  export const CSV = 'text/csv';
+  export const CALENDAR = 'text/calendar';
+  export const CSS = 'text/css';
+  export const HTML = 'text/html';
+  export const JS = 'text/javascript';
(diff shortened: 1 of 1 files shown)


packages/contents/src/tokens.ts
```diff
@@ -129,10 +129,14 @@ export interface IContents {
  * Commonly-used mimetypes
  */
 export namespace MIME {
-  export const JS = 'application/javascript';
```
martinRenou, line 132: This mimetype is deprecated, see "Legacy JavaScript mimetypes" in https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types. `text/javascript` replaces it

jtpio: Maybe we still need to keep it around for now?

Uploading and opening a `.js` file seems to be giving the following with this PR:

image

While the file is correctly handled on `main`: https://jupyterlite.readthedocs.io/en/latest/_static/lab/

image

martinRenou: Thanks for the review, I reverted the change then.

jtpio (approved):
Thanks!

bollwyvl:
While it's fine to fix certain, common cases for which we ship default renderers, there's no way we can get them all. Will issue up...

PR #668, opened 2022-06-14
Add configurable (extra)FileTypes

### LINKED ISSUE #667: More configurable mime types for serving, contents
### Problem

At present we hard-code an understanding of some known MIME types. This creates problems when:
- #569 serving files in development where the os/python/browser are misconfigured by a third-party tool
- #665 uploading files where user files which a renderer can show, but the lite contents manager misidentifes

### Proposed Solution

Not mutually exclusive, but roughly in order of impact:

- add some build-time `file_types` configuration options
  - probably reuse [IFileType](https://github.com/jupyterlab/jupyterlab/blob/v3.4.3/packages/rendermime-interfaces/src/index.ts#L167)
  - use this extended knowledge
    - in the browser contents manager
    - in the server, use these to overload the served types (#569)
- do more robust in-browser guessing of well-known types
  - extension-based with [`mime-types`](https://github.com/jshttp/mime-types)
  - extension- and content-based [`@npcz/magic`](https://github.com/npcz/magic)... but doesn't work in browser yet 
- cheat and use the registered file types from the [document registry](https://github.com/jupyterlab/jupyterlab/blob/v3.4.3/packages/docregistry/src/registry.ts#L292)
  - in the browser
  - this is not available at build time, because those JS values aren't available
- add user settings for registering file types
 - this would let someone overcome any build time problems if they _really_ knew what they needed
 - would work well with the above, probably

### Additional context

- might address #534
  - already fixed by #655?
- continues work from #541, #666
- sensible defaults should include e.g #570

bollwyvl:
## References

- fixes #667
- partially includes #570

## Code changes

- adds `(extraFile|file)Types`, using a subset of the `IRenderMime.IFileType` to `jupyter-lite.json`
- adds to `jupyter_lite_config.json`
- removes most hard-coded mimetypes from client-side contents manager
- uses mime types in server as well

## User-facing changes

- a larger collection of uploaded files will work out-of-the-box
- site owners can customize (or entirely replace) these 

- a number of constants are no longer hoisted


bollwyvl:
> content may not correctly be parsed on file upload:

thanks for the review, have busted this down to draft and will do some more looking...

bollwyvl:
aside from some dumb mistakes, I think we might need some of the other fixes from #655... i'll try out a sub-PR that merge the two.

bollwyvl:
I'm going to close this one out in favor of #670: we can decide when it would be clever to merge, but I'll try to keep this up to date with #655.

Issue #664, opened 2022-06-10
Local development without `conda`


jtpio:
### Problem

Currently we depend on the `conda list --explicit` command here:

This is hit when running `doit` locally to set up the environment.

For those using `micromamba` only, this will give the following error:

```
PythonAction Error
Traceback (most recent call last):
  File "/home/gitpod/micromamba/envs/jupyterlite-dev/lib/python3.9/site-packages/doit/action.py", line 437, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/workspace/jupyterlite/dodo.py", line 1263, in sync_lite_config
    raw_lock = subprocess.check_output([which("conda"), "list", "--explicit"])
  File "/workspace/jupyterlite/dodo.py", line 21, in which
    Path(
  File "/home/gitpod/micromamba/envs/jupyterlite-dev/lib/python3.9/pathlib.py", line 1082, in __new__
    self = cls._from_parts(args, init=False)
  File "/home/gitpod/micromamba/envs/jupyterlite-dev/lib/python3.9/pathlib.py", line 707, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/home/gitpod/micromamba/envs/jupyterlite-dev/lib/python3.9/pathlib.py", line 691, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

### Proposed Solution

`micromamba` does not seem to support the `--explicit` flag for now.

Maybe there is a way to make that task work without it, or allow for `conda` not being installed.

### Additional context

cc @martinRenou who might be interested as well

bollwyvl:
Looks like this might work?

```bash
micromamba env export --explicit
```

maybe offer a `USE_MICROMAMBA` flag or something? Or just a `EXPLICIT_ENV_CMD`?

c/f https://github.com/mamba-org/mamba/issues/1209#issue-1023768187

another option would be to embrace `conda-lock`: it's gotten rather heavy (brings in `poetry`), and nothing else speaks its new lockfile format, but it does offer some nice ways to avoid our env-copy-pasta stuff, and cross-platform stuff. It might then be reasonable to actually check in lockfiles, which would would only be updated on request.

PR #669, opened 2022-06-14
Allow for using `micromamba` for the dev setup

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/664

## Code changes

For now just trying to use `micromamba` if available, defaulting to `conda` otherwise.

## User-facing changes

None

None


jtpio:
@bollwyvl this seems to be doing the trick locally.

We could also indeed use some sort of flag as mentioned in https://github.com/jupyterlite/jupyterlite/issues/664#issuecomment-1152553892.

martinRenou (commented):
LGTM! Thanks for doing this!

jtpio:
Getting this one in since it fixes the local dev install with micromamba, and CI is still passing.

Happy to switch to a different approach as a follow-up if needed.

PR #670, opened 2022-06-15
Add configurable fileTypes for upload and HTTP serving

### LINKED ISSUE #667: More configurable mime types for serving, contents

bollwyvl:
## References

- fixes #667
- partially includes 
  - #570
  - #668
  - #655

## Code changes

- [x] adds `fileTypes`, using a subset of the `IRenderMime.IFileType` to `jupyter-lite.json`
- [x] adds to `jupyter_lite_config.json` as `(extra_)file_types` which get merged together
- [x] removes most hard-coded mimetypes from client-side contents manager
- [x] uses configured mime types in server as well
- [x] defaults uploaded files to `file`/`base64`  instead of `json`/`notebook`

## User-facing changes

- a larger collection of uploaded files will work out-of-the-box
- site owners can customize (or entirely replace) these 

- a number of constants are no longer hoisted


jtpio:
The docs failed to build on the latest `main`, which looks related to the new release of `myst-nb` on conda mreged yesterday: https://github.com/conda-forge/myst-nb-feedstock/pull/18

https://github.com/jupyterlite/jupyterlite/pull/670/commits/5d6e5223289f1af7cb950cf99b0ea335b15a76d5 seems to be fixing it.

bollwyvl:
Thanks for the assist. When last I put it down, most of the arbitrary file types i tried were working for various upload/download scenarios. PDF was still bothersome.

jtpio:
Yes I gave it a spin on RTD and it looked good. Let's get it in.

jtpio:
> a larger collection of uploaded files will work out-of-the-box

This fixes the fasta examples which was failing to open until now:

image

bollwyvl:
>  fixes the fasta example 

right, but the more telling thing is:
- download the fasta
- re-upload the fasta

I was specifically seeing SVG come out as `json`, as its `Untitled` file was starting life as a notebook. I hope to test this out once more, but am pretty slammed these days and might not be able to fix anything i find. 

Hate to hold things up, but getting reports like _this doesn't work for my niche file format_ are really boring when the feature is otherwise awesome, and we have a plan for letting owners fix arbitrary formats.

bollwyvl:
i also haven't tested _any_ of the #655 stuff vs the webrtc backend. if it breaks that, we need to stop and re-think whether we need to be able to put all of the serviceworker stuff behind a build flag.

bollwyvl:
hm... looks like this is breaking config somehow, e.g. not propagating `collaborative`...

jtpio:
> vs the webrtc backend. if it breaks that, we need to stop and re-think whether we need to be able to put all of the serviceworker stuff behind a build flag.

I tried with the RTD preview for this PR, and editing seems to be working as before:

image


New files and directories created from the kernel are not replicated across peers, but that was also the case before when creating new files from the UI.

Although it's not clear yet how this will work, there will likely be substantial changes needed when updating to JupyterLab 4.0 packages with the RTC logic progressively moving to the server.

bollwyvl:
>  files and directories created from the kernel are not replicated across peers

i tried this out too... it looks like they _will_ work, once you _open_ them in both browsers, and somebody decides to accept the _Revert_ option. That's good enough for me!

No further changes planned (other than cleaning up the title of the PR).

martinRenou (commented):
LGTM! Thanks for doing this

jtpio (approved):
Thanks!

PR #672, opened 2022-06-16
Add Gitpod configuration

jtpio:
## References

This should make it easier for folks to start contributing to JupyterLite, especially when working on the docs and not wanting to install node and all the tooling.

The dev setup can sometimes be challenging with the combination of JS / Node and Python.

## Code changes

Add a Gitpod config file to:

- [x] Automatically setup a `micromamba` environment on `init`
- [x] Serve the assets with `doit serve`
- [x] Watch for changes with `doit watch`
- [x] Serve and watch the docs

## User-facing changes

None for end users.

Contributors should be able to work on the JupyterLite repo from a web browser directly:

image

None

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -24,6 +24,7 @@ tasks:
   - name: watch
     command: |
       gp sync-await setup
+      source ~/.bashrc
       micromamba activate
       doit watch
 ports:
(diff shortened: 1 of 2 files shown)


jtpio:
Looking good from that branch:

image

Will get that one in to be able to use it in practice.

PR #680, opened 2022-06-20
Add encoding to LiteStream

eagleoflqj:
## References

https://github.com/sympy/live/issues/10

## Code changes

`LiteStream` should have a field `encoding` to mock `sys.stdout`.
Since it's just a placeholder, I'm not sure whether it should be put as a class field, as `self.encoding`, or as a read-only `property`.

## User-facing changes

User code that access `sys.stdout.encoding` won't throw.


martinRenou (approved):
LGTM thanks!

PR #685, opened 2022-06-21
Change the Pyolite display name

martinRenou:
## References

Going in the direction of https://github.com/jupyterlite/jupyterlite/issues/116

> Another option would be to use the "Python (Pyodide)" display name, similar to "Python (ipykernel)" and "Python (XPython)" in regular Jupyter.

Similar PR in xeus-python: https://github.com/jupyterlite/xeus-python-kernel/pull/38

## Code changes

Change the Pyolite display name to `Python (Pyodide)`


jtpio:
Sounds good thanks :+1: 

Looks like some reference snapshots will need to be updated since the display name has changed.

jtpio (approved):
Thanks!

PR #686, opened 2022-06-22
Caching service worker

martinRenou:
## References

Port changes from https://github.com/jupyterlite/jupyterlite/pull/137 into the new service worker

Closes https://github.com/jupyterlite/jupyterlite/pull/137

## Code changes

Add caching logic to the service worker, for offline support.


app/lab/manifest.webmanifest
```diff
@@ -0,0 +1,21 @@
+{
+    "short_name": "JupyterLite",
+    "name": "JupyterLite",
+    "description": "WASM powered Jupyter running in the browser",
+    "icons": [
+      {
+        "src": "../icon-120x120.png",
```
jtpio, line 7: Probably we could also use the yellow icon instead of the orange so it's easier to recognize?

https://github.com/jupyterlite/jupyterlite/blob/main/docs/_static/icon.svg

martinRenou:
For some reason my browser is not able to find the manifest on my PR

jtpio:
Showing up nicely with Chromium 103: 

image

martinRenou:
Once installed it shows up nicely in the Gnome UI:

https://user-images.githubusercontent.com/21197331/175316824-52f6a272-5c87-4678-8191-1a0dd3b0c59a.mp4

bollwyvl:
Are the multiple manifests, etc. necessary? Could we do it with a single root-level one some kind of preference for which UI? [shortcuts](https://developer.mozilla.org/en-US/docs/Web/Manifest/shortcuts)?

If it does need to be that way: is `repl` going to get one as well? It seems like a pretty big win for certain situations, though I don't know how it would play vs being embedded within a larger application: frankly seems like a pretty useful function for a sphinx theme to have, anyway.

martinRenou:
> Could we do it with a single root-level one some kind of preference for which UI? [shortcuts](https://developer.mozilla.org/en-US/docs/Web/Manifest/shortcuts)?

Oh I didn't know about shortcuts, sweet! I'll try that.

martinRenou:
I was thinking maybe the retro one doesn't make sense anyway? Because it's not single page, or maybe there is a feature to keep open tabs in the same window?

martinRenou:
Here with shortcuts: 

https://user-images.githubusercontent.com/21197331/175329143-f9d318ae-5c95-47a6-97c3-7bd51af4bed4.mp4

looks quite nice :D (the screen recording was a bit laggy but it works nicely)

app/manifest.webmanifest
```diff
@@ -0,0 +1,32 @@
...
+  "scope": "./",
+  "shortcuts" : [
+    {
+      "name": "JupyterLite",
+      "url": "/lab",
+      "description": "The main JupyterLite application"
+    },
+    {
+      "name": "Replite",
+      "url": "/repl",
```
jtpio, line 28: Maybe with the toolbar enabled?

martinRenou: Thanks! Done :)

jtpio:
Just tried again with the latest after uninstalling the previous app, and this time the icon does not seem to be displayed in the Gnome launcher:

image

jtpio:
OK I just reinstalled it and restarted the Gnome Shell and it is showing up correctly now:

image

jtpio:
It's also cool to see many resources served from the ServiceWorker, for example when starting a new kernel:

image

jtpio (approved):
Thanks!

eugene-f02:
Is there a way to configure jlite to disable ServiceWorker altogether or to disable it for requests made to certain domains? I have a custom extension that makes requests to external service to fetch some data and I don't want it to be cached

PR #690, opened 2022-06-23
Fix micromamba environment on Gitpod

jtpio:
## References

Follow-up to #672 

Turns out the prefix under `~/micromamba` is not persisted when the Gitpod workspaces go idle and need to be restarted. 

## Code changes

Try to place the `micromamba` environment under `/workspace`.

## User-facing changes

None

None

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -10,29 +10,32 @@ github:
 tasks:
   - name: setup
     init: |
-      pushd ~
+      pushd /workspace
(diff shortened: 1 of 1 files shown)


jtpio:
Not the prettiest, but this seems to be more usable now when the workspace becomes idle and is started again.

At least the config is still all self-contained in the `.gitpod.yml` file, but maybe it would make sense to move the micromamba env bootstrap to the `scripts` folder at some point.

jtpio:
Will continue iterating if there are more issues when using the setup.

PR #698, opened 2022-06-28
Remove hardcoded default kernel

### LINKED ISSUE #697: Improve handling of the default kernel
## Description

In deployment where the default `pyolite` is not activated, the default kernel is still `python` which can lead to warnings in the dev tools console such as:

image

## Reproduce

Deploy a JupyterLite website with the xeus kernels and pyolite disabled with:

```json
"disabledExtensions": [
        "@jupyterlite/javascript-kernel-extension",
        "@jupyterlite/pyolite-kernel-extension"
]
```

## Expected behavior

If `jupyterlite-xeus-python` (https://github.com/jupyterlite/xeus-python-kernel) is available and not the default Pyolite, probably it should be the default.

## Context

The default is currently defined here:

https://github.com/jupyterlite/jupyterlite/blob/51172d360e66875b0dfdfae765541840ea3ef94b/packages/kernel/src/kernelspecs.ts#L16

Instead there could be some logic in the server plugin to find a kernel that has `python` as the language. If there is none then pick any kernel as default?

jtpio:
## References

Fixes #697 

## Code changes

Remove the hardcoded default `python` kernel.

## User-facing changes

Should be none for a stock JupyterLite deployment.

For those using `jupyterlite-xeus-python` instead of Pyolite it should default to it.

None


jtpio:
Right that could indeed be handy.

Just checked and with the regular stack the default seems to be coming from `jupyter_client`: 

https://github.com/jupyter/jupyter_client/blob/c81771416d9e09e0e92be799f3e8549d0db57e43/jupyter_client/multikernelmanager.py#L56-L58

app/jupyterlite.schema.v0.json
```diff
@@ -54,6 +54,11 @@
           "default": "./lab",
           "$ref": "#/definitions/non-trailing-slash-uri"
         },
+        "defaultKernelName": {
+          "description": "The name of the default kernel",
```
bollwyvl, line 58: As this goes straight to the docs: perhaps, (and if sorted, as suggested):

> _If no matching kernel is found, the alphabetical first kernel name from the available kernels will be chosen._

packages/kernel/src/kernelspecs.ts
```diff
@@ -12,8 +15,11 @@ export class KernelSpecs implements IKernelSpecs {
     if (this._specs.size === 0) {
       return null;
     }
+    // pick the first kernel if there is no default
+    const defaultKernelName =
+      PageConfig.getOption('defaultKernelName') || this._specs.keys().next().value;
```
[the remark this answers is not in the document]
bollwyvl: and, i guess... if there are _really_ no kernels, it should return... something. probably javascript, as that is the implementation language.

app/jupyterlite.schema.v0.json
bollwyvl: Created https://github.com/jtpio/jupyterlite/pull/4

jtpio:
CI back to green, thanks @bollwyvl for the assist!

psychemedia:
> For those using jupyterlite-xeus-python instead of Pyolite it should default to it.

Does that need to be in the JupyterLite docs, perhaps in the [Bundling additional packages by default](https://jupyterlite.readthedocs.io/en/latest/howto/python/packages.html#bundling-additional-packages-by-default) section?

Maybe adding something to that section along the lines of:

*Alternatively, if you install the `jupyterlite-xeus-python` kernel, packages can be installed into that kernel at build time. If the `jupyterlite-xeus-python` is installed, it will be used as the default kernel in the JupyterLite environments.*

Re: "JupyterLite environments", how should the various environments be described in the JupyterLite and "traditional" full server?) variants? The tooltips on the JupyterLite website demo buttons suggest "Jupyterlab Lite", "RetroLab Lite" and "JupyterLite REPL". Also, presumably Retro/RetroLab will just become "notebook" when the notebook v7 release happens?

jtpio:
> Does that need to be in the JupyterLite docs, perhaps in the [Bundling additional packages by default](https://jupyterlite.readthedocs.io/en/latest/howto/python/packages.html#bundling-additional-packages-by-default) section?

Ah I forgot to update the description in the top comment. This is not the case anymore with the latest changes (before there was some logic to find a kernel with `python` as the language). But we could indeed add that do the docs to mention how to use the `defaultKernelName` key in `jupyter-lite.json` to explicitly choose another default kernel.

Issue #684, opened 2022-06-21
Simplify the dev setup


martinRenou:
## Description

The development setup on JupyterLite is a bit heavy:
- `dodo.py` is 1800 lines of code https://github.com/jupyterlite/jupyterlite/blob/main/dodo.py
- some commands don't do **only** what they advertise, for example `doit serve` will run linters (which takes **way too much** memory) and some other extra things. I'd personally expect `doit serve` to **only** start a static server (could be just a "shortcut" for `python -m http.server`)

## Proposition

Simplify the `dodo.py`.

bollwyvl:
> 1800 lines of code

I can't really apologize much for the length: between the demo, docs, precise ordering of multi-language steps to generate many, many packages with one command, e.g. `doit check`.

We _could_ break some of the task details and utility methods into individual files, but this would actually net _additional_ total lines, because of extra imports, etc.

> some other extra things

Yes, it also does a whole lot, but does it with very few "gotchas." about doing stuff at the right time, in the right order. it's going to re-do as much stuff as it needs to to give you an _accurate_ setup. This was introduced once we started packaging python packages in wheels in webpack in npm tarballs in python packages: without that entire chain _entirely_ automated, it would either be very slow, or very error-prone, or worse yet, both.

> just a "shortcut" for `python -m http.server`

`doit -s serve:core:py` will do precisely that. However, if any `.ts` files had changed, they would not be re-built... much less `py` files that needed to get re-built. We could add that to the docs.

> run linters

Yep, our `eslint` setup is probably kinda broken, and would rather see that fixed (perhaps per-`package.json`) rather than removed from semi-frequent running: we try to stick pretty close to the bleeding edge of several upstreams, and I would hate to lose any of the quality control stuff.

jtpio:
Hopefully some online setup like the one proposed in https://github.com/jupyterlite/jupyterlite/pull/672 can help hide that complexity. Especially when making smaller edits by occasional contributors.

Probably `dodo.py` will also be cleaned up quite a bit when we move Pyolite to a separate repository and make it a federated kernel instead.

jtpio:
Posting some more thoughts about this here as observations that came up when working on this repo on both Gitpod and locally:

- when a task fails it can be difficult to look up the name of the task since some of them are generated in code. Also the *implicit* definitions like `def task_build()` make looking for tasks like `build:docs:favicon` very difficult with full-text search tools (IDEs, `ag`, `grep`...)
[...]

PR #702, opened 2022-06-30
[wip] split up eslint, add linter skips env var

bollwyvl:
## References

- more work towards #194
- makes #684 worse :imp: 

## Code changes

- [x] runs `eslint` per directory in `packages`
  - it is now slower, as starting eslint has overhead, but doesn't _seem_ to error out?
    - probably need to do some actual dev to see more issues
  - subsequent runs will be smaller
  - i personally disable `pre-commit` hooks, this won't impact it
- [x] skips `eslint` for `_metapackage`
- [x] adds some docs about various run options

## User-facing changes

- n/a

- n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,7 +98,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, windows, macos]
-        python-version: ['3.7', '3.10', 'pypy3.8']
+        python-version: ['3.7', '3.10', 'pypy-3.8']
         include:
           # OS-specifics for running pip
           - os: ubuntu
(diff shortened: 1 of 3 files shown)


jtpio:
Nice thanks @bollwyvl for starting this!

> it is now slower, as starting eslint has overhead, but doesn't seem to error out? 

yeah ESLint started to error out again almost consistently locally lately, even on GitPod. Not sure if there is something special we are missing in this repo, since this doesn't not seem to be the case in the larger jupyterlab repo.

martinRenou:
> makes https://github.com/jupyterlite/jupyterlite/issues/684 worse 👿 

😆

jtpio:
Closing as the ESLint issues reported in https://github.com/jupyterlite/jupyterlite/issues/194 have mostly been fixed by https://github.com/jupyterlite/jupyterlite/pull/1212.

PR #711, opened 2022-07-05
Run the JavaScript kernel in a Web Worker

jtpio:
## References

More steps towards https://github.com/jupyterlite/jupyterlite/issues/463 to streamline how the JupyterLite kernels are set up.

## Code changes

- [x] Run the JavaScript kernel in a Web Worker instead of an IFrame
- [x] Some minor tweaks and docstrings on the Pyolite kernel
- [x] Check downstream kernels like p5 -> Will handle updates in https://github.com/jupyterlite/p5-kernel/pull/11

## User-facing changes

TBD

Some changes in the JavaScript kernel:

- remove `protected _eval()` method

### DIFF
--- a/packages/javascript-kernel/src/kernel.ts
+++ b/packages/javascript-kernel/src/kernel.ts
@@ -214,40 +214,42 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
       return;
     }
 
+    const parentHeader = msg.parentHeader || this.parentHeader;
+
     switch (msg.type) {
       case 'stream': {
         const bundle = msg.bundle ?? { name: 'stdout', text: '' };
-        this.stream(bundle, msg.parentHeader);
+        this.stream(bundle, parentHeader);
         break;
       }
       case 'input_request': {
         const bundle = msg.content ?? { prompt: '', password: false };
-        this.inputRequest(bundle, msg.parentHeader);
+        this.inputRequest(bundle, parentHeader);
         break;
       }
       case 'display_data': {
         const bundle = msg.bundle ?? { data: {}, metadata: {}, transient: {} };
-        this.displayData(bundle, msg.parentHeader);
+        this.displayData(bundle, parentHeader);
         break;
       }
       case 'update_display_data': {
         const bundle = msg.bundle ?? { data: {}, metadata: {}, transient: {} };
-        this.updateDisplayData(bundle, msg.parentHeader);
+        this.updateDisplayData(bundle, parentHeader);
         break;
       }
       case 'clear_output': {
         const bundle = msg.bundle ?? { wait: false };
-        this.clearOutput(bundle, msg.parentHeader);
+        this.clearOutput(bundle, parentHeader);
         break;
       }
       case 'execute_result': {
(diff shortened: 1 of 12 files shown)


jtpio:
> * Check downstream kernels like p5 -> Will handle updates in

Changes will be required downsteam, but the p5 kernel will first need the new `@jupyterlite` packages to be released.

packages/javascript-kernel/src/kernel.ts
```diff
@@ -233,61 +177,108 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
...
-        console.warn = console.error;
-
-        window.onerror = function(message, source, lineno, colno, error) {
-          console.error(message);
-        }
-      `
-    );
+    switch (msg.type) {
+      case 'stream': {
+        const bundle = msg.bundle ?? { name: 'stdout', text: '' };
```
bollwyvl, line 221: maybe all the empty responses could be hoisted to  `emptyStreamResponse = Object.freeze({name: 'stdout': text: ''})`, e.g constants up top?

packages/pyolite-kernel/src/worker.ts
```diff
@@ -326,16 +326,13 @@ export class PyoliteRemoteKernel {
    *
    * @param content The incoming message with the comm target name.
    */
-  async commInfo(content: any, parent: any) {
+  async commInfo(content: { target_name: string }, parent: any) {
```
bollwyvl, line 329: seems like we should hoist that `content` to a type.

packages/javascript-kernel/src/kernel.ts
```diff
@@ -15,25 +21,10 @@ export class JavaScriptKernel extends BaseKernel implements IKernel {
...
-      window.addEventListener('message', (e: MessageEvent) => {
-        const msg = e.data;
-        if (msg.event === 'stream') {
-          const content = msg as KernelMessage.IStreamMsg['content'];
-          this.stream(content);
-        }
-      });
-    };
-    document.body.appendChild(this._iframe);
+    this._worker = this.initWorker();
```
bollwyvl, line 24: Same thing here: i think passing the options down is a good idea for "base" kernels like this.

jtpio: Good point :+1:

packages/pyolite-kernel/src/kernel.ts
```diff
@@ -36,12 +36,18 @@ export class PyoliteKernel extends BaseKernel implements IKernel {
    * Subclasses must implement this typographically almost _exactly_ for
    * webpack to find it.
    */
-  protected initWorker(options: PyoliteKernel.IOptions): Worker {
+  protected initWorker(): Worker {
```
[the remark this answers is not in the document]
jtpio: Makes sense based on the comment above: https://github.com/jupyterlite/jupyterlite/pull/711#pullrequestreview-1213706627

packages/pyolite-kernel/src/worker.ts
jtpio: Might revert back to `any` for now and do a pass on the types separately on the other methods as well.

jtpio:
Getting this one in and following-up with a new release right after.

Issue #712, opened 2022-07-05
Writing json to/from files not working - apparent double JSON.stringify()


oscar6echo:
In the following sequence I would expect `dic2` to be the same as `dic1`.  
It is in jupyterlab, not in jupyterlite.

It seems there is an extra `JSON.stringify` when writing to a `.json` file.  
This is a bit crippling if you want to generate, edit, parse such a file.  

Is it easy to fix ?

```py
# Run this code in https://jupyterlite.readthedocs.io/en/latest/_static/lab/
# version Version 0.1.0-beta.10

from pathlib import Path
import json

path = Path('/drive/toto.json')

dic = {'a': 11, 'b': 'qwerty', 'c': [22, 'azerty']}
dic
# {'a': 11, 'b': 'qwerty', 'c': [22, 'azerty']}

txt = json.dumps(dic)
txt
# '{"a": 11, "b": "qwerty", "c": [22, "azerty"]}'

with path.open('w') as f:
    f.write(txt)

with path.open('r') as f:
    txt2 = f.read()
txt2
# '"{\\"a\\": 11, \\"b\\": \\"qwerty\\", \\"c\\": [22, \\"azerty\\"]}"'

dic2 = json.loads(txt2)
dic2
# '{"a": 11, "b": "qwerty", "c": [22, "azerty"]}'

dic2['a']
# error

json.loads(dic2)
# {'a': 11, 'b': 'qwerty', 'c': [22, 'azerty']}
# OK then but surely not what is meant
```

jtpio:
Thanks @oscar6echo for the report and the steps to reproduce.

This probably explains why some of the pandas read / write methods are also not behaving properly: https://github.com/jupyterlite/jupyterlite/issues/682

Issue #682, opened 2022-06-21
Some pandas read / write methods fail with some file types


jtpio:
## Description

As initially reported by @psychemedia in https://github.com/pandas-dev/pandas/pull/47428#issuecomment-1160786870

## Reproduce

Upload and run this example notebook in a JupyterLite instance running `0.1.0b9`: https://gist.github.com/psychemedia/6fd3655782daf031299cfec03043d6bc

For example it looks like there is an issue saving the json file:

![image](https://user-images.githubusercontent.com/591645/174676780-8a23da8a-d9e3-4007-ab27-98d2bbb23230.png)

## Expected behavior

Eventually it would be great to get most of these methods work seamlessly.

## Context

- JupyterLite version: `0.1.0b9`
- Operating System and version: any
- Browser and version: all

hamedgibago:
In method `def test_csv_write(df):` line: `df.to_csv("data/test.csv")`, results should be saved in file in notebook. `results=df.to_csv("data/test.csv")`. But it just convert it. How could I access the file to change it?

jtpio:
Looks like writing to a `.sql` file with the following is giving some errors:

```py
if os.path.exists("test.sql"):
    os.remove("test.sql")
sql_db = sqlite3.connect("test.sql")
df.to_sql(name="test_table", con=sql_db)
sql_db.close()
```

![image](https://user-images.githubusercontent.com/591645/177501165-099aa427-6db0-4780-9a52-48b654459552.png)

psychemedia:
The `pandas.to_csv()` function seems to have stopped working in latest demo site? (How do I check the `jupyterlite` version number?)

jtpio:
> How do I check the `jupyterlite` version number?)

@psychemedia: `Help > About JupyterLite`

> The pandas.to_csv() function seems to have stopped working in latest demo site?

Wondering if this could be related to the change in https://github.com/jupyterlite/jupyterlite/pull/753 which might have invalidated https://github.com/jupyterlite/jupyterlite/pull/714 (cc @martinRenou for confirmation).

jtpio:
Regardless once #386 is done we should add more extensive testing to make sure we don't break kernel-related things.

martinRenou:
[...]

PR #714, opened 2022-07-06
Fix handling of JSON files

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/712

## Code changes

Looks like `.json` files had special treatment.

## User-facing changes

This should fix the use case described in https://github.com/jupyterlite/jupyterlite/issues/712.

As well as some read / write methods of `pandas` and mentioned in https://github.com/jupyterlite/jupyterlite/issues/682:

image

None

### DIFF
--- a/app/services.js
+++ b/app/services.js
@@ -39,7 +39,11 @@ self.addEventListener('fetch', async (event) => {
 
         let content = '';
         if (event.request.method === 'PUT') {
-          content = await event.request.text();
+          if (args.includes('json')) {
+            content = await event.request.json();
+          } else {
+            content = await event.request.text();
+          }
         }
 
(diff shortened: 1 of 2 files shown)


jtpio:
This might still be incomplete for now, as the JSON file seems to be saved encoded: 

image

jtpio:
cc @oscar6echo if you want to try on the ReadTheDocs build:

https://jupyterlite--714.org.readthedocs.build/en/714/_static/lab/index.html

also cc @psychemedia since this should fix the loading of json files with pandas.

jtpio:
Thanks @oscar6echo for checking :+1: 

Getting that one in and will make a new release soon.

PR #722, opened 2022-07-08
Extra Meta Info for Developers

Enforcer007:
It helps devs in understanding about default tasks


jtpio (approved):
Thanks @Enforcer007!

I made a small edit to link to the `dodo.py` file directly, so in case the default tasks change in the future the documentation will less likely be out of sync.

PR #726, opened 2022-07-09
Update no wheels found error message

rickwierenga:
## References

None.

## Code changes

`list_wheels` only returns supported wheels (`ALL_WHL` in `constants.py`). This error is (rightfully) raised when no supported wheels are present, even though there may be other wheels in the directory. In that case, this new error message is more descriptive.

## User-facing changes

```
ValueError: no wheels found in /dir
```

will be

```
ValueError: no supported wheels found in /dir
```

None.


jtpio (approved):
Thanks!

Issue #727, opened 2022-07-09
Using a local bz2 compressed pyodide distribution


rickwierenga:
## Description

Using a local `.bz2` archive is not supported, `UnboundLocalError: local variable 'will_fetch' referenced before assignment` will always be raised, even when it shouldn't.

## Reproduce

`jupyter lite build --pyodide /path/to/somewhere/pyodide.tar.bz2`

## Expected behavior

\-

## Context

- JupyterLite version: latest
- Operating System and version: macOS 12.3.1 
- Browser and version: n/a

jtpio:
Closing as fixed by https://github.com/jupyterlite/jupyterlite/pull/728, thanks @rickwierenga!

PR #728, opened 2022-07-09
Declare will_fetch early

rickwierenga:
This prevents an error `UnboundLocalError: local variable 'will_fetch' referenced before assignment` when using a local `.bz2` file.

## References

#727 

## Code changes

Set `will_fetch = False` if not `https?` scheme.

## User-facing changes

\-

\-


jtpio (approved):
Thanks!

PR #733, opened 2022-07-11
Bugfix:Make input work again

joemarshall:
The input and getpass callbacks have become object methods, so they need to be passed as bound objects now. Otherwise calls to await input fail with undefined this errors. 

Try calling await input("hello") on the jupyterlite demo page. In Chrome it fails for me on current build (and any build since comlink came in).


jtpio (approved):
Good catch, thanks!

PR #737, opened 2022-07-13
Service worker configuration options in jupyter-lite.json

joemarshall:
I added a couple of options to the serviceworker plugin, to allow you to 

a) Point it at somewhere else in the case that you want to host jupyterlite files in a subfolder of your site, but want the service worker to be able to handle things at that level.
b) Disable the service worker (deleting the services.js script isn't enough, because if you use your own service worker, jupyterlite clobbers it every time you load the page, even if the script is missing)

### DIFF
--- a/packages/server/src/serviceworker.ts
+++ b/packages/server/src/serviceworker.ts
@@ -7,7 +7,8 @@ import { IServiceWorkerRegistrationWrapper } from './tokens';
 export class ServiceWorkerRegistrationWrapper
   implements IServiceWorkerRegistrationWrapper
 {
-  constructor() {
+  constructor(config: ServiceWorker.IConfig) {
+    this._config = config;
     this.initialize();
   }
 
@@ -29,6 +30,19 @@ export class ServiceWorkerRegistrationWrapper
   }
 
   private async initialize() {
+    if (this._config['disabled']) {
+      // don't install service worker
+      return;
+    }
+    const workerUrl =
+      this._config['workerUrl'] || URLExt.join(PageConfig.getBaseUrl(), 'services.js');
+    if (!('serviceWorker' in navigator)) {
+      console.error(
+        'ServiceWorker registration failed: Service Workers not supported in this browser'
+      );
+      this.setRegistration(null);
+    }
+
     if (!('serviceWorker' in navigator)) {
       console.error(
         'ServiceWorker registration failed: Service Workers not supported in this browser'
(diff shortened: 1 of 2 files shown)


martinRenou (approved):
Thanks!

joemarshall:
Any chance this could be merged into main ready for the next release, so I can drop my patches from the build?

packages/server/src/serviceworker.ts
```diff
@@ -29,6 +30,19 @@ export class ServiceWorkerRegistrationWrapper
...
+    }
+    const workerUrl =
+      this._config['workerUrl'] || URLExt.join(PageConfig.getBaseUrl(), 'services.js');
+    if (!('serviceWorker' in navigator)) {
+      console.error(
+        'ServiceWorker registration failed: Service Workers not supported in this browser'
+      );
+      this.setRegistration(null);
+    }
+
```
martinRenou, line 45: This code has been duplicated (it's already just below)

```suggestion
```

martinRenou (changes requested):
Sorry, my previous review was from the phone I missed something

packages/server/src/serviceworker.ts
```diff
@@ -29,6 +30,12 @@ export class ServiceWorkerRegistrationWrapper
   }
 
   private async initialize() {
+    if (this._config['disabled']) {
```
jtpio, line 33: Normally the following notation is preferred when the object is properly typed:

```suggestion
    if (this._config.disabled) {
```

jtpio:
>  Disable the service worker (deleting the services.js script isn't enough, because if you use your own service worker, jupyterlite clobbers it every time you load the page, even if the script is missing)

@joemarshall curious if you tried using the plugin system to swap the default plugin with your own?

Thinking that as an alternative to adding a config option like in this PR, maybe it's possible to add `@jupyterlite/server-extension:service-worker` to `disabledExtensions` so the plugin is disabled. Then you can provide your own plugin with full control on it.

echarles:
@joemarshall I was looking for that feature, actually not only for the `'services.js` loading but also for the resources loaded by the Pyolite (python) kernel. The needed resources (all.json...) are for now expected to be served under the `build/pypi` folder of the generated jupyterlite website. 

https://github.com/jupyterlite/jupyterlite/blob/dac5a8202e25c713cf653454025c832c5b7b0a2d/packages/pyolite-kernel/src/kernel.ts#L61

At some point, those resources have the same requirements as the `services.js`, served by another separated server under a configurable URL.

Have you though expanding this PR to that case (anyway worth thinking about that while adding more configuration).

jtpio:
> are for now expected to be served under the `build/pypi` folder of the generated jupyterlite website.

Right. This will likely be fixed when the Pyodide kernel is moved outside of the main repository: https://github.com/jupyterlite/jupyterlite/issues/386

jtpio:
@joemarshall just checking if this PR is still valid?

There have been a few iterations on the service worker lately which should allow for more flexibility:

- https://github.com/jupyterlite/jupyterlite/pull/839
- https://github.com/jupyterlite/jupyterlite/pull/899

joemarshall:
I think probably you can close this, at least for my purposes - I ended up requiring site isolation for the streaming code in pyodide-http, because if you can isolate, that is much cleaner, and if you don't have server access to isolate the site, you can isolate using a service worker anyway.

It'll all be a moot point once the Webassembly/promise integration proposal is no longer behind experimental flags in browsers, because no one will need serviceworkers for handling io any more, so they can just stick to being for caching as they are designed for.

jtpio:
> I think probably you can close this

Alright thanks @joemarshall for getting back and for investigating this :+1:

PR #738, opened 2022-07-14
Alter parent settings to messages

joemarshall:
I don't understand quite how the parent message thing works, but setting the parent on every message breaks things. 

This is why the pyolite kernel breaks if you call await input twice, because it sets 'input_reply' as the parent message, rather than keeping the 'execute_request' there. With this fix, await input works fine.

I'm not convinced that this fix is perfect and that it doesn't need setting for other messages, but I thought it was worth throwing it out for discussion.

### DIFF
--- a/packages/kernel/src/kernel.ts
+++ b/packages/kernel/src/kernel.ts
@@ -103,9 +103,10 @@ export abstract class BaseKernel implements IKernel {
   async handleMessage(msg: KernelMessage.IMessage): Promise<void> {
     this._busy(msg);
 
-    this._parent = msg;
-
     const msgType = msg.header.msg_type;
+    if(msgType=='execute_request'){
+      this._parent = msg;
+    }    
     switch (msgType) {
       case 'kernel_info_request':
         await this._kernelInfo(msg);


joemarshall:
This could just check if the message is 'input_response' and not update parent there. I'm not sure whether there are any other messages that need a parent message or not.

joemarshall:
I've just had a read through
https://ipython.org/ipython-doc/3/development/messaging.html

and I think parent header should only be set for execute requests for now. I think otherwise there is a risk that other messages will break the parent header stuff. I think if anything uses both execute and custom comms it will break, but that would happen now also. Really for it to work correctly, parent should be passed into the kernel execute request, which could then pass it out again, so you've always got the correct parent message.

jtpio:
Thanks @joemarshall.

The current handling of the parent messages is indeed not optimal at the moment. However it's not clear whether limiting it to `execute_request` might have some unintended side-effects on the other messages. The parent message could for example be used by kernels to track a chain of messages (although this is not currently done in the Pyolite and JS kernels).

packages/kernel/src/kernel.ts
```diff
@@ -103,9 +103,10 @@ export abstract class BaseKernel implements IKernel {
   async handleMessage(msg: KernelMessage.IMessage): Promise<void> {
     this._busy(msg);
 
-    this._parent = msg;
-
     const msgType = msg.header.msg_type;
+    if (msgType === 'execute_request') {
+      this._parent = msg;
```
Carreau, line 108: Could this be moved to `this._execute` maybe ?

PR #746, opened 2022-07-22
Upgrade to jupyterlab 3.4.4

bollwyvl:
## References

- n/a

## Code changes

- [x] updates to jupyterlab 3.4.4
- [x] add `jupyterlab-github` to conda dev environment 
  - [x] set `defaultRepo` to `jupyterlite/jupyterlite` in `overrides.json`
- [x] disabled webpack caching in CI :crying_cat_face: 
  - the `ui-tests` were doing something awful i didn't have the patience to deal with right now

## User-facing changes

- should fix some memory leaks

- ???


bollwyvl:
some notes from RTD (maybe doesn't need to block this):
- the ipycanvas API has changed a bunch
  - maybe there's a new version we can pull
- emscripten filesystem is wrong
  -  maybe we just get rid of it
- is there a way to get at `GitHub` content sfrom inside a kernel?

app/lab/package.json
```diff
@@ -85,7 +85,7 @@
     "@jupyterlite/types": "~0.1.0-beta.11",
     "@jupyterlite/ui-components": "~0.1.0-beta.11",
     "@lumino/algorithm": "~1.9.1",
-    "@lumino/application": "~1.28.1",
+    "@lumino/application": "^1.27.0",
```
jtpio, line 88: Wondering if this one needs the `^`? Or should all  `@lumino` dependencies use the same `~` for consistency?

jtpio:
> * is there a way to get at `GitHub` contents from inside a kernel in `/drive`?

I was expecting this to be supported yes. Although there might be some issues as the list of files doesn't seem to be matching the one from the GitHub drive?

image

martinRenou:
Thanks for pinging me @bollwyvl, I pushed a PR that should fix the ipycanvas example https://github.com/jupyterlite/jupyterlite/pull/748

jtpio:
> * the `ui-tests` were doing something awful i didn't have the patience to deal with right now

Curious on the issue so this could be fixed separately.

jtpio:
And thanks @bollwyvl for doing the update!

martinRenou:
> is there a way to get at GitHub contents from inside a kernel in /drive?

Yes I made sure it supported custom drives. It also works with the jupyterlab-filesystem-access extension and I'd expect it to work with the s3 drive https://github.com/IBM/jupyterlab-s3-browser

The downside of it is that people won't be able to create files with the Github drive, because they are simply not allowed to do it. 

> Although there might be some issues as the list of files doesn't seem to be matching the one from the GitHub drive?

This is odd indeed. It might be interesting to see what the `get` drive method gave back from the request.

app/lab/package.json
bollwyvl: was emulating [upstream](https://github.com/jupyterlab/jupyterlab/blob/v3.4.4/jupyterlab/staging/package.json#L113). with the `~` was getting some `console.warning`, but maybe adding it to the `resolution` in the root `package.json` will resolve it...

jtpio: Alright let's keep that way then :+1:

jtpio (approved):
Thanks!

jtpio:
> some notes from RTD (maybe doesn't need to block this):

Let's track these separately then.

Issue #701, opened 2022-06-29
IPython display failure (caused by service worker)


qqdaiyu55:
## Description

Jupyterlite is stuck on below codes. I inspected the request, it was pending on `https://jupyterlite.readthedocs.io/en/latest/_static/api/drive//%3Cspan%20style=%22color`. Looks like service worker somehow truncates string after `#`.

code snippet
```python
from IPython.display import display, HTML

display(HTML('<span style="color#6c5ce7">debug</span>'))
```

martinRenou:
This is odd. The service worker should not be involved in this, in any way.

PR #753, opened 2022-08-02
Refactor drivefs API

martinRenou:
## References

Attempt to fix https://github.com/jupyterlite/jupyterlite/issues/701

## Code changes

Refactoring the DriveFS service API so that we don't mess with any URLs formatting


martinRenou:
I restarted the CI and it's green now, not sure why it was failing before.

packages/contents/src/drivefs.ts
```diff
@@ -11,6 +12,12 @@ export const DRIVE_SEPARATOR = ':';
 const encoder = new TextEncoder();
 const decoder = new TextDecoder('utf-8');
 
+export interface IDriveRequest {
```
jtpio, line 15: Maybe we can add a quick docstring to this interface since it's exported?

Or remove the `export` if it doesn't need to be exported?

martinRenou: Thank you for the review! I added a docstring and used this interface where we need it

jtpio (approved):
Nice thanks!

PR #754, opened 2022-08-02
Trigger CI

martinRenou:
## References

UI tests are failing in https://github.com/jupyterlite/jupyterlite/pull/753 but it seems irrelevant to the changes. So I'm testing here.


martinRenou:
It looks like it's passing here. So let's close.

PR #757, opened 2022-08-03
Add mock jupyterlab, notebook for piplite, ignore_piplite_builtins

bollwyvl:
## References

- n/a

## Code changes

- [x] adds mock packages 
  - [x] `jupyterlab`
  - [x] `notebook`
- [x] revs versions of other mocked packages
- [x] adds optional `CHECK_MOCK_VERSIONS` for `doit lint` to compare against as-installed versions
- [x] add a `ignore_piplite_builtins` key to avoid indexing/copying mocks

## User-facing changes

- users should be able to `piplite.install` packages that specify a compatibility constraint for `jupyterlab` or `notebook`, common in widget packages

- these packages, if _actually_ used inside a `pyolite` kernel will break
  - if problems arise, a site owner can use `ignore_piplite_mocks` can be used to avoid them


bollwyvl:
Some notes from RTD, maybe unrelated to this PR:

- cytoscape
  - attachments are broken... maybe serviceworker related? @martinRenou 
    ```
    GET attachment:16109f49-1053-42ab-b02d-9685e0743e6e.png net::ERR_UNKNOWN_URL_SCHEME
    attachment:16109f49-1053-42ab-b02d-9685e0743e6e.png?1659547370229:1          
    GET attachment:16109f49-1053-42ab-b02d-9685e0743e6e.png?1659547370229 net::ERR_UNKNOWN_URL_SCHEME
```

jtpio:
Thanks @bollwyvl.

> common in widget packages

Do you have examples of such widget packages that still have `notebook` or `jupyterlab` as a dependency?

Normally they can only depend on `ipywidgets` like the examples we have in this repo. Wondering if there is a way to avoid mocking more packages in pyolite (which complicates things), or if requiring custom widgets to drop their dependencies on `notebook` and `jupyterlab` is not practical?

bollwyvl:
I agree for widgets they shouldn't, but practically they do... don't have a
full list right now, as some of the pain of this was from work stuff, but
have encountered the legitimate case where there's a client component
(mime, custom ui) bundled with a server extension... Heck, should add
jupyter_server, too, as it's definitely not importing any time soon.

martinRenou:
Maybe instead of packaging mock modules, we could have piplite simply ignore those packages for the installation?

jtpio:
Closing as this would more likely be implemented in https://github.com/jupyterlite/pyodide-kernel directly if still relevant.

Thanks!

PR #758, opened 2022-08-04
Change kernel from notebook

### LINKED ISSUE #752: Change kernel
Changing the kernel doesn't work.

## Description
When trying to change the kernel, there is no response.
The [Kernels class](https://github.com/jupyterlite/jupyterlite/blob/c5f87fc0dcf3dd597d819c50296b71bd427dd201/packages/kernel/src/kernels.ts#L20) has only `startNew`, `restart`, and `stop`. Maybe it is necessary to implement more functionalities, but I'm not sure, I haven't debugged this issue in depth yet.

## Reproduce
Open a notebook, and try to change the kernel.
For example, open the p5.js example and try to change the kernel to the `javascript`. It doesn't work.

Instead, open the same notebook and change the kernel to `No kernel` and then change it to `javascript`. It does work.

## Context
https://github.com/QuantStack/jupyterlab-blockly/issues/55

https://user-images.githubusercontent.com/26092748/181744193-faebf13c-2d6c-4324-9063-3ce3ae6f46cd.mp4

hbcarlos:
## References
Fixes #752 

## Code changes
Implemented the logic to change the kernel of a session.

N/A

https://user-images.githubusercontent.com/26092748/182847438-0edd44c7-c951-4d9d-a98e-db1a09b40c33.mp4

### DIFF
--- a/packages/session/src/sessions.ts
+++ b/packages/session/src/sessions.ts
@@ -83,7 +83,7 @@ export class Sessions implements ISessions {
         });
 
         if (newKernel) {
-          patched['kernel'] = newKernel;
+          patched.kernel = newKernel;
         }
       }
     }
--- a/packages/session/src/sessions.ts
+++ b/packages/session/src/sessions.ts
@@ -73,7 +73,7 @@ export class Sessions implements ISessions {
           (session) => session.kernel?.id === kernel?.id
         );
         if (session) {
-          patched['kernel'] = session?.kernel;
+          patched.kernel = session.kernel;
         }
       } else if (kernel.name) {
         const newKernel = await this._kernels.startNew({
--- a/packages/session/src/sessions.ts
+++ b/packages/session/src/sessions.ts
@@ -46,7 +46,6 @@ export class Sessions implements ISessions {
   /**
    * Path an existing session.
    * This can be used to rename a session.
-   * TODO: read path and name
    *
    * - path updates session to track renamed paths
    * - kernel.name starts a new kernel with a given kernelspec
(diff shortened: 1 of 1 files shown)


packages/session/src/sessions.ts
```diff
@@ -62,6 +65,29 @@ export class Sessions implements ISessions {
       path: path ?? session.path,
       name: name ?? session.name,
     };
+
+    if (kernel) {
+      // Kernel id takes precedence over name.
+      if (kernel?.id) {
```
jtpio, line 71: Looks like the `?` here might not be needed if there is already a guard above with `if (kernel)`?

packages/session/src/sessions.ts
```diff
@@ -62,6 +65,29 @@ export class Sessions implements ISessions {
...
+        const session = this._sessions.find(
+          (session) => session.kernel?.id === kernel?.id
+        );
+        if (session) {
+          patched['kernel'] = session?.kernel;
```
jtpio, line 76: Wondering if it's possible to use the following notation?

```suggestion
          patched.kernel = session.kernel;
```

Normally TypeScript should be able to infer the type based on the function return type.

hbcarlos:
Thanks Jeremy

packages/session/src/sessions.ts
```diff
@@ -62,6 +65,29 @@ export class Sessions implements ISessions {
...
+        }
+      } else if (kernel.name) {
+        const newKernel = await this._kernels.startNew({
+          id: UUID.uuid4(),
+          name: kernel.name,
+          location: PathExt.dirname(patched.path),
+        });
+
+        if (newKernel) {
+          patched['kernel'] = newKernel;
```
jtpio, line 86: Another one here: 

```suggestion
          patched.kernel = newKernel;
```

jtpio (approved):
Thanks!

PR #775, opened 2022-08-19
Properly download binary content

### LINKED ISSUE #676: Download binary file from the UI does not work
## Description

If you try downloading a file with type `base64` from the UI (right click + download) then it downloads a file with the content being the base64 string, which makes it broken.

## Reproduce

1. Upload an image in the UI
2. Download it back
3. Try to open it with your favorite image viewer, it doesn't work
4. Look at the file content, you'll see the base64 string instead of raw bytes

## Expected behavior

It should save bytes

vasiljevic:
## References

Attempt to fix #676 

## Code changes

Functions  [downloadContent](https://github.com/jupyterlite/jupyterlite/blob/34a19bbc89973a9faaf8155e40e827ab8b6ca190/packages/application-extension/src/index.tsx#L228) and [formatContent](https://github.com/jupyterlite/jupyterlite/blob/34a19bbc89973a9faaf8155e40e827ab8b6ca190/packages/application-extension/src/index.tsx#L237) from /packages/application-extension/src/index.tsx are merged into the [modified function downloadContent](https://github.com/vasiljevic/jupyterlite/blob/26ddfa2508e748a4a1cde03e6a631eb342a1c0ea/packages/application-extension/src/index.tsx#L228) and additionally there are related modifications of code at two places in the same file where we call those functions.

## User-facing changes

Binary files can be regularly downloaded.

None

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -225,21 +225,30 @@ const downloadPlugin: JupyterFrontEndPlugin<void> = {
       return !!(currentWidget && docManager.contextForWidget(currentWidget));
     };
 
-    const downloadContent = (content: string, path: string) => {
+    const downloadContent = async (contentPath: string, fileName: string) => {
+      const model = await contents.get(contentPath, { content: true });
       const element = document.createElement('a');
-      element.href = `data:text/json;charset=utf-8,${encodeURIComponent(content)}`;
-      element.download = path;
+      if (model.type === 'notebook' || model.format === 'json') {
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

vasiljevic:
> Thanks!

You’re welcome. I'm looking forward to a release with this fix included.

jtpio:
> I'm looking forward to a release with this fix included.

Yes I'll check if we can get a few more PRs in before the next release.

Issue #312, opened 2021-08-31
Update the examples to ipywidgets 8


jtpio:
### Problem

The latest `ipywidgets` 7 depends on `ipython_genutils` in some places in the code. See https://github.com/jupyter-widgets/ipywidgets/pull/3262 and #311 for reference.

This doesn't seem to be the case with `ipywidgets` 8 anymore. So that could be a good time to start testing the latest ipywidgets 8 alpha releases.

But it's also nice to start testing this version to help with the final release.

### Proposed Solution

We can start updating the example notebooks to the latest 8 prereleases. That could be a draft PR for now.

We might need the prereleases to be published to a `ipywidgets` conda channel to fetch the lab prebuilt extension.

bollwyvl:
Would certainly want to see a release out with the stable deps before jumping on another alpha train :stuck_out_tongue_closed_eyes:

jtpio:
> Would certainly want to see a release out with the stable deps 

A JupyterLite release? Yes we should be able to cut a new one soon :+1:

bollwyvl:
Very likely these don't all work yet... but we need to offer a no-op `widgetsnbextension@4` to explore further...

bollwyvl:
It seems like it's possible to overload the `widgetsnbextension` mock package:

https://github.com/deathbeds/ipydrawio/pull/99 (very wip)

bollwyvl:
There are two PRs that update all the demos (which appear to work) for ipywidgets 8:
- #795
- #793

The main difference is the latter will be slightly more performant, as widgetsnbextension only consists of js/css that we won't use.

bollwyvl:
Some numbers:
- without shim: 30.9mb
  -  https://jupyterlite--795.org.readthedocs.build/en/795/_static/lab/index.html?path=pyolite/interactive-widgets.ipynb
- with shim: 28.8mb  
  - https://jupyterlite--793.org.readthedocs.build/en/793/_static/lab/index.html?path=pyolite/interactive-widgets.ipynb

PR #783, opened 2022-08-30
Add widgetsnbextension 4.x shim

bollwyvl:
## References

- starts #312
  - it's possible (but seems unlikely) every one or our upstream example dependencies is going to work out of the box with ipywidgets 8
  - this gives us a way to incrementally move forward.

## Code changes

- [x] rework dodo piplite indexing for multiple versions of the same package
- [x] add `widgetsnbextension4` shim package

## User-facing changes

- `await piplite.install('ipywidgets=8')` should work
  - note that `federated_extensions` only allow us to install one at a time
- downstream sites can start updating to `ipywidgets=8` without undue [hacks](https://github.com/deathbeds/ipydrawio/pull/99)

- if not specified, it's possible mismatching versions of ipywidgets could get installed if not pre-cached
  - this might have been prevented previously by `widgetsnbextension` not being available


jtpio:
Thanks!

Looks good overall, but might need a second pass now that https://github.com/jupyterlite/jupyterlite/pull/791 has been merged:

https://github.com/jupyterlite/jupyterlite/blob/4da5f0157c2334fc3bf4979bfdb133ed09b6181a/dodo.py#L915

bollwyvl:
tried to cheap merge, but will look more at "home" (another chair/laptop). 

I think for the RTD demo we need to hang tough with 7 until we're ready to look at #312 in earnest. I would think _missing_ demos is better than _broken_ demos, but we won't know until we look, and probably shouldn't wait, as it's causing active breakage.

bollwyvl:
Tried everything out locally on RTD, etc. and widget-adjacent stuff looks ok with 7.

I'll start another PR that investigates going to 8.

martinRenou (approved):
Looks good to me!

I wonder why we needed a shim for widgetsnbextension in the first place? (I know I was the one adding it, but I don't recall why the piplite install was failing) Just asking because it'd be nice to get rid of this entirely at some point.

bollwyvl:
I'll take another look at whether it is even needed...

bollwyvl:
Opened #795 which doesn't ship the shim... the biggest downside is the wheel is 2mb of js we won't use inside the emscripten vm...

martinRenou:
> the biggest downside is the wheel is 2mb of js we won't use inside the emscripten vm...

Yeah.. That's one good reason to keep the shim.

I remove the widgetsnbextension shim from emscripten-forge because it became a noarch package in conda-forge recently, but that makes me think we should probably filter out widgetsnbextension from what's included in the page.

bollwyvl:
>  filter out widgetsnbextension

as mentioned in a few places: i am loathe to dig too much into the underlying `micropip` stuff to do that kind of surgery... there isn't a whole lot of stable API surface to work with.

bollwyvl:
At any rate, #795 appears to be working well on the site. The only thing I could imagine that would improve things were if `ipywidgets` became a metapackage, somehow, and there was a way to install _just_ the kernel piece... the "simple" way, `ipywidgets[notebook,lab]` is unsatisfying for various reasons, though...

bollwyvl:
Closing this in favor of the decision between #795 (simplicity) and #783 (speed)

PR #791, opened 2022-09-07
Update to JupyterLab 3.4.6

bollwyvl:
## References

- probably conflicts with #783

## Code changes

- [x] update to [JupyterLab 3.4.6](https://github.com/jupyterlab/jupyterlab/releases/tag/v3.4.6) packages
- [x] resolve examples with new versions
- [x] pin demo ipywidgets to `<8` for now
- [x] refreshes kernespec icons with optimized SVGs from upstream sources (saves ~10kb)

## User-facing changes

- not a lot

- n/a


bollwyvl:
Looks like the UI tests are to the non-conforming python logo, which we [also need to update](https://github.com/jupyterlite/jupyterlite/blob/main/app/kernelspecs/python.png)... or better still, source from the upstream code.

bollwyvl:
Oh, it's the python _icon_, which we do already pull in properly.

I've gone ahead and pulled fresh javascript and python svg logos, and run them through svgo... saves about 10kb, which isn't nothing!

jtpio (approved):
Thanks!

bollwyvl:
thanks for cleaning up the snapshots... will look at #783 presently...

PR #793, opened 2022-09-08
Update demos to ipywidgets 8, use widgetsnbextension shim

bollwyvl:
## References

- fixes #312
- includes #783
  - at least for a while, we'll need to ship `widgetsnbextension 3 || 4`... doesn't cost much

## Code changes

- [x] bump to `ipywidgets >=8`
- [x] handle `ipycytoscape`
  - [ ] removed from environment for now
  - [ ] restore after another release of
    - [x] overly-conservative conda package https://github.com/conda-forge/ipycytoscape-feedstock/pull/24
    - [ ] (better) the pip upstream _and_ conda package https://github.com/cytoscape/ipycytoscape/pull/336
- [x] resolve examples
- [ ] check on RTD
- [ ] ???
- [ ] docs about `widgetsnbextension` compatibility?

## User-facing changes

- visitors to the RTD website should be able to use ipywidgets 8

## Backwards-incompatible changes

- n/a

PR #795, opened 2022-09-09
Remove widgetsnbextension shim, update demos to ipywidgets 8

### LINKED ISSUE #312: Update the examples to ipywidgets 8

bollwyvl:
## References

- fixes #312
- alternative to #793
- alternative to (but includes) #783 
  
## Code changes

- [x] removes references to `widgetsnbextension`
- [x] leaves machinery for multiple versions for future use
- [x] handle `ipycytoscape`
  - [ ] removed from environment, downloaded wheels for now
  - [ ] restore after another release of
    - [x] overly-conservative conda package https://github.com/conda-forge/ipycytoscape-feedstock/pull/24
    - [ ] (better) the pip upstream _and_ conda package https://github.com/cytoscape/ipycytoscape/pull/336

## User-facing changes

- `ipywidgets 8` or `7` should both be `piplite.install`able
  - still tied to the version of the labextension, but that will always be a problem with un-pinned dependencies
- `piplite.install` of `ipywidgets` will download ~2mb of `widgetsnbextension` the first time, and have to extract it every time

- removes some ts-exported members like `widgetsnbextensionUrl`
- increases the likelihood of kernel/labextension mismatches


bollwyvl:
Aside from `ipycytoscape`, this is ready to roll. I guess if anything i'd rather pull out the notebook and resolve with `depfinder` before merging, rather than shipping a known-broken demo.

martinRenou:
Thanks @bollwyvl ! Should we go with this PR and close https://github.com/jupyterlite/jupyterlite/pull/793 and https://github.com/jupyterlite/jupyterlite/pull/783 ?

bollwyvl:
Welp, every time we claw back a few hundred kb is a win, and keeping the shim saves 2mb.

The real win would be in figuring out how to more fully trick micropip into doing less work/downloading with configuration (and without a build step), and/or handle patches more lazily with loader hacks, but as i've said a few times, every additional thing we do around `micropip` internals is a liability: we break it, we buy it. It has required some additional hacks almost every pyodide release. Things might get better upstream, though, as they are recognizing that single versions of packages from exactly two sources (pyodide distribution and pypi) isn't really sustainable.

martinRenou:
So, as of today, you'd prefer to keep the shim around so we can save those 2mb? Should we close this PR and merge https://github.com/jupyterlite/jupyterlite/pull/793 then?

bollwyvl:
Yes. Of course, what we really need is getting `matplotlib` (and therefore `numpy`) off the critical path, but in the meantime, I don't think it's a price we should pay.

martinRenou:
Sounds good! Closing in favor of https://github.com/jupyterlite/jupyterlite/pull/793

bollwyvl:
Welp... Every time we save a couple 100k on the wire before getting to live
pixels, it's great, and this is 2mb. I don't really know how to feel in
this case, much like the one that shims notebook and jupyterlab.

If we did dig into micropip further, and make it easier to create noop
packages by configuration without a build step somehow, that might be the
best of all.

PR #805, opened 2022-09-16
[wip] Support workspaces

### LINKED ISSUE #139: Support Workspaces
### Problem

When I start JupyterLite, it shows the default UI with the Launcher and File tree, which is good! When I get a few documents and/or linked outputs open, and restart JupyterLite, I lose the state of what I was working on.

### Proposed Solution

Support an implementation of `api/workspaces/`, as described in the [jupyterlab_server API spec](https://github.com/jupyterlab/jupyterlab_server/blob/master/jupyterlab_server/rest-api.yml#L160).

The first time out, starting from the `default`, the workspace can be saved either to the _same_ `localForage` (as they reference contents paths), or to a separate one altogether, and introduce (further) router hacks for `?workspace=<name-of-workspace>` (and `?reset, probably`).

When reloading, this would be consulted. This is _might_ be sufficient for an MVP, but would probably still be broken.

Supporting the import/export of `*.jupyterlab-workspaces` files (and being able to pre-populate/pre-index this folder) would make it much easier to drive/link to a very specific lab configuration e.g. 
- for a docs site:
  - `https://my-docs.rtfd.io/lite/?workspace=docs`
    - `https://my-docs.rtfd.io/lite/?workspace=dashboard` 
- for a particular snippet in a blog post
  -  `https://my-blog.engineering/lite/?workspace=example-1-layout`

### Additional context

- might help #131
- might need #104, but this should probably be first-party

bollwyvl:
## References

- fixes #139
- (mistakenly) includes #793
  - will keep working on this, just to get more eyes on the potential changes, even if we go with #795

## Code changes

- [ ] add a workspaces API handler 
  - [x] `localForage`
  - [ ] server-backed in `/api/workspaces/all.json`?
- [ ] ???
- [ ] support workspaces definitions in 
  - [ ] `jupyter_build_config.json` against given files/folders
  - [ ] `jupyter-lite.json`?
  - [ ] in-contents `.jupyterlab-workspace` files?

## User-facing changes

- when reloading `/lab/`, existing windows should be re-opened
- kernel state/ids are lost, so some things won't work

- ???


psychemedia:
Will a user be able to havr two different workspaces open in different browser tabs for the same jupyterlite environment?

dodo.py
```diff
@@ -1706,11 +1756,88 @@ def _ensure_resolutions(app_name):
...
+
+            print(
+                "\n".join(
+                    difflib.context_diff(
+                        raw.splitlines(), new_raw.splitlines(), "old", "new"
+                    )
+                )
+            )
+
+            if C.CI and not C.CHECKING_RELEASE:
```
jtpio, line 1772: Maybe we still need to run this in the check release workflow to make sure resolutions have been updated and not leave a git dirty state.

bollwyvl: Hm... i guess i though that check-release _should_ be allowed to changes things, while "plain old lint" should just fail hard if a new package gets added, for example (which is what bit me)

jtpio:
Thanks @bollwyvl for working on this.

I did a quick check on RTD and there seems to be an issue with RetroLab. Here is what it looks like:

https://user-images.githubusercontent.com/591645/194490260-09b8d907-59aa-49ed-a8ff-c8272b5e3a80.mp4

bollwyvl:
Thanks for the thorough review. Need to sit back and take another look at it... maybe a path forward is for me to split up the `repo` plumbing stuff and then do the workspaces stuff on top of that...

jtpio:
Bumping to `0.2.0` so we have time to update to Notebook 7 in the meantime (#826) and address issues in https://github.com/jupyter/notebook if needed, since RetroLab is not maintained anymore.

jtpio:
Notebook 7 does not use workspaces at the moment. I think this was discussed recently. Probably it won't make it for 7.0, but could be considered for 7.x (as long as we don't create another JupyterLab).

juntyr:
What’s the status of this PR?

jtpio:
> What’s the status of this PR?

Looks like it has stalled and there are a couple of conflicts to fix. But it's still a feature that would be useful to have in JupyterLite! If anyone wants to help with this, please feel free, thanks!

PR #811, opened 2022-09-21
Configure Pyolite to pre-install packages

vasiljevic:
## References

This is proposed solution for jupyterlite/pyodide-kernel#60 

## Code changes

The central code change is in the `PyoliteRemoteKernel.initKernel()` method. Those lines are added:

```ts
    for (const packageName of options.pipliteRequiredPackages) {
      await this._pyodide.runPythonAsync(`
         await piplite.install(['${packageName}'], keep_going=True);
      `);
    }
```

The idea is to install configurable packages in the same way as some other packages are already pre-installed (in the same method above the added lines). 

New option `pipliteRequiredPackages` may be configured in jupyter-lite.json like:

```json
        "litePluginSettings": {
          "@jupyterlite/pyolite-kernel-extension:kernel": {
            "pipliteRequiredPackages": ["openpyxl"]
          }
        }
```

All other changes implement propagation of the new option from the runtime configuration to the `options` parameter of the `initKernel()` method.

Documentation is also added in `docs/howto/python/packages.md`.

## User-facing changes

A user who build JupyterLite website may configure packages to be pre-installed. Then a who access the JupyterLite website does not need to do `piplite.install` for those packages and may just import them.

No backwards-incompatible changes


vasiljevic:
I have just added the commit 'Make matplotlib optional but included by default'. Currently, there is hard-coded pre-installation of matplotlib. 

When we introduce an option to configure what packages will be pre-installed, it makes sense to let matplotlib be configurable. The reason to include this change in this PR is backward compatibility:  this change is backwards compatible now, but will not be backwards compatible later. 

In other words, we should decide now will we instruct those who use `pipliteRequiredPackages` parameter to put matplotlib in the list when matplotlib is needed.

jtpio:
Thanks @vasiljevic for opening this PR.

Although there are good reasons for not encouraging code to be run silently (as discussed in https://github.com/jupyterlite/jupyterlite/issues/186), the new config option looks simple enough and opt-in. This would allow site deployers the ability to configure it if they want to. But it could indeed make it more difficult to troubleshoot in case something goes wrong during the kernel initialization.

If things move forward, it would be interesting to check how to unify this with what xeus-python offers: https://xeus-python-kernel.readthedocs.io/en/latest/configuration.html#pre-installed-packages

@bollwyvl what is your opinion on this? Maybe something that could be more easily iterated on once Pyolite is moved to its own repo (https://github.com/jupyterlite/jupyterlite/issues/386).

jtpio:
Also the UI tests failure indicate some issue while executing a cell, which looks relevant:

image

bollwyvl:
I'll say it again, If stuff breaks at this level, it leaves the kernel in an unrecoverable statte, and you can never get out. _I_ don't want this repo to have to field support for those kinds of failures.

We can leave more hooks open to allow these things with additional extensions, but it should not be in core.

vasiljevic:
> Also the UI tests failure indicate some issue while executing a cell, which looks relevant:

Thank you @jtpio, I have overlooked that pyolite package depends on matplotlib and pillow since since pyolite need to patch matplotlib and PIL modules. I have reverted the commit 'Make matplotlib optional but included by default' and tests have passed now.

BTW, how can I access images and videos from UI test report on GitHub? I this particular case I could not reproduce the error since it is race condition related and tests had not failed in my local environment.

vasiljevic:
> If things move forward, it would be interesting to check how to unify this with what xeus-python offers: https://xeus-python-kernel.readthedocs.io/en/latest/configuration.html#pre-installed-packages

This feature is why I like xeus-python. But xeus-python do package installation in build time by using empack as opposed to Pyodide.  It would be nice if there was an easy way to create a custom Pyodide distribution and integrate it into pyolite build, just like it works for xeus-python. 

Before Pyodide offers an easy build of custom distributions, we can achieve similar behavior to `XeusPythonEnv.packages` by enhanced version of `LiteBuildConfig.piplite_urls` that additionally pre-installs appropriate packages in runtime. In this case a solution like one proposed in this PR may implement the runtime part of pre-installation. Then we can also unify configuration syntax for pre-installing packages in xeus-python and Pyodide, but to use different solutions under the hood.

vasiljevic:
> I'll say it again, If stuff breaks at this level, it leaves the kernel in an unrecoverable statte, and you can never get out. _I_ don't want this repo to have to field support for those kinds of failures.
> 
> We can leave more hooks open to allow these things with additional extensions, but it should not be in core.

I fully understand your concerns, but I just have not find better way to run package pre-installation. 

The mayor constraint is that pre-installation must happen before anything is executed in kernel. During the Pyolite kernel initialization a pending execution may occur already awaiting at [the beginning of the PyoliteRemoteKernel.execute() method](https://github.com/jupyterlite/jupyterlite/blob/631424941bc90e4d52e8920cb2dc0b0ef3ec2b49/packages/pyolite-kernel/src/worker.ts#L294).  The last statement in the Pyolite kernel initialization is [this._initializer?.resolve()](https://github.com/jupyterlite/jupyterlite/blob/631424941bc90e4d52e8920cb2dc0b0ef3ec2b49/packages/pyolite-kernel/src/worker.ts#L161). When that statement finishes, the pending execution starts immediately. Thus, we need to pre-install packages before  this._initializer?.resolve() .

We can put such configurable staff just before this._initializer?.resolve(), can properly catch all exceptions, etc, but looks like those staff cannot be placed after the kernel initialization.

yuvipanda:
With my deployer hat on, this is the absolute one single feature that means I *must* use XPython now for end users to have a decent experience, even if that means I've to spend a bunch of time adding packages to `conda-forge`. That's super un-ideal, as PyPI will (IMO) always be a superset of (python) packages available on conda-forge. I don't have enough of an understanding of how this should be implemented here, and understand the difficulties with error recovery. However, from a 'can end users actually use this?' perspective, this would go a *long* way in helping us actually deploy Pyodide (vs xeus-python)

jtpio:
> We can leave more hooks open to allow these things with additional extensions, but it should not be in core.

Maybe a good compromise would indeed be to leave some hooks. Something that could be done after https://github.com/jupyterlite/jupyterlite/pull/854

tomjakubowski:
Our site's build pulls some tricks to pre-install wheels - in short, at build time we boot up a pyodide interpreter with node, install the packages we need with piplite, and then call `micropip.freeze()` to generate a new repodata.json file which we replace in the final build.

> leave some hooks [for an addon]

I'm interested in working on this - are hooks for a third-party "pre-installer addon" still the way you all envision supporting this feature?

jtpio:
Thanks all for the discussion.

Support for `loadPyodide` landed in https://github.com/jupyterlite/pyodide-kernel/pull/103, which should help support this use case at some point. So it looks like we can close this PR for now, as this is related to the Pyodide kernel only.

@vasiljevic if you would like to continue exploring and discussing this in https://github.com/jupyterlite/pyodide-kernel, please feel free to open an issue or comment in https://github.com/jupyterlite/pyodide-kernel/issues/40. Thanks!

PR #821, opened 2022-09-27
Condition all addon tasks for well-formedness 

### LINKED ISSUE #820: Support of `=`-character in names of in provided assets
### Problem

When there is the notebook e.g. called `example=1.ipynb`, the GitLab pages build throws this error:
```console
contents:copy:example=1.ipynb
########################################
Execution aborted.
Task 'build:contents:copy:example=1.ipynb': name must not use the char '=' (equal sign).
Cleaning up project directory and file based variables
ERROR: Job failed: exit code 2
```

### Proposed Solution

The task file names are probably mungeable by replacing non-standard characters when [initializing the task](https://github.com/jupyterlite/jupyterlite/blob/main/py/jupyterlite/src/jupyterlite/addons/contents.py#L73) (idea from bollwyvl in https://github.com/jupyterlite/jupyterlite/issues/816).

bollwyvl:
## References

- fixes #820

## Code changes

- normalize all addon-generated task names to not include `=` (replace with `--`)
  - this is due to:
    - > = is used to pass global variables through the command line.
      > https://github.com/pydoit/doit/issues/117
- leaves options open for future fixes

## User-facing changes

- files with `=` in the name (such as `contents` or `archive` or `federated_extensions`, etc) will not fail

- n/a

### DIFF
--- a/py/jupyterlite/src/jupyterlite/addons/archive.py
+++ b/py/jupyterlite/src/jupyterlite/addons/archive.py
@@ -25,7 +25,7 @@ class ArchiveAddon(BaseAddon):
 
     def status(self, manager):
         tarball = manager.output_archive
-        yield dict(
+        yield self.task(
             name="archive",
             actions=[(self.log_archive, [tarball])],
         )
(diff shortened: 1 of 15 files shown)


jtpio (approved):
Thanks!

kolibril13:
Thanks, that was a very quick fix!

PR #832, opened 2022-10-12
Add pyolite pre-transformer, %pip

### LINKED ISSUE #237: Replacement for %pip
### Problem

1. The %pip that comes with IPython doesn't work with jupyterlite as there is no shell.
2. ipywidgets is not included, and requires using `await micropip.install()`. This is confusing for new users because it uses asyncio, and because it isn't pip

### Proposed Solution

If we override `%pip`, then we can install ipywidgets (and other packages) without having to use asyncio and micropip. This could make many notebooks 100% identical to their regular IPython brothers.

Here is an implementation (although this is my first asyncio code):

```python
from IPython.core.magic import register_line_magic
from IPython.core import magic_arguments
import asyncio

IN_BROWSER = None
try:
    import micropip
    IN_BROWSER = True
except ImportError:
    IN_BROWSER = False

if IN_BROWSER:
    async def pip_install(package):
        try:
            await micropip.install(package)
            return "    Success! Installed %r" % package
        except Exception as exception:
            return "    Failed installing %r: %s" % (package, exception)

    @register_line_magic
    @magic_arguments.magic_arguments()
    @magic_arguments.argument('-q', '--quiet', help='Give less output.', action='store_true')
    @magic_arguments.argument('command', help="Install.", type=str)
    @magic_arguments.argument('packages', nargs="+", type=str, help='Packages to install.')
    def pip(line):
        args = magic_arguments.parse_argstring(pip, line)
        if args.command == "help":
            print("usage: %pip install package1 package2 ...")
        elif args.command == "install":
            print("Installing %s..." % args.packages)
            loop = asyncio.get_running_loop()
            for package in args.packages:
                task = loop.create_task(pip_install(package))
                task.add_done_callback(
                    lambda t: print(t.result()))
        else:
            print("error: command should be: `install`")
```
If this is a viable solution, where would this code go?

Output examples:

Screenshot from 2021-07-08 15-54-38

bollwyvl:
## References

- fixes #237
- continues (and applies) notebook linter in #833

### `piplite`
- [x] adds a "cli" (really just for the argparse)
  - [x] `install`
    - [x] robustly handle errors during rewrite
      - [x] this stage should _never_ result in an unusable kernel
        - [x] though missing packages are probably fine, and could suggest using `-v` to trigger `keep_going`
    - [x] support `--no-deps`, `-q`, `-v` (as `keep_going`, today)
    - [x] support `-r`
       - [x] and nested `-r`

### `pyolite`
- [x] make `Pyolite` a `LoggingConfigurable`
  - [x] use some `traitlets.defaults` 
- [x] add a custom `Pyolite.lite_transform_manager: TransformerManager` that runs before the interepreter's
  - [x] `async` all over to leave options open for async transformations
    - for example, if `-r` could be a URL (can't with `pip`), would need this
  - [x] initialize with `pip_magic`
    - at present, this is a `line_transforms`, unaware of context
      - this _would_ work with `x = %pip install`
      - this means it could fail with the false positive
      ```python
      help = """
      install more packages with:

      %pip install nbformat
      """)
      ```
    - a `token_transformer`) would be more accurate, but has some more complexity, as much transform behavior is hard-coded

### `doit`
- [x] add a notebook linter with black and prettier
  - > _aside: `jupyterlab` should really just make a first-party `prettier-plugin-ipynb`_

## User-facing changes

- a user... 
  - ...can more conveniently install simple dependencies
  - ...can use their notebook in more environments
  - ...can add new lite-specific transformers interactively
    - ...or with packages that offer a  `%load_ext`
- a content creator 
  - ...can capture their site's build and run requirements across many notebooks in a
     - ... single `requirements.txt` with `; platform_machine != "wasm32"` in a single file
     - ... or multiple files the reference each other
     - ... and use this to lock down their deps with `pip-tools`, etc. for building
- a viewer of the docs site 
  - ... will see more about `%pip`, etc. in `pyolite/python-packages.ipynb`
  - ... see more copy-pasta examples

- probably none
  - any code using `await piplite.install` will still work
  - downstream kernels shouldn't be bothered by the traitlets stuff

## Future Work
- use `%pip` in `examples` #833
  - [x] replace `piplite.install` with `%pip install -q`
  - [x] blacken, pretty notebooks
- other `pip <action>`s
  - > while in `micropip`, not sure how portable these will be? also a large surface area
  - `list`
  - `freeze`


bollwyvl:
I've rolled back the wholesale changes to the example notebooks to make this easier to review, and moved those concrete changes to #833.

jtpio (approved):
Thanks!

PR #833, opened 2022-10-13
Lint example notebooks, use %pip magic

bollwyvl:
## References

- continues #832

## Code changes

- [x] replaces `await piplite.install` with `%pip`
- [x] locally, applies linter to notebooks
  - [x] `nbstripout` all outputs  
  - [x] `pretty` all markdown
  - [x] `black`en all python code
    - use `#fmt: off/on` in `ipycanvas`
  - [x] `nbconvert` for canonical JSON (see also #834)
- [x] remove `emscripten-filesystem.ipynb`
  - it's no longer very accurate
- [x] skips `.md` files in CI linting

## User-facing changes

- for a docs site reader
  - example notebooks will not contain (trusted) outputs
  - copy-pasta should be more canonical
- for reviewers
  - diffs should be cleaner

- n/a


jtpio:
Thanks @bollwyvl!

>  replaces `await piplite.install` with `%pip`

Sounds good. The `piplite` calls specific to JupyterLite have always been a bit awkward in notebooks, like in https://ipywidgets.readthedocs.io/en/stable/#try-it-online. At least `%pip` might look more familiar.

bollwyvl:
Yeah, i'll circle back up on these conflicts.

I guess at the high level:
- i think `black` is definitely worth keeping, and is fast and simple due to upstream support 
- do we feel good about stripping the outputs?
  - is it useful to have a few that are in there so we can evaluate whether they look right?
- do we feel good about  prettier?
  - i personally find it a bit easier to review changes to shorter lines (even with wrapping noise) vs some of the monster blocks.

jtpio:
> do we feel good about prettier?

Makes sense. The only drawback might be for folks contributing changes to the notebooks (like a typo fix or small edits) to have a local dev setup or they will likely get the linting wrong otherwise. Which is already the case when editing markdown files like README (for example in https://github.com/jupyterlite/jupyterlite/pull/787).

bollwyvl:
Linty fresh now after 45b48fe

jtpio (approved):
Thanks!

Issue #671, opened 2022-06-16
Mounted contents files always report 0 size, maybe other missing attributes


bollwyvl:
## Description

Some file attributes aren't getting propagated (correct or otherwise) from the contents API. This is easiest  noticeable with file sizes.

## References

- noted on #670
- follow-on to #655

## Reproduce

- upload a bunch of different kinds of files
- make a pyolite notebook 

    ```py   
    from pathlib import Path
    {p: p.stat().st_size for p in Path.cwd().glob("*")}
    ```

- all the files have a size of `0`

## Expected behavior

- files have sizes
- other attributes are correct

## Context

- JupyterLite version: #670
- Operating System and version: linux
- Browser and version: 102.0.5005.115 (Official Build) snap (64-bit)

martinRenou:
Assigning myself on this.

martinRenou:
So it looks like the following works-ish **if** we require the content:

https://github.com/jupyterlite/jupyterlite/blob/2974eaae5796cb7025f632b7aaa96c9799854c4f/packages/server-extension/src/index.ts#L307-L324

with:

```typescript
model = await contentManager.get(path, { content: true });
```

But the `model.size` is not set properly, it's either `undefined` or set with a weird (visibly wrong) value. I suspect this has to do with the IModel stored in the localforage.

bollwyvl:
Yes: for example, it seems that `base64` files are always `75b`.

I guess this was less of an issue when we just had to fool the browser, but "real" code (even if running on a "fake" computer) seems to be a bit pickier... but the payoff is huge.

martinRenou:
I wonder where this could be fixed. Should we compute the size when saving the IModel on localforage?

jasongrout:
Possibly related to this: I'm often seeing js errors using jupyterlite 0.1.0b9 in the console of the form:

```
Uncaught (in promise) TypeError: a.path is undefined
    onmessage http://localhost:8000/lite/build/4884.7d948f6.js:1
    activate http://localhost:8000/lite/build/4884.7d948f6.js:1
    promise http://localhost:8000/lite/build/6648.62de94d.js:2
    promise callback*16731/a</e.prototype.activatePlugin http://localhost:8000/lite/build/6648.62de94d.js:2
```
Tracking it down in the debugger, it seems to come from this line:
https://github.com/jupyterlite/jupyterlite/blob/4af0da94f820730d91261db472409abb108d1068/packages/server-extension/src/index.ts#L231
[...]

PR #839, opened 2022-10-16
Include some more accurate file stats from ServiceWorker

bollwyvl:
## References

- more work on #671

## Code changes

- report/save `size` in more contents places
  - even if `contents` is `False`, but we happen to have the content around
- more best-effort at calculating `IStats`
  - use the Emscripten `node.id` for inode
  - calculate blocks/blocksize
- add some more narrative and (counter) examples in `virtual_drive.ipynb`

## User-facing changes

- more low-level file options should work
  - sqlite still doesn't, and will need more investigation 

- n/a


martinRenou (approved):
LGTM!

We should probably do the same for the jupyterlab-filesystem-access and jupyterlab-github drives if we can.

bollwyvl:
> We should probably do the same for the jupyterlab-filesystem-access and jupyterlab-github drives if we can.

Does that happen on this repo, or the other side?

bollwyvl:
Created this follow-on #840

jtpio (approved):
Thanks!

PR #844, opened 2022-10-20
Bump various GitHub Actions, handle new `jupyter_core`

jtpio:
## References

Perform a bit of maintenance on the GitHub Actions to fix some of the following warning:

image

## Code changes

- [x] Bump various actions
- [x] Add `dependabot.yml` to make this process more automated: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#example-dependabotyml-file-for-github-actions
- [x] Handle new `jupyter_core`: https://github.com/jupyter/jupyter_core/commit/1118c8ce01800cb689d51f655f5ccef19516e283

## User-facing changes

None

None

### DIFF
--- a/py/jupyterlite/src/jupyterlite/app.py
+++ b/py/jupyterlite/src/jupyterlite/app.py
@@ -1,7 +1,7 @@
 """the JupyterLite CLI App(s)"""
 from pathlib import Path
 
-from jupyter_core import jupyter_config_path
+from jupyter_core.paths import jupyter_config_path
 from jupyter_core.application import JupyterApp, base_aliases, base_flags
 from traitlets import Bool, Instance, Unicode, default
 
--- a/py/jupyterlite/src/jupyterlite/app.py
+++ b/py/jupyterlite/src/jupyterlite/app.py
@@ -87,7 +87,7 @@ class BaseLiteApp(JupyterApp, LiteBuildConfig, DescribedMixin):
 
     @default("config_file_paths")
     def _config_file_paths_default(self):
-        return [os.getcwd()] + jupyter_config_path()
+        return [str(Path.cwd())] + jupyter_config_path()
 
 
 class ManagedApp(BaseLiteApp):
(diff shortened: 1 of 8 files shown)


jtpio:
Looks like the CI failures need proper investigation (also failing on `main`).

jtpio:
`jupyter_core==4.11.2` seems to be cause of the failures. So probably related to this commit: https://github.com/jupyter/jupyter_core/commit/1118c8ce01800cb689d51f655f5ccef19516e283

jtpio:
Most deprecation warnings should now be gone (still a couple of left, will have a look another time).

py/jupyterlite/src/jupyterlite/app.py
```diff
@@ -80,6 +85,10 @@ class BaseLiteApp(JupyterApp, LiteBuildConfig, DescribedMixin):
 
     flags = lite_flags
 
+    @default("config_file_paths")
+    def _config_file_paths_default(self):
+        return [str(Path.cwd())] + jupyter_config_path()
```
bollwyvl, line 90: yep, let's get rid of cwd.

bollwyvl: it'll probably break some tests/binder, though.

jtpio: ok let's do the change in its own PR so it's more visible in the changelog.

I guess that means `--config jupyter_lite_config.json` should now be used explicitly.

bollwyvl: ugh. a good point. i guess we need to think about it a bit more: running `jupyter lite build` from a CLI is a different activity altogether than browsing a web page.

bollwyvl (approved):

jtpio:
Following up with a new release since new deployments are likely affected by the latest `jupyter_core` release.

PR #848, opened 2022-10-25
Update to JupyterLab 3.5

jtpio:
## References

Mechanical update to the latest JupyterLab 3.5 release: https://github.com/jupyterlab/jupyterlab/releases/tag/v3.5.0

## Code changes

Dependency update

## User-facing changes

Some from: https://github.com/jupyterlab/jupyterlab/releases/tag/v3.5.0. To check:

- [x] Prompt for renaming at first manual save
- [x] File Browser: add support for filtering directories on search

None


bollwyvl (approved):
LGTM! I'd also update the appropriate `environment.yml` and `requirement.txt`s to pull in the same version as we're using here.

PR #852, opened 2022-10-27
Update app resolutions for JupyterLab 3.5.0

bollwyvl:
## References

- continues #848

## Code changes

- Runs `doit lint repo` to update all the resolutions
  - maybe if this doesn't have any effect, we don't need these in the files?

## User-facing changes

- ?

- ?


jtpio (approved):
Thanks for catching this!

bollwyvl:
> should be run somewhere on CI to detect when these go out of sync.

K, seems appropriate on this PR, will see if i can do it quickly.

jtpio:
Thanks!

At some point it would be nice to have a more robust integrity check like in JupyterLab. Need to check whether it would be possible to reuse some of the JupyterLab `buildutils`.

bollwyvl:
Thanks!

PR #854, opened 2022-10-27
Move the Pyodide kernel to a separate repo

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/386
Fixes https://github.com/jupyterlite/jupyterlite/issues/116

Opening early to get an clearer idea of potential issues. Thinking about it more, it would also be nice to make that change before releasing a final 0.1.0.

## Code changes

- [x] Remove `pyolite` and `pyolite-extension` packages from the `jupyterlite` code base
- [x] Create the new repo: https://github.com/jupyterlite/pyodide-kernel
- [x] Extract the pyolite kernel to https://github.com/jupyterlite/pyodide-kernel while preserving history: https://github.com/jupyterlite/pyodide-kernel/pull/2
- [x] Clean up `dodo.py`, might help with https://github.com/jupyterlite/jupyterlite/issues/684
- [x] Add `jupyterlite-pyodide-kernel` as a federated kernel to the demo site
- [x] ~Should the pyodide kernel be shipped by default in JupyterLite? (similar to `ipykernel` being a dependency of `jupyterlab`)~
  -  ~or maybe just documenting `pip install jupyterlite-pyodide-kernel` before running `jupyter lite build` will be enough~
  - -> to be handled separately with the creation of `jupyterlite-core` and using `jupyterlite` as metapackage
- [x] Check how to handle `litePluginSettings` -> should not require any changes
- [x] ~How to handle examples and the demo RTD site?~ -> leaving in the main repo for now since the Pyodide will be available on the demo site
- [x] https://github.com/jupyterlite/jupyterlite/pull/934
- [x] We can keep the Pyodide related configuration and documentation in this repo for now so it's more visible (to be reconsidered)
- [x] include `dist` in the RTD build so it can be consumed downstream
- [x] add `html_theme_options` for newer `pydata-sphinx-theme`
  - [x] `icon_links`
  - [x] `pygment_(light|dark)_style` 

### Follow-ups

- [ ] Move Pyodide related issues to https://github.com/jupyterlite/pyodide-kernel

## User-facing changes

None if the Pyodide kernel is installed as a federated kernel.

The assumption that the Pyodide is always available will not be true anymore.


bollwyvl:
> Should the pyodide kernel be shipped by default in JupyterLite

I don't think it can, directly, as `jupyterlite-pyodide-kernel` must carry a dependency on `jupyterlite` itself.

The other option would be to offer `jupyterlite_core` (`jupyterliter`?) with the `jupyterlite` package being a metapackage that included the tested `jupyter-` first-party offerings (perhaps including `ipywidgets`).

This would also beg the question of eventually splitting out the `apps` as separate packages... though they still need to be built as part of The One True WebPack build to maintain the deduplication properties. But it would put a firm stake in the ground as to how to make sub-packages, which `voila` will start pushing boundaries on.

> Check how to handle `litePluginSettings`

Yep: this will be a bit trickier.

I've been thinking for a while that we need to store all of the `jupyter-lite.json` contributions as JSON fragments, and join them all together at the very end.

jtpio:
> While this is still a good idea, I'm taking a look at updating to the 0.22.x line... tightening up how that gets updated/tested will be nice on the new repo.

Sounds good, we can make a new lite release with the new pyodide before continuing the move to a different repo.

bollwyvl:
I don't think it's really blocking one way or the other... Really just how
people's time works out, I guess.

There's enough stuff in the 0.22 PR that would work _now_, such that I
could take that version bump out and we'd still be better off... The
blocker there was really hoping to get a handle on that SQLite thing, even
as pyodide will be taking it out of the up-front download. Getting the
service worker totally solid seems like a pretty firm need for us to feel
good about 0.1.

jtpio:
Most CI checks are now passing except for the UI tests (expected since the Pyodide kernel is now missing).

I'll be continuing with the extraction to the separate repo.

py/jupyterlite/src/jupyterlite/app.py
```diff
@@ -82,9 +80,6 @@ class BaseLiteApp(JupyterApp, LiteBuildConfig, DescribedMixin):
             # server-specific things
             "port": "LiteBuildConfig.port",
             "base-url": "LiteBuildConfig.base_url",
-            # pyolite things likely to move away
-            "piplite-wheels": "LiteBuildConfig.piplite_urls",
```
bollwyvl, line 86: so this is big-break town. I think we need to solve this issue, potentially on a separate PR first. I'll start an issue...

bollwyvl: Created https://github.com/jupyterlite/jupyterlite/issues/929

bollwyvl:
As mentioned above, created https://github.com/jupyterlite/jupyterlite/issues/929: if we think this is a good way to proceed, I can work up a PR, or we can discuss alternatives.

bollwyvl:
#934 is up, which might create some small conflicts.

Question remains there whether we do a release with a shim for `pyodide_url` and `piplite_urls` in `LiteBuildConfig`, or just let stuff fail.

jtpio:
I think most of the removal work is done here in the `jupyterlite/jupyterlite` repo.

We basically need to finish https://github.com/jupyterlite/pyodide-kernel/pull/2 so it builds, tests, and so on. Then publish `jupyterlite-pyodide-kernel` as a federated kernel so we can include it on this repo.

The checklists in https://github.com/jupyterlite/jupyterlite/pull/854#issue-1425925563 and https://github.com/jupyterlite/pyodide-kernel/pull/2#issue-1472543152 should still be accurate though.

jtpio:
Appears to be fine on RTD:

image

Will be looking into the UI tests.

jtpio:
> with the `jupyterlite` package being a metapackage that included the tested `jupyter-` first-party offerings (perhaps including `ipywidgets`).

Right I guess it would be fine to offer such a metapackage. But the addon base classes would then have to be moved to the other package (let's say `jupyterlite-core`). This would also require the federated kernels to import `BaseAddon` (and other things) from `jupyterlite_core` instead, and depend on `jupyterlite_core` instead of `jupyterlite`.

Thinking of other options as alternative:

1. Just like `jupyterlite` bundles a tarball of the main application by default, maybe we could (temporarily) also bundle the tarball for `jupyterlite-pyodide-kernel`. That way the Pyodide kernel is still distributed by default for now and does not break existing deployments for 0.1.0.
  - downsides:
    - requires updating the version of the kernel here in `jupyterlite` and make a new release for folks
  - advantages:
    - the assumption there is still a default Python kernel available after a vanilla `jupyter lite build` would still hold true. Similar to the current dependency on `ipykernel` in the Jupyter stack
2. `jupyterlite` does not ship any other kernel by default anymore (or just the JavaScript kernel) and there is no dependency on `jupyterlite-pyodide-kernel`. It feels like this should eventually happen, although doing it right now might be quite sudden and we may want to give users proper notice. Something to consider for 0.2.0 for instance.

bollwyvl:
As a metapackage, it's _still_ a package, and could shim the public parts of the API for the rest of the 0.1.0 line, with a single warning, even if the implementations and CLI are still part of this repo. And the releases would get cut from this repo... initially, anyway. As a first demonstrator for this split, we can have `jupyterlite_pydodide_kernel` use the new importable locations, and folk _would_ be able to use `jupyterlite_core` independently, already.

As to downstreams: well, we never made many guarantees! The number of public uses in the field is discoverable and countable, though, so we could even have a "kitchen sink" repo that tried to install all the kernels, but again, I really don't think the shim would be so bad. 

If folk have private stuff that we break, not our problem. 

If folk have already hard forked this repo, again, not our problem, and it won't impact them anyway.

Looking forward, `0.2.0ax` (based on lab 4) remove the pyodide kernel as a hard dep.

jtpio:
I just did a quick check of the example notebooks available on the RTD preview and they seem to be working fine: https://jupyterlite--854.org.readthedocs.build/en/854/_static/lab/index.html.

So it looks like a good next step would be to merge this PR, and proceed with `jupyterlite-core` and the metapackage in a follow-up PR.

examples/jupyter_lite_config.json
```diff
@@ -21,7 +21,8 @@
       "https://conda.anaconda.org/conda-forge/noarch/jupyterlab_miami_nights-0.3.2-pyhd8ed1ab_0.tar.bz2",
       "https://conda.anaconda.org/conda-forge/noarch/jupyterlab_widgets-3.0.5-pyhd8ed1ab_0.conda",
       "https://conda.anaconda.org/conda-forge/noarch/theme-darcula-3.1.1-pyh3684270_0.tar.bz2",
-      "https://github.com/jupyterlite/p5-kernel/releases/download/v0.1.0/jupyterlite_p5_kernel-0.1.0-py3-none-any.whl"
+      "https://github.com/jupyterlite/p5-kernel/releases/download/v0.1.0/jupyterlite_p5_kernel-0.1.0-py3-none-any.whl",
+      "https://github.com/jupyterlite/pyodide-kernel/releases/download/v0.0.2/jupyterlite_pyodide_kernel-0.0.2-py3-none-any.whl"
```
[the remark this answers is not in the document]
jtpio: So the demo site available on https://jupyterlite.rtfd.io/en/latest/try/lab is pretty much the same as before.

jtpio:
We can think about what to do with the Pyodide specific docs later. For now they are still more discoverable on https://jupyterlite.readthedocs.io, but we might want to restructure slightly to leave space for other kernels (#933)

bollwyvl (approved):
LGTM as it will help us move forward. there is still a lot to do, and i feel we're a fairly long way off from a release, but with e.g. publishing the wheels on RTD, we at least have a way forward for assessing the impact on downstreams.

jtpio:
> fairly long way off from a release

Well now that we have https://github.com/jupyterlite/pyodide-kernel and can iterate on it and release it more easily, we should be able to just do https://github.com/jupyterlite/jupyterlite/issues/993 and release a new `0.1.0b19` (for both `jupyterlite-core` and `jupyterlite`) to continue towards 0.1.0.

I'm afraid that if we don't do this fast enough we'll still be stuck in pre-releases for a long time.

PR #878, opened 2022-11-25
Prefer `libarchive-c` if available for unarchiving, support `.conda` packages

### LINKED ISSUE #877: Allow extracting federated extensions from `.conda` packages
### Problem

`conda-forge` has transitioned to only building `.conda` packages. 

### Proposed Solution

Expand the federated extensions logic to look inside the double-wrapped archive. Old packages will not be transmuted, so we need to handle both.

### Additional context

May require a new (optional) dependency for dealing with `zstd` files.

bollwyvl:
## References

- fixes #877

## Code changes

- prefers `libarchive-c` if available over stdlib
- always unpacks full `federated_extension` sources locally

## User-facing changes

- newer `.conda` packages can be used as sources for federated extensions
- builds _should_ be faster (need to do some more benchmarking)

- n/a

## Future work

the `archive` addon still uses stdlib, and is quite slow. I had some issues generating reproducible tarballs between `libarchive-c` and `gzip+tarfile`, so will leave that for another day.


jtpio:
Thanks @bollwyvl.

Do you have an example of a package already using the `.conda` extension?

Also does that mean we'll eventually be able to switch to this new format for the `federated_extensions` of the demo site?

bollwyvl:
> Do you have an example of a package already using the .conda extension?

there's the littlest-extension in the fixture: i generated that with `cph transmute <the old one> .conda` or something.

> Also does that mean we'll eventually be able to switch to this new format for the federated_extensions of the demo site?

`conda-forge` won't be back-filling any `.conda` packages, but unless configured otherwise, all _new_ packages as of last week will be `.conda`, so really we just need _any_ new release of a package. I can try a fresh env, see if we get some.

bollwyvl:
Nope: resolve isn't doing it. Would love a non-jank, `console.error`-free `jupyterlab-github`, maybe i'll try to get something going over there.

py/jupyterlite/src/jupyterlite/addons/federated_extensions.py
```diff
@@ -122,137 +127,112 @@ def resolve_one_extension(self, path_or_url, init):
...
-
-        def _filter_members(infos):
-            return [
-                p
-                for p in infos
-                if not p.isdir() and p.name.startswith(pkg_root_with_slash)
-            ]
-
-        targets = [dest / m.name[prefix:] for m in _filter_members(all_infos)]
+    def copy_conda2_extensions(self, conda_pkg):
```
jtpio, line 196: Should this method still be called `copy_conda_extensions`? Or does the `2` refer to something in particular?

bollwyvl: Yeah, it's referencing that fact that under the hood, these are built (currently) only when you specify [`pkg_format: 2`](https://conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#specify-conda-build-package-version-pkg-version)

bollwyvl:
Just merged https://github.com/conda-forge/ipydatagrid-feedstock/pull/23 (which would build a `.conda`), so we could add `ipydatagrid` and an example. It is also, of course, _sweet as hell_.

py/jupyterlite/src/jupyterlite/addons/federated_extensions.py
jtpio: ok thanks :+1:

jtpio:
Nice!

We can indeed see the new `.conda` here: https://anaconda.org/conda-forge/ipydatagrid/files

image

> we could add ipydatagrid and an example

Sounds good, we can also do this in a separate PR.

jtpio (approved):
Thanks!

PR #871, opened 2022-11-11
[wip] Update to pyodide 0.22.0, use pyodide and serviceworker types

bollwyvl:
## References

- could be moved to new repo if #854 lands first
- fixes #792 (it wasn't really broken)
- investigates #751

## Code changes

- [x] add exact-pinned `pyodide` to `pyodide-kernel/package.json#/devDependencies`
  - only ever `import type`d in `.ts`
  - is the source of truth for all other versions
- [x] bump to [`pyodide 0.22.0a3`](https://github.com/pyodide/pyodide/blob/0.22.0a3/docs/project/changelog.md#unreleased)
- [x] remove duplicated `ERRNO_CODES`
- [x] remove use of private `pyodide._module`
- [x] update key dev deps
  - [x] `lerna 6`
  - [x] `typedoc`
  - [x] `typescript`
  - [x] `prettier`    
- [x] refactor `BroadcastChannel` out of server, put in contents
- [x] make pyolite more resilient to missing `ServiceWorker` and `BroadcastChannel` 
- [x] moves `services.js` into `server/sw.ts`
  - [x] add a new webpack plugin to hoist it up to the root of the app (or the scope is wrong) and add the content hash to the file name
- [x] add `SKIP_ESLINT` env var
  - > the pre-commit hooks still fail sometimes, so `git commit -n` it shall be!

## User-facing changes

- potentially better error messages on startup if Things Go Wrong

## Backwards-incompatible changes

- n/a
  - the `sqlite3` change might have been surprising...
    - but we'll always be loading it so `IPython` can load

PR #892, opened 2022-12-07
Service worker fixes, JupyterLab 3.5.1

bollwyvl:
## References

- #871, but without the `pyodide` version change

## Code changes

- [x] add exact-pinned `pyodide` to `pyodide-kernel/package.json#/devDependencies`
  - only ever `import type`d in `.ts`
  - is the source of truth for all other versions
- > ~~bump to [`pyodide 0.22.0a3`](https://github.com/pyodide/pyodide/blob/0.22.0a3/docs/project/changelog.md#unreleased)~~
- [x] add `@types/emscripten` (such as it is)
- [x] remove duplicated `ERRNO_CODES`
- [x] remove use of private `pyodide._module`
- [x] update key dev deps
  - [x] `lerna 6`
  - [x] `typedoc 0.23`
  - [x] `typescript 4.9`
  - [x] `prettier 2.8`    
- [x] refactor `BroadcastChannel` out of server, put in contents
- [x] make pyolite more resilient to missing `ServiceWorker` and `BroadcastChannel` 
- [x] moves `services.js` into `server/sw.ts`
  - [x] add a new webpack plugin to hoist it up to the root of the app (or the scope is wrong) and add the content hash to the file name
- [x] add `SKIP_ESLINT` env var
  - > the pre-commit hooks still fail sometimes, so `git commit -n` it shall be!
- [x] remove `plotly` until compatible with ipywidgets 8
  - > https://github.com/plotly/plotly.py/issues/3686 

## User-facing changes

- site deployers
  - can disable the `ServiceWorker`
- end users
  - should get fewer stale serves of `services.js` with cache-busting

- some Emscripten-related types and classes have moved
- the name of `services.js` is no longer predictable (hash-based)


packages/contents/src/drivefs.ts
```diff
@@ -321,8 +345,8 @@ export class ContentsAPI {
     baseUrl: string,
     driveName: string,
     mountpoint: string,
-    FS: any,
-    ERRNO_CODES: any
+    FS: Pyodide.PyodideInterface['FS'],
```
martinRenou, line 348: Is this interface really bound to Pyodide?

We're using this DriveFS in xeus-python-kernel (and we could technically use it with another language kernel!) and I wonder if this interface does not make it too specific.

bollwyvl: Hm, having some issues with `@types/emscripten`...

bollwyvl: The pyodide ones aren't very strong yet, either:

```ts
export declare type PyodideInterface = {
	FS: typeof FS;
	...
	ERRNO_CODES: typeof ERRNO_CODES;
};
...
declare let FS: any;
...
declare let ERRNO_CODES: any;
```

packages/pyolite-kernel/src/worker.ts
```diff
@@ -1,136 +1,12 @@
...
-// Remove this shameless copy
-const ERRNO_CODES = {
-  E2BIG: 1,
-  EACCES: 2,
-  EADDRINUSE: 3,
-  EADDRNOTAVAIL: 4,
-  EADV: 122,
-  EAFNOSUPPORT: 5,
-  EAGAIN: 6,
-  EALREADY: 7,
```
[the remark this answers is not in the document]
bollwyvl: love a good meme, but what's your thoughts here? I don't see an easy way to get these (typed or otherwise) other than (at present) the pyodide declarations. We can keep them duplicated, of course.

packages/contents/src/drivefs.ts
bollwyvl: hmmm... this works, but `ERRNO_CODES` is not hoisted,

```ts
// eslint-disable-next-line @typescript-eslint/triple-slash-reference
/// <reference path="../../../node_modules/@types/emscripten/index.d.ts" />
export type EmscriptenFS = typeof FS;
...
FS: EmscriptenFS
```

packages/pyolite-kernel/src/worker.ts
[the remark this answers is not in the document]
martinRenou: Those ERRNO_CODE were not defined in `Pyodide<0.21.0`, it has been fixed in https://github.com/pyodide/pyodide/pull/2582. I really didn't like having those ERRNO_CODES values hardcoded here.

I guess we can assume they are defined now.

packages/contents/src/drivefs.ts
bollwyvl: Yep, guess it's true they're not providing a bunch of value.

I guess maybe the best way: we declare our own `any` in `@jupyerlite/types`, which we could fill in in the future. Then at least all of our references are consistent.

bollwyvl: because there were some constants, i ended up just keeping them in `contents` for now. I also did bring `@types/emscripten` in, which at least gives us partial coverage.

bollwyvl:
Didn't really intend to update lab, but it kinda came along for the ride with the `yarn.lock` updates, so figured would move forward.

martinRenou:
Did you mean to close it?

bollwyvl:
whoops! too many tabs! I'll see if i can re-open it!

bollwyvl:
i don't see any way in the UI... will start another :crying_cat_face:

PR #876, opened 2022-11-25
Update to JupyterLab 3.6

jtpio:
## References

Trying to update to the JupyterLab 3.6 packages before the first beta release to catch potential upstream breaking changes.

## Code changes

- [x] Update `@jupyterlab` packages
- [x] Update Python packages
- [ ] `@jupyterlab/shared-models` extraction: https://github.com/jupyterlab/jupyterlab/pull/13389
- [ ] More updates?

## User-facing changes

Some from JupyterLab 3.6 (to be posted here later).

## Backwards-incompatible changes

Should be none (TBC).

PR #898, opened 2022-12-08
Bump to JupyterLab 3.6.0

fcollonval:
This reverts commit 06c55f0a7320f84c2cffeaf637e69658ef8ef8a8.

## References

Superseeds https://github.com/jupyterlite/jupyterlite/pull/876

## Code changes

Upgrade to 3.6.0 alpha 5

## User-facing changes

Use latest version


martinRenou:
This probably goes in conflict with https://github.com/jupyterlite/jupyterlite/pull/899 (also in conflict with `main` apparently)

fcollonval:
I fully agree with @jtpio 

With the nice drive approach targeting 4 and as our resources are sparse, I advice to skip touching https://github.com/jupyterlite/jupyterlab-webrtc-docprovider before 4. Hence this implies not upgrading JupyterLite to 3.6.

bollwyvl:
This works for me! I guess it will then be relevant to continue getting upstream bugfixen to a `3.5.x` branch, as well for a while, then... we can't be the only downstream that will be unable to quickly shift and throw a bunch of resources at the new `ydoc`-related upgrades when `4.x` and `notebook 7.x` are coming Real Soon Now :tm:

jtpio:
Closing in favor of https://github.com/jupyterlite/jupyterlite/pull/1019 as we'll skip the update to JupyterLab 3.6.

Issue #826, opened 2022-10-04
Update to JupyterLab 4 and Notebook 7


jtpio:
### Problem

There is no more development happening on RetroLab: https://github.com/jupyterlab/retrolab#%EF%B8%8F-retrolab-and-notebook-v7-%EF%B8%8F

Also JupyterLab 4 and Notebook 7 will be sibling releases and their final versions are expected in the coming months.

### Proposed Solution

After `0.1.0` is released (https://github.com/jupyterlite/jupyterlite/issues/383), we can start updating the `@jupyterlab` packages to 4.0 and the `@jupyter-notebook` packages to 7.0.

#### TODO

- [ ] Reuse upstream Playwright config, fix TODO added in https://github.com/jupyterlite/jupyterlite/pull/923
- [ ] Check https://github.com/jupyterlite/jupyterlite/pull/457
- [ ] Check or close https://github.com/jupyterlite/jupyterlite/pull/898

### Additional context

This is likely to land in a `0.2.0` (or later) release of JupyterLite.

bollwyvl:
Starting a `next` branch already might be interesting... though I don't know who would have a ton of time to work on it at the moment with everything else going on.

jtpio:
Right, maybe that's why keeping `main` on the latest versions would be easier w.r.t maintenance.

Nothing urgent yet, just so we have this on the radar.

jtpio:
Also all the RTC changes moving more logic to the backend (with ypy, jupyter_ydoc and so on) might require quite a lot changes here as well...

Unless most of it can be done in https://github.com/jupyterlite/jupyterlab-webrtc-docprovider directly.

So starting to update to the lab 4 and notebook 7 packages at some point soon will hopefully help catch issues that can be reported and fixed upstream before the final releases.

bollwyvl:
Looks like we'll get a preview of the pain involved in a much shorter timeframe with https://github.com/jupyterlab/jupyterlab/issues/13221

jtpio:
Thinking we could then get JupyterLite 0.1.0 final released based on JupyterLab 3.5 and RetroLab.

And skip 3.6 to go with lab 4 / nb 7 directly on `main` for the `0.2.0` pre-releases.

bollwyvl:
[...]

PR #899, opened 2022-12-08
Service worker fixes,  JupyterLab 3.5.1 

bollwyvl:
## References

- #871, but without the `pyodide` version change
- #892, but not closed

## Code changes

- [x] add exact-pinned `pyodide` to `pyodide-kernel/package.json#/devDependencies`
  - only ever `import type`d in `.ts`
  - is the source of truth for all other versions
- > ~~bump to [`pyodide 0.22.0a3`](https://github.com/pyodide/pyodide/blob/0.22.0a3/docs/project/changelog.md#unreleased)~~
- [x] add `@types/emscripten` (such as it is)
- [x] remove duplicated `ERRNO_CODES`
- [x] remove use of private `pyodide._module`
- [x] update key dev deps
  - [x] `lerna 6`
  - [x] `typedoc 0.23`
  - [x] `typescript 4.9`
  - [x] `prettier 2.8`    
- [x] refactor `BroadcastChannel` out of server, put in contents
- [x] make pyolite more resilient to missing `ServiceWorker` and `BroadcastChannel` 
- [x] moves `services.js` into `server/sw.ts`
  - [x] add a new webpack plugin to hoist it up to the root of the app (or the scope is wrong) and add the content hash to the file name
- [x] add `SKIP_ESLINT` env var
  - > the pre-commit hooks still fail sometimes, so `git commit -n` it shall be!
- [x] remove `plotly` until compatible with ipywidgets 8
  - > https://github.com/plotly/plotly.py/issues/3686 

## User-facing changes

- site deployers
  - can disable the `ServiceWorker`
- end users
  - should get fewer stale serves of `services.js` with cache-busting

- some Emscripten-related types and classes have moved
- the name of `services.js` is no longer predictable (hash-based)


martinRenou:
Headsup that this will go in conflict with https://github.com/jupyterlite/jupyterlite/pull/898

bollwyvl:
Not worried too much about that given the breaking changes in 3.6 anyway.
We may have to keep 3.5 running for a while

stevejpurves:
couple of questions:

How can a site deployer disable the service worker? should we expect to be able to set a new option in the `jupyter-config-data`? 

> ... and add the content hash to the file name
So this is the content hash of the `sw.ts` file and will remain stable across releases/builds otherwise?

bollwyvl:
> How can a site deployer disable the service worker?

added some docs here: [ada7479](https://github.com/jupyterlite/jupyterlite/pull/899/commits/ada7479b433cc87228b4897bd3203488ccd457cd)

>  the content hash of the sw.ts file and will remain stable across releases/builds otherwise?

I believe it's the hash of the _output_, which isn't precisely deterministic (or has rather, has many independent variables of which the source of `sw.ts` is only one). A new `sw-{hash}.js` will only (but very possibly) change with a new release of `jupyterlite 0.1.0b{n+1}`. Two successive runs of `jupyter lite build` without `pip install`ing something will always have the same `sw-{hash}.js`.

packages/server/src/sw.ts
```diff
@@ -0,0 +1,154 @@
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
```
jtpio, line 1: Maybe we could rename this file from `sw.ts` to `serviceworker.ts` so it's more explicit?

And the existing `serviceworker.ts` to `serviceworkerwrapper.ts`?

bollwyvl: perhaps `serviceworker.ts` and `servicemanager.ts`?

bollwyvl: 3a39e7f3 went with `ServiceManager`, etc

jtpio: Thanks. Although `ServiceManager` can probably be a bit confusing since there is already a `ServiceManager` in JupyterLab?

jtpio: `ServiceWorkerManager` would be fine I guess?

bollwyvl: _Excellent_ point, updated in 0bf86c21.

qqdaiyu55:
@jtpio With this upgrade, we should include your fix: https://github.com/jupyterlab/jupyterlab/pull/12849.
However I checked the bundle file `package/build/7941.dc36874.js` of jupyterlite v0.1.0-b16, it does not include your change. Did I miss anything here?
image

jtpio:
@qqdaiyu55 I think https://github.com/jupyterlab/jupyterlab/pull/12849 was not backported to 3.5 so it's not available in JupyterLite.

We should however be able to pick it up when updating to the JupyterLab 4 packages in https://github.com/jupyterlite/jupyterlite/issues/826.

qqdaiyu55:
@jtpio Thanks for reply. I'm implementing a custom drive and try to get it work for jupyterlite, and see the same issue reported here: https://github.com/jupyterlab/jupyterlab/pull/12755.

From my understanding, for any custom drive extension (e.g. jupyterlab-github), I should see the same renaming error when I opened a new document, if we didn't pick the [fix](https://github.com/jupyterlab/jupyterlab/pull/12849) into jupyterlite. However, I opened a document from github extension with latest https://jupyterlite.readthedocs.io/en/latest/_static/lab/, and there is no such error info.

Issue #897, opened 2022-12-08
delay kernel ready status


stevejpurves:
### Problem
When starting a `jupyterlite` server using the `pyolite` kernel, a client receives a `kernel-ready` message before `pyolite` initialization has finished. Even though the kernel blocks(awaits) it's ready promise before executing, there is no way for the frontend to signal that the kernel is not yet ready.

### Proposed Solution

Delay the ready signal until `pyodide` iniitialisation completes

### Additional context

related issue: https://github.com/executablebooks/thebe/issues/580

jtpio:
Closing as fixed by https://github.com/jupyterlite/jupyterlite/pull/900

PR #900, opened 2022-12-08
do not signal `ready` until initialize completes

stevejpurves:
## References

Addresses issue: https://github.com/jupyterlite/jupyterlite/issues/897

## Code changes

Previously the `_ready` promise for the kernel was immediately resolved without awaiting the initialization promise to resolve, now the `_ready` promise resolved after initialization completes.

The effect of this on a frontend like thebe, is that the UI can better signal when the kernel is actually ready -- aligning with behaviour seen with normal jupyter server connections.

## User-facing changes

In jupyterlite apps and thebe, like kernel ready indicator appears later, and shows a status `initializing...` during the package load. Which is the expected user facing change 👍

none expected

### DIFF
--- a/packages/kernel/src/kernels.ts
+++ b/packages/kernel/src/kernels.ts
@@ -63,6 +63,7 @@ export class Kernels implements IKernels {
 
       const processMsg = async (msg: KernelMessage.IMessage) => {
         await mutex.runExclusive(async () => {
+          await kernel.ready;
           await kernel.handleMessage(msg);
         });
       };
@@ -143,8 +144,6 @@ export class Kernels implements IKernels {
       name,
       location,
     });
-
-    await kernel.ready;
 
     this._kernels.set(kernelId, kernel);
     this._kernelClients.set(kernelId, new Set<string>());
--- a/packages/pyolite-kernel/src/kernel.ts
+++ b/packages/pyolite-kernel/src/kernel.ts
@@ -25,7 +25,6 @@ export class PyoliteKernel extends BaseKernel implements IKernel {
     this._worker = this.initWorker(options);
     this._worker.onmessage = (e) => this._processWorkerMessage(e.data);
     this._remoteKernel = this.initRemote(options);
-    this._ready.resolve();
   }
 
   /**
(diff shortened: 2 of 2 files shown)


martinRenou (approved):
Thanks!

jtpio:
Thanks @stevejpurves :+1: 

Looks like the UI tests failure might be relevant?

https://user-images.githubusercontent.com/591645/206654265-2db5ae3b-b894-4e61-b176-63350232a509.mp4

Haven't looked into it more, but maybe it's because the UI tests try to run cells before the kernel is ready? (since it should now take more time to be ready)

stevejpurves:
yes it very probably is, I'll look into it

bollwyvl (approved):

bollwyvl:
> try to run cells before the kernel is ready

we may need to enqueue pre-`ready` messages so they are sent in the right order.

jtpio:
Executing cells with `Shift-Enter` on the latest `main` (https://jupyterlite.rtfd.io/en/latest/try/lab), it looks like the messages are "queued" (`[*]`).

Maybe an extra `await kernel.ready` is needed somewhere (just a guess).

jtpio:
The last commit seems to be fixing it, and now the Pyodide kernel startup workflow looks a bit nicer:

https://user-images.githubusercontent.com/591645/208923115-8f75d9aa-4536-4211-b27e-9752396ea1c1.mp4

jtpio:
Normally JupyterLab already handles slow starting kernels which is why the `unknown` status shows up in the kernel status indicator:

- https://github.com/jupyterlab/jupyterlab/pull/8174
- https://github.com/jupyterlab/jupyterlab/pull/11358

Ideally we'll probably want the status indicator to show "Kernel Initializing" instead, but we could also investigate this separately.

martinRenou:
> and now the Pyodide kernel startup workflow looks a bit nicer:

That's great!! Thanks for doing it!

jtpio (approved):
Thanks!

stevejpurves:
thanks @jtpio - the change in the last commit makes sense to me now!

PR #911, opened 2022-12-12
Pyolite - Matplotlib: Setup inline backend by default

### LINKED ISSUE #906: JupyterLite doesn't display plt plots by default
## Description

This bug was originally found while investigating this PyScript issue: https://github.com/pyscript/pyscript/issues/983

As discussed with @jtpio in real life, the following piece of code behaves differently in Jupyter and JupyterLite: in Jupyter, it displays a plot; in JupyterLite, it displays `<matplotlib.collection.PathsColletion ...>`.

If you add `%matplotlib_inline`, it works as expected.

## Reproduce

```python
import pandas as pd
import matplotlib.pyplot as plt
url = "https://raw.githubusercontent.com/Cheukting/pyscript-ice-cream/main/bj-products.csv"

# try to autodetect whether it's jupyter or jupyterlist
try:
    from pyodide.http import open_url
except ImportError:
    print("Normal jupyter")
else:
    print("jupyterlite!")
    url = open_url(url)

df = pd.read_csv(url)
plt.scatter(df['rating'], df['rating_count'])
```

## Expected behavior

Jupyter and JupyterLite should behave the same.

martinRenou:
## References

Fix #906

### DIFF
--- a/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
+++ b/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
@@ -21,9 +21,6 @@ del patches
 from .display import LiteStream
 from .interpreter import LitePythonShellApp
 
-if not os.environ.get("MPLBACKEND"):
-    os.environ["MPLBACKEND"] = "module://matplotlib_inline.backend_inline"
-
 stdout_stream = LiteStream("stdout")
 stderr_stream = LiteStream("stderr")
 
--- a/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
+++ b/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
@@ -3,7 +3,6 @@
 __version__ = "0.1.0b15"
 
 import sys
-import os
 
 # 0. do early mocks that change `sys.modules`
 from . import mocks
--- a/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
+++ b/packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
@@ -3,6 +3,7 @@
 __version__ = "0.1.0b15"
 
 import sys
+import os
 
 # 0. do early mocks that change `sys.modules`
 from . import mocks
@@ -19,6 +20,9 @@ del patches
 # 2. set up the rest of the IPython-like environment
 from .display import LiteStream
 from .interpreter import LitePythonShellApp
+
+if not os.environ.get("MPLBACKEND"):
+    os.environ["MPLBACKEND"] = "module://matplotlib_inline.backend_inline"
 
 stdout_stream = LiteStream("stdout")
 stderr_stream = LiteStream("stderr")
--- a/packages/pyolite-kernel/src/worker.ts
+++ b/packages/pyolite-kernel/src/worker.ts
@@ -78,7 +78,7 @@ export class PyoliteRemoteKernel {
   protected async initKernel(options: IPyoliteWorkerKernel.IOptions): Promise<void> {
     // from this point forward, only use piplite (but not %pip)
     await this._pyodide.runPythonAsync(`
-      await piplite.install(['ipykernel'], keep_going=True);
+      await piplite.install(['pillow', 'ipykernel'], keep_going=True);
       await piplite.install(['pyolite'], keep_going=True);
       await piplite.install(['ipython'], keep_going=True);
       import pyolite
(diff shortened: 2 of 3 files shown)


bollwyvl:
Maybe this could go in patches.py?

martinRenou:
I deliberately did not put it in `patches.py` because it's not a monkey patch. ipykernel does the same here https://github.com/ipython/ipykernel/blob/4dc303364ca73b5185685807de41c7fe4b4a4dc7/ipykernel/kernelapp.py#L559

jtpio:
Kicking CI.

packages/pyolite-kernel/py/pyolite/pyolite/__init__.py
```diff
@@ -20,6 +21,9 @@
 from .display import LiteStream
 from .interpreter import LitePythonShellApp
 
+if not os.environ.get("MPLBACKEND"):
+    os.environ["MPLBACKEND"] = "module://matplotlib_inline.backend_inline"
```
jtpio, line 25: Actually this may be conflicting with the following patch:

https://github.com/jupyterlite/jupyterlite/blob/f11fb9a05d56d69fa5ec6e9b04ab96a097f2a83a/packages/pyolite-kernel/py/pyolite/pyolite/patches.py#L5-L7

bollwyvl:
> pykernel does the same

I'd maintain that, too, is also a patch, as its overriding whatever might be in the user's environment to start with, since Kernel Knows Best. But as it is conflicting with stuff that _is_ in, as `pyodide` does its _own_ stuff to `matplotlib`, seems appropriate in any case.

>  try to remove all those lines from patches.py

This would be _wonderful_ as `matplotlib` + kitchen sink is a _brutal_ up-front cost to pay for every kernel start.

jtpio:
Wondering if we would then still need to import `matplotlib` on startup here?

https://github.com/jupyterlite/jupyterlite/blob/f11fb9a05d56d69fa5ec6e9b04ab96a097f2a83a/packages/pyolite-kernel/src/worker.ts#L81

martinRenou:
> Wondering if we would then still need to import matplotlib on startup here?

I'd personally be fine removing matplotlib from the default installation 👍🏽

jtpio:
iirc this was needed because the current patch. So if the patch is removed it might be possible to remove it there.

martinRenou:
This means that people wanting to use Matplotlib will need to install it, so the old assumption that it is installed is not valid anymore if we remove it.

jtpio:
Doesn't Pyodide include `matplotlib` by default?

image

I'm not exactly sure why it had to be installed with `piplite.install()` in lite.

packages/pyolite-kernel/src/worker.ts
```diff
@@ -78,7 +78,7 @@ export class PyoliteRemoteKernel {
   protected async initKernel(options: IPyoliteWorkerKernel.IOptions): Promise<void> {
     // from this point forward, only use piplite (but not %pip)
     await this._pyodide.runPythonAsync(`
-      await piplite.install(['matplotlib', 'ipykernel'], keep_going=True);
+      await piplite.install(['pillow', 'ipykernel'], keep_going=True);
```
bollwyvl, line 81: one patch down, a couple to go: now we do need `piplite.install(['pillow',)` for the _other_ patch to work, unless we can magic that into existence somehow. but this should get kernels starting.

vasiljevic:
We should also take care of backsword compatibility issues when removing piplite.install for matplotlib from initKernel(). Many, if not most, JupyterLite websites contains notebooks that use matplotlib. After removing matplotlib install from initKernel(), all those websites will be broken when rebuilded with updated JupyterLite. 

We may consider to introduce an option to enabe/disabe piplite.install for matplotlib in initKernel(). This lead us to a solution similar to the PR #811.

jtpio:
@vasiljevic it looks fine when trying the latest state of this PR on ReadTheDocs: https://jupyterlite--911.org.readthedocs.build/en/911/_static/lab/

image

`matplotlib` is included by default in Pyodide so there should be any extra `piplite.install()` required.

bollwyvl:
Some numbers for starting a kernel, ignoring things downloaded before the launcher draws (both about 14.5.mb):

|  ref   | resources (mb) | compressed (mb) | `len(sys.modules)` |
| :----: | -------------: | --------------: | -----------------: |
| `main` |           66.7 |            24.3 |                946 |
|  #911  |           44.1 |            13.0 |                730 |

And, even though `matplotlib` is in `sys.modules` on `main`, it _isn't_ in `user_ns`, so no breaking change! Great success!

image

**SO**: can we figure out a way to get `pillow` out, too!? That would be a magic festivus present :barber: :muscle: :crying_cat_face:

bollwyvl:
oooh... it's just `_repr_png_`... because we control how that works with access to the display system, we can check if `PIL` is in `sys.modules`, and then see if the thing is (a subclass of) `PIL.Image` and hack it more lazily. I'll take a look at it on a sub-pr...

jtpio (approved):
Thanks!!

jtpio:
Right, maybe something to keep track of in https://github.com/jupyterlite/jupyterlite/issues/235.

PR #913, opened 2022-12-13
Remove `Image` hacks, vendor and use upstream `ipykernel.jsonutil`

### LINKED ISSUE #235: Lazy Patching

### LINKED ISSUE #254: Add back `importhook` for patches

### LINKED ISSUE #912: Remove PIL patch
### References

- continues #911

### Problem

Our last "heavyweight" patch is for the `PIL.Image._repr_png_`, but as `pyolite` controls the interpreter and the kernel, we can overload it elsewhere.

### Proposed Solution

In the correct layer of `IPython`, when evaluating `_repr_png`, check if `PIL` is in `sys.modules`, then see if the current thing is a `PIL.Image`, and base64 encode it there.

### Additional context

...

bollwyvl:
## References

- fixes #912
- fixes #235
- fixes #254
- continues #911

## Code changes

- [x] vendor `ipykernel.jsonutil`
  - lightly patched, as we don't have `jupyter_client` because zmq 
- [x] use `json_clean` and `encode_images` in `DisplayHook`
- [x] provide upstream `Image` class in place of shim 
- [x] remove up-front `piplite.install` of `pillow`
- [x] remove `patch_pillow`

## User-facing changes

- _all_ `_repr_{img}_` should be properly base64-encoded (previously, we only patch `Image` and `PIL.Image`)
- pyodide kernels should start slightly faster :laughing: 

|     ref     | resources (mb) | compressed (mb) | `len(sys.modules)` |
| :---------: | -------------: | --------------: | -----------------: |
| before #911 |           66.7 |            24.3 |                946 |
|    #911     |           44.1 |            13.0 |                730 |
|   #913   |           43.1 |            12.6 |                724 |

- n/a

### DIFF
--- a/packages/pyolite-kernel/py/ipykernel/ipykernel/jsonutil.py
+++ b/packages/pyolite-kernel/py/ipykernel/ipykernel/jsonutil.py
@@ -1,0 +1,165 @@
+"""Utilities to manipulate JSON objects."""
+
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import math
+import numbers
+import re
+import types
+from binascii import b2a_base64
+from datetime import datetime
+
+# lite: cannot import this due to zmq
+# from jupyter_client._version import version_info as jupyter_client_version
+
+next_attr_name = "__next__"
+
+# -----------------------------------------------------------------------------
+# Globals and constants
+# -----------------------------------------------------------------------------
+
+# timestamp formats
+ISO8601 = "%Y-%m-%dT%H:%M:%S.%f"
+ISO8601_PAT = re.compile(
+    r"^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})(\.\d{1,6})?Z?([\+\-]\d{2}:?\d{2})?$"
+)
+
+# holy crap, strptime is not threadsafe.
+# Calling it once at import seems to help.
+datetime.strptime("1", "%d")
+
+# -----------------------------------------------------------------------------
+# Classes and functions
+# -----------------------------------------------------------------------------
+
+
+# constants for identifying png/jpeg data
+PNG = b"\x89PNG\r\n\x1a\n"
+# front of PNG base64-encoded
+PNG64 = b"iVBORw0KG"
+JPEG = b"\xff\xd8"
+# front of JPEG base64-encoded
+JPEG64 = b"/9"
+# constants for identifying gif data
+GIF_64 = b"R0lGODdh"
+GIF89_64 = b"R0lGODlh"
+# front of PDF base64-encoded
+PDF64 = b"JVBER"
+
+# lite: we do not know this
+# JUPYTER_CLIENT_MAJOR_VERSION = jupyter_client_version[0]
+
+
+def encode_images(format_dict):
+    """b64-encodes images in a displaypub format dict
+
+    Perhaps this should be handled in json_clean itself?
+
+    Parameters
+    ----------
+    format_dict : dict
+        A dictionary of display data keyed by mime-type
+
+    Returns
+    -------
+    format_dict : dict
+        A copy of the same dictionary,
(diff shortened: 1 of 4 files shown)


bollwyvl:
Buh, github is having _a day_, so RTD isn't building :crying_cat_face:

bollwyvl:
A quick look at our remaining hundreds of `sys.modules`:
- 116 `prompt_toolkit` (presumably nothing to be done about this!)
- 76 `jedi` and 21 `parso`
  - given we already turn off jedi (presumably because it doesn't work for some reason, or is horribly slow), perhaps we can provide another heavyweight shim package?

jtpio:
> Buh, github is having _a day_, so RTD isn't building crying_cat_face

Just kicked it.

> presumably because it doesn't work for some reason

Yes this was disabled in https://github.com/jupyterlite/jupyterlite/pull/448.

packages/pyolite-kernel/py/ipykernel/ipykernel/jsonutil.py
```diff
@@ -0,0 +1,165 @@
...
+GIF_64 = b"R0lGODdh"
+GIF89_64 = b"R0lGODlh"
+# front of PDF base64-encoded
+PDF64 = b"JVBER"
+
+# lite: we do not know this
+# JUPYTER_CLIENT_MAJOR_VERSION = jupyter_client_version[0]
+
+
+def encode_images(format_dict):
```
jtpio, line 54: Is this actually needed if `format_dict` is returned as is?

jtpio: Maybe it can be removed to reduce the amount of vendored code.

bollwyvl: right: if it didn't need the patch, i was going to copy it from `sys.prefix` :stuck_out_tongue_closed_eyes:

jtpio: also in case we need to diff it against upstream at some point it's probably better to not touch it too much :+1:

jtpio (approved):
Thanks!

PR #916, opened 2022-12-13
Resolve example package lists, update matplotlib example

bollwyvl:
## References

- uncommitted things while working on #912
- update docs after #911

## Code changes

- [x] run `doit env repo lint` a couple times in a clean env
- [x] remove `plt.show` from `matplotlib` example 

## User-facing changes

- n/a

- n/a


bollwyvl:
Ran through all the RTD examples, everything is looking good, but we can get rid of `plt.show()` :tada:

PR #917, opened 2022-12-14
add @jupyterlite/contents to list of singleton packages

sglyon:
## References

From discussion here: https://github.com/jupyterlite/jupyterlite/discussions/887#discussioncomment-4394390

## Code changes

- [x] Only change (So far!) is to add @jupyterlite/contents to list of singletonPackages so it can be extended/replaced by external extensions.
- [x] add more explicit dependencies on `@jupyterlite/contents`
- [x] enable more sourcemaps
  - > we _could_ roll this back, as they are _big_ but should help while debugging... and won't  be downloaded by most users
- [x] move import of `DriveFS` to a lazy import in pyodide kernel worker
  - > TODO: probably need an `IDriveFS` so this can be more explicit 
- [ ] ???

## User-facing changes

I am working on a custom content extension that will allow me to hook into the flow of saving a notebook. To do this I am working on creating a new `IContents` provider in an external extension. To get my custom `IContents` provider to work with the `contentRoutesPlugin` (see here https://github.com/jupyterlite/jupyterlite/blob/52ba29fddd7cbf45451e8235f5a1fac81dc07d29/packages/server-extension/src/index.ts#L93-L94), I need to have the singleton package

None


app/lab/package.json
```diff
@@ -232,6 +232,7 @@
       "@jupyterlab/tooltip",
       "@jupyterlab/translation",
       "@jupyterlab/ui-components",
+      "@jupyterlite/contents",
```
bollwyvl, line 235: :+1: on putting in lexicographical order :mage:

bollwyvl: Ah, looks like it needs to appear in:

```
Checking app `resolutions`...
... app/lab/package.json `resolutions` are out-of-date!
... app/repl/package.json `resolutions` are up-to-date!
... app/retro/package.json `resolutions` are up-to-date!
```

The `resolutions` field helps keep the installed versions working properly... not so relevant for lite, as of yet, but kind of a big deal with jupyterlab, which has multiple active "head" versions (e.g. 2.x, 3.5.x, 3.6.x, 4.x).

sglyon: ahh ok thanks for catching that. I'm pretty new to the "under the hood" part of how the jlab system works. Thanks for the patience

jtpio:
Thanks @sglyon and @bollwyvl.

Looks like the UI tests failure is relevant:

image

app/lab/package.json
bollwyvl: yeah: this is kinda like working at the engine fab rather than the auto mechanic's shop, trying to figure out how to make engines that work in more kinds of cars.

Of note, there are a few other custom jupyterlab apps out there that can make use of essentially "vanilla" labextensions (jupyterlite adds the "serverextension" flavor, so is decidedly non-"standard"):
- https://github.com/voila-dashboards/voila/tree/main/packages/voila
- https://github.com/jupyter/notebook/tree/main/packages/application

bollwyvl:
> difference here might be the worker

Yeah, i think it's the worker being built in the _same_ webpack build. I think once properly hoisted, it will work fine. Been meaning to dust off the robot kernel work... of course, all of this will be clearer after #386, sorry we keep piling stuff on!

bollwyvl:
I've recently been seeing some timeouts on RTD on other projects, not sure what's up...

bollwyvl:
I wonder if it's related to `packaging 22`, which seems to be breaking things all over.

bollwyvl:
RTD cleared, firefox still unhappy...

bollwyvl:
@jtpio can we make it gather browser logs on fail?

jtpio:
Probably we could log the console messages with `page.on('console')` by default.

I'll have a look at the UI tests locally, as this seems to be working fine now in a "normal" Firefox.

jtpio:
> Probably we could log the console messages with `page.on('console')` by default.

We could indeed start logging the dev tools console messages and pipe them to a file so they can be downloaded as artifact.

> I'll have a look at the UI tests locally,

Getting the following locally:

```
Uncaught (in promise) ReferenceError: setImmediate is not defined
```

image

Oddly the Firefox version indicates 93 (very old) even after a clean `yarn playwright install firefox`. Need to investigate more.

bollwyvl:
If we can find where, we should just polyfill setimmediate. I fought with
this for a couple hours on jupyterlite-lsp

jtpio:
> If we can find where, we should just polyfill setimmediate. I fought with this for a couple hours on jupyterlite-lsp

We could indeed polyfill this, might might be possible with a simple `setTimeout`.

However bringing the update to the latest Playwright (from https://github.com/jupyterlite/jupyterlite/pull/923), which comes with recent browsers, seems to be fixing the failing test.

I think we can move forward to get this change and the other recent improvements out in a release.

jtpio (approved):
Thanks!

PR #923, opened 2022-12-20
Update to Playwright 1.29

jtpio:
## References

Update to the latest Playwright to avoid lagging behind for too long.

## Code changes

Dependency update.

Playwright 1.18.0 introduced breaking changes to the configuration options: https://github.com/microsoft/playwright/releases/tag/v1.18.0. 

This is fixed upstream in JupyterLab https://github.com/jupyterlab/jupyterlab/pull/13140 but it's only available in the 4.0 packages. We can remove the `use` override when updating to the latest packages in https://github.com/jupyterlite/jupyterlite/issues/826.

## User-facing changes

None

None

### DIFF
--- a/ui-tests/package.json
+++ b/ui-tests/package.json
@@ -16,10 +16,10 @@
     "test:update": "playwright test --update-snapshots"
   },
   "dependencies": {
-    "@jupyterlab/galata": "~4.4.0",
+    "@jupyterlab/galata": "~4.4.1",
     "yarn-deduplicate": "^3.1.0"
   },
   "resolutions": {
-    "@playwright/test": "~1.16.2"
+    "@playwright/test": "~1.26.0"
   }
 }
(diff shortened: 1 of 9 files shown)


jtpio:
> This is fixed upstream in JupyterLab [jupyterlab/jupyterlab#13140](https://github.com/jupyterlab/jupyterlab/pull/13140) but it's only available in the 4.0 packages. We can remove the `use` override when updating to the latest packages in #826.

This is now tracked as part of https://github.com/jupyterlite/jupyterlite/issues/826.

jtpio:
Getting this one in. I think we can live with the `test.use()` workaround for now until we update to lab 4 and nb 7 (https://github.com/jupyterlite/jupyterlite/issues/826).

Curious to see how this will affect https://github.com/jupyterlite/jupyterlite/pull/917/

PR #933, opened 2022-12-23
Add docs for using xeus-python

martinRenou:
## References

As discussed in https://github.com/jupyterlite/xeus-python-demo/issues/1

## Code changes

Only docs

## User-facing changes

None

### DIFF
--- a/docs/howto/python/preinstalled_packages.md
+++ b/docs/howto/python/preinstalled_packages.md
@@ -1,0 +1,9 @@
+# Pre-install additional packages with emscripten-forge
+
+Using [xeus-python](https://github.com/jupyterlite/xeus-python-kernel), you can pre-install packages available both on ``conda-forge`` and ``emscripten-forge`` by specifying them to the ``environment.yml`` file in the JupyterLite build directory.
+
+By pre-installing packages, they are readily usable in the kernel and can be imported without the need for ``piplite``.
+
+When pre-installing packages that provides JupyterLab extensions (*e.g.* ipywidgets or ipyleaflet), those JupyterLab extensions are automatically included in the JupyterLite build output without the need for extra configuration.
+
+As an example, a deployment can easily be made using the [JupyterLite deployment demo with xeus-python](https://github.com/jupyterlite/xeus-python-demo)].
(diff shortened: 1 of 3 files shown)


docs/howto/python/preinstalled_packages.md
```diff
@@ -0,0 +1,9 @@
+# Pre-install additional packages with emscripten-forge
```
jtpio, line 1: Since most of the other guides in this category are about the Pyolite kernel, maybe we should make it more explicit that this will only work with xeus python:

image

jtpio: Also maybe a small `warning` at the top of this file to make it more visible this is for a different kernel.

jtpio:
Thanks!

I left a couple of comments. Probably we could rework the docs later on after the Pyodide kernel is moved to a separate repo (https://github.com/jupyterlite/jupyterlite/pull/854).

jtpio:
> Probably we could rework the docs later on after the Pyodide kernel is moved to a separate repo (#854).

I meant so it will be clearer what can be done with each of the two Python kernels.

But for now we can probably keep the guides in the same category.

martinRenou:
> what can be done with each of the two Python kernels

Yep totally, we should make a clear separation between the two. I will update my PR.

docs/howto/python/preinstalled_packages.md
```diff
@@ -0,0 +1,9 @@
...
+
+Using [xeus-python](https://github.com/jupyterlite/xeus-python-kernel), you can pre-install packages available both on ``conda-forge`` and ``emscripten-forge`` by specifying them to the ``environment.yml`` file in the JupyterLite build directory.
+
+By pre-installing packages, they are readily usable in the kernel and can be imported without the need for ``piplite``.
+
+When pre-installing packages that provides JupyterLab extensions (*e.g.* ipywidgets or ipyleaflet), those JupyterLab extensions are automatically included in the JupyterLite build output without the need for extra configuration.
+
+As an example, a deployment can easily be made using the [JupyterLite deployment demo with xeus-python](https://github.com/jupyterlite/xeus-python-demo)].
```
tfmark, line 9: trailing `]` ?

jtpio: Good catch thanks!

image

jtpio, line 9:
```suggestion
As an example, a deployment can easily be made using the [JupyterLite deployment demo with xeus-python](https://github.com/jupyterlite/xeus-python-demo).
```

jtpio:
FYI @martinRenou I'll likely do a quick pass on this PR soon.

Now that we are decoupling the Pyodide and JS kernels out of the main bundle the documentation could also highlight how deployers can choose and configure the kernels they want.

Will be pushing either here or in a new PR that includes your changes.

jtpio:
Continuing this in https://github.com/jupyterlite/jupyterlite/pull/1018. Thanks!

PR #934, opened 2022-12-23
Allow Addons to provide CLI aliases and flags

### LINKED ISSUE #929: Allow addons to register short CLI args
### References
- from https://github.com/jupyterlite/jupyterlite/pull/854/files#r1055746387

### Problem

Any third-party addons can only specify "extra long" traitlets args.

```
jupyter lite build --XeusPythonEnv.environment_file=custom.yml
```

#854 brings this to a head, where we already have documentation and use in the wild that references the few CLI args for the pyodide-based functionality. It doesn't seem we need to _needlessly_ break backward CLI compatibility, aside from a new line in a `requirements.txt` for `pyodide-kernel`.

### Proposed Solution
- _Core_ `Addons` (e.g. in this repo) would continue to have config set in the `LiteBuildConfig`.
  - #854 already removes the `pyodide` and `piplite` ones
- allow `Addon`s _may_ specify a static mapping of CLI arguments to traits:
  - `pyodide-kernel`
    - `extra_args = {"piplite-wheels": "piplite_urls", "pyodide": "pyodide_url"}`
  - `xeus-python`
    - `extra_args = {"environment-file": "environment_file"}`
- in the main app, capture [`extra_args`](https://github.com/ipython/traitlets/blob/main/traitlets/config/application.py#L408) 
  - hand them down as appropriate to all addons that want them
- include the all the traits' `help` in the `jupyter lite --help`
  - could show potential collisions
    - but, in the future, if `xeus-python` can also consume `pyodide` wheels, _a la_ `conda-pack`, all the better!

### Additional context

- I don't know of a _good_ example of this pattern on `traitlets`
  - eventually might be worth hoisting

bollwyvl:
## References

- fixes #929

## Code changes

- [x] extend instance of `LiteApp` with `.aliases` and `.flags` found in addons
  - [x] load all `entry_points` up-front (but not on _import_)
- [x] ensure they are hoisted to `--help`
- [x] move `piplite_urls` and `pyodide` to `PipliteAddon` and `PyodideAddon`  
  - [x] update docs for these
- [x] add developer docs for adding aliases and flags to Addons
- [x] resolve examples, update with `PipliteAddon` config example

## User-facing changes

- downstream Addons such as the xeus one explored in #929 could now add short CLI params
- the CLI does not change :tada:
- the config file format has now changed :crying_cat_face: 

- This (perhaps prematurely) moves the pyodide-related CLI changes, which will break with e.g. `LiteBuildConfig.piplite_urls`, and will create conflicts with #854
  - we _could_ add a shim for the time being, but will complicate things
    - basically leave the top-level trait, and consult that (with a warning) if not configured on the right place
- previously, addon implementations which were disabled by config would still be discovered, but then would not be loaded
  - addons with side-effects of being imported (other than being slow) might now take effect


py/jupyterlite/src/jupyterlite/addons/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    """Load (and cache) addon implementations.
+
+    Pass some noise (like `date.date`) to the ``force`` argument to reload.
+    """
+    addon_implementations = {}
+    for name, entry_point in get_addon_entry_points(force).items():
+        try:
+            addon_implementations[name] = entry_point.load()
+        except Exception as err:  # pragma: no cover
+            warnings.warn(f"[lite] [{name}] failed to load: {err}")
```
martinRenou, line 69: It could be convenient if this was an exception instead, I've seen cases where the xeus-python addon would fail to load and it would have been convenient for `jupyter lite build` to not continue building.

What do you think?

martinRenou (commented):
Super neat! Thanks for doing this!

I just have a minor comment/question, but it can be discussed in a separate issue later as it's not concerning a change in the curent jupyterlite behavior.

martinRenou (approved):

jtpio (approved):
Thanks!

PR #939, opened 2023-01-03
normalize all indexed piplite wheel names

bollwyvl:
## References

- n/a

## Code changes

- [x] apply PEP 503 name normalization to the top-level of the Warehouse API index

## User-facing changes

- Packages with non-503 names (e.g. including `_`, uppercase letters) should be indexed and found more easily

- probably none, as `micropip` has been doing this for a while


jtpio (approved):
Thanks!

PR #943, opened 2023-01-09
Update to lerna 6.4

jtpio:
## References

This might help fix a couple of dependabot alerts.

## Code changes

Update the `lerna` dependency.

## User-facing changes

None

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "jest-junit": "^11.1.0",
     "jest-raw-loader": "^1.0.1",
     "jest-summary-reporter": "^0.0.2",
-    "lerna": "^6.1.0",
+    "lerna": "^6.4.0",
     "lint-staged": "^10.4.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.8.1",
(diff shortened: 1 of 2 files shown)


bollwyvl:
ideally, the `@jupyterlab/builder` most extension authors _must_ add to their package would _not_ rely on `verdaccio` at all.

if we actually use `verdaccio` in the release process for this, we can `resulitions: {"verdaccio": "file:packages/not-a-package"}` (with a dummy, but well-formed) `packages/not-a-package/package.json`, which will clobber a whole bunch of the alerts. `got` is still problematic.

jtpio:
> ideally, the `@jupyterlab/builder` most extension authors _must_ add to their package would _not_ rely on `verdaccio` at all.

Right, we should probably open an issue upstream in JupyterLab and try to fix this.

jtpio:
`verdaccio` is not used for the JupyterLite release process. But it is for JupyterLab.

jtpio:
> Right, we should probably open an issue upstream in JupyterLab and try to fix this.

This is now tracked in https://github.com/jupyterlab/jupyterlab/issues/13740.

Issue #320, opened 2021-09-09
Plotly not work


qqdaiyu55:
## Description

Paste sample plotly code
```
import plotly.graph_objects as go
fig = go.Figure()
fig.add_trace(go.Scatter(y=[2, 1, 4, 3]))
fig.add_trace(go.Bar(y=[1, 4, 3, 2]))
fig.update_layout(title = 'Hello Figure')
fig.show()
```

Error logged in console
![image](https://user-images.githubusercontent.com/9846706/132708022-9e0fc2c0-2847-4ff4-bb16-a81099b26c52.png)

I see it works in [demo site](https://jupyterlite.readthedocs.io/en/latest/_static/lab/index.html), maybe a regression for new jupyterlite version or I need to install some extensions?

---
FYI my setting
- Jupyter lite version: v0.1.0a6
- System: Windows 21H1
- Python: 3.9.1
- pip version: 20.2.3
- Browser and version: Version 92.0.4515.159 (Official Build) (x86_64)

qqdaiyu55:
More detail (in `VM773:1` showed in pic)
![image](https://user-images.githubusercontent.com/9846706/132708778-6af7e59f-34c5-4639-9d9b-0b1f8140bb42.png)

```
[...]

PR #950, opened 2023-01-14
Add back `plotly` to the examples

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/320

## Code changes

According to the `5.12.0` release notes `plotly` should now work with ipywidgets 8: https://github.com/plotly/plotly.py/releases/tag/v5.12.0

## User-facing changes

End users should be able to use `plotly` with the latest JupyterLite and Pyodide kernel.

None


jtpio:
Ah looks like there might be some issues probably on the plotly side: https://jupyterlite--950.org.readthedocs.build/en/950/_static/lab/index.html

```
Uncaught ReferenceError: require is not defined
```

image

.binder/environment.yml
```diff
@@ -46,6 +46,7 @@ dependencies:
   - ipympl
   - ipyvue
   - ipyvuetify
+  - plotly
```
bollwyvl, line 49: probably add a bottom pin, e.g.
```
- plotly >=5.12.0
```
... but this will force a conflict, because 

https://github.com/conda-forge/plotly-feedstock/pull/118

jtpio:
Adding the `5.12.0` lower bound yields the same require error:

image

Might be related to the plotly renderer issue noticed a while ago in https://github.com/jupyterlite/jupyterlite/issues/110#issuecomment-866077035.

MRigal:
Almost at the same time, on the plotly side: https://github.com/plotly/plotly.py/issues/4020

jtpio:
Not sure why it's an issue here on the ReadTheDocs demo site. It appears to be working fine on the GitHub Pages demo:

image

jtpio:
ok so exactly sure why but it looks fine now from this PR:

image

jtpio:
cc @qqdaiyu55 who originally opened the issue: https://github.com/jupyterlite/jupyterlite/issues/320

I also left some notes in an issue on the plotly repo: https://github.com/plotly/plotly.py/issues/4143

Issue #910, opened 2022-12-12
Installing JupyterLite as a pwa fails on iPadOS


dhombios:
## Description

Installing jupyterlite on android results in an application indistinguishable from the ones provided in the play store (from the point of view of the user) as it normally happens with any pwa. 

However, doing the same on iPadOS just generates a bookmark that opens safari and tries to connect to jupyterlite website. If no internet connection is available, it won't be loaded (doing the same with Vs code for the web allows using it offline as any other app)

## Reproduce

1. Open safari 
2. Go to https://jupyterlite.github.io/demo/lab/index.html
3. Add to Home Screen
4. Go to Home Screen and press jupyterlite icon
5. https://jupyterlite.github.io/demo/lab/index.html will open in safari 

It seems that iPadOS is not recognizing jupyterlite as a pwa

## Expected behavior

Pressing jupyterlite icon should open jupyterlite as a separate app instead of opening its website in Safari

## Context

- JupyterLite version: Version 0.1.0-beta.15
- Operating System and version: iPadOS 16.1.1
- Browser and version: Safari

<details><summary>Browser Output</summary>
<pre>
Browser output is not available in Safari in iPadOS 
</pre>
</details>

bollwyvl:
PRs welcome! I don't have a device to test this on.

jtpio:
Yes any help would be really appreciated!

> Installing jupyterlite on android results in an application indistinguishable from the ones provided in the play store (from the point of view of the user) as it normally happens with any pwa.

This should probably be mentioned in the documentation at some point since it's quite of a cool feature :+1: 

<img src="https://user-images.githubusercontent.com/591645/207124737-aa7fa22b-ce48-4812-afc0-75b351f62278.jpg" height="512">

dhombios:
Taking into account that all its features seem to work properly, I think it might be a problem with the manifest. Maybe safari requires an additional field

Regarding the documentation, it is a really nice feature, as it allows bringing jupyter to tablets

dhombios:
Looking at the manifest, it may be a problem with the display option. It seems that iPadOS does not support minimal-ui and therefore it is using browser. It seems that there is an option for changing that behavior according to this website: https://web.dev/add-manifest/

dhombios:
As it's been a while since the last message and it doesn't seem that any contributor has the tools required to test it on an iPad, I could try the suggested modifications. However, I don't have a Mac so I won't be able to provide debug information

[...]

PR #951, opened 2023-01-14
Display field in webmanifest changed to standalone to support iPadOS

dhombios:
Minimal-ui is unsupported on iPadOS, so in order to allow installing jupyter lite and running it offline as a native app it needs to be changed to standalone.

See https://firt.dev/notes/pwa-ios/ for more information about supported PWA features on iOS and iPadOS 

## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/910

## Code changes

Display field in manifest.webmanifest has been changed to “standalone”, as minimal-ui is not supported on iPadOS. This avoids falling back to browser, which caused the app to be opened on the browser instead of showing it as an installed app

## User-facing changes

On iPad, the app now is shown as if it was a native app, instead of being opened on the browser (the appearance of the window on other systems will also change, however the appearance of the app will remain unchanged)

Before:
image

After:
image

This pull request does not cointain backwards-incompatible changes

### DIFF
--- a/app/manifest.webmanifest
+++ b/app/manifest.webmanifest
@@ -15,7 +15,7 @@
   ],
   "start_url": "./",
   "background_color": "#fff",
-  "display": "minimal-ui",
+  "display": "standalone",
   "scope": "./",
   "shortcuts" : [
     {


jtpio:
Nice thanks @dhombios for looking into this :+1:

dhombios:
It's been a pleasure. iPad needs Jupyter support

jtpio:
I just tried on an Android phone and using `standalone` removes the bar at the top of screen, which helps save some space. 

https://user-images.githubusercontent.com/591645/213018779-c9f164a4-3aa3-41e5-a568-281c181b0cef.mp4

*Left is the latest 0.1.0b17 release, right is this PR using the ReadTheDocs preview available at https://jupyterlite--951.org.readthedocs.build/en/951/_static/lab/index.html*

jtpio:
@dhombios do you think we could undraft this PR so it can be merged and released?

dhombios:
I have undrafted it. I was just waiting until we knew why it was changed to minimal-ui, but I suppose that if an issue is found this change can be reverted quite easily as its just a line

jtpio:
>  I was just waiting until we knew why it was changed to minimal-ui

Right, not sure why but as you say it should be easy to revert in case this causes major issues. It was also `standalone` in the original PR that added the manifest: https://github.com/jupyterlite/jupyterlite/pull/137

jtpio (approved):
Thanks!

PR #959, opened 2023-01-25
Fix mamba environment creation command

davidbrochart:
`mamba env update` doesn't seem to be a valid command.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ Various package managers on different operating systems provide these.
 > Binder environment description checked into the repo.
 >
 > ```bash
-> mamba env update --file .binder/environment.yml
+> mamba env create --file .binder/environment.yml
 > mamba activate jupyterlite-dev
 > ```
 >


bollwyvl:
What kind of errors were you seeing? I pretty much exclusively use `env update`, as otherwise you have to blow away the env, or use `-f`.

davidbrochart:
```bash
mamba --version
# 1.2.0

mamba env update --file .binder/environment.yml
# The following arguments were not expected: .binder/environment.yml --file update
# Run with --help for more information
```

bollwyvl:
sounds like a regression in mamba

jtpio:
Is there an issue tracking this in mamba?

davidbrochart:
Yes: https://github.com/mamba-org/mamba/issues/2369

davidbrochart:
It seems to be caused by a conflict when both `mamba` (e.g. through `mambaforge`) and `micromamba` are installed.

jtpio:
Thanks @davidbrochart for opening the issue upstream and the extra context :+1:

Issue #963, opened 2023-01-26
Docs fail to build


jtpio:
As noticed in https://github.com/jupyterlite/jupyterlite/pull/960#issuecomment-1403889354, the documentation has started to fail to build:

```
$ /home/docs/checkouts/readthedocs.org/user_builds/jupyterlite/checkouts/959/node_modules/.bin/svgo --multipass --pretty --indent=2 --final-newline -o
error: option '-o, --output <OUTPUT...>' argument missing
error Command failed with exit code 1.
```

It looks related to `svgo`, although the dependency hasn't changed for a long time in `yarn.lock`.

So maybe something changed on the ReadTheDocs CI.

Example failure: https://readthedocs.org/projects/jupyterlite/builds/19271915/

jtpio:
Also the docs seem to be working fine when not on ReadTheDocs, for example on Gitpod.

Maybe we should skip `svgo` on RTD for now.

jtpio:
This is the generated command being executed:

```
subprocess.CalledProcessError: Command '['yarn', 'svgo', '--multipass', '--pretty', '--indent=2', '--final-newline', '-o']' returned non-zero exit status 1.
```

Maybe the glob is not able to find the files for some reasons:

https://github.com/jupyterlite/jupyterlite/blob/06d8a555844323112cb0e8e862b9c133e574cbb0/dodo.py#L643

jtpio:
Closing as fixed by https://github.com/jupyterlite/jupyterlite/pull/964

PR #964, opened 2023-01-26
Troubleshoot ReadTheDocs

jtpio:
## References

Just trying to troubleshoot https://github.com/jupyterlite/jupyterlite/issues/963.

## User-facing changes

None

None

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -145,6 +145,9 @@ def before_rtd_build(app: Sphinx, error):
 
 def after_build(app: Sphinx, error):
     """sphinx-jsonschema makes duplicate ids. clean them"""
+    os.environ.update(
+        JLITE_DOCS_OUT=app.builder.outdir
+    )  # <--- dodo.py already looking for this
     do_tasks("post", RTD_POST_TASKS)
 
 


dodo.py
```diff
@@ -647,12 +647,13 @@ def _optimize_images():
                 cwd=str(P.ROOT),
             )
 
-        yield dict(
-            name="post:images",
-            doc="clean sphinx-jsonschema duplicate ids",
-            actions=[_optimize_images],
-            file_dep=[B.YARN_INTEGRITY],
-        )
+        if not C.RTD:
```
jtpio, line 650: Probably doesn't make sense to skip on RTD since the task is about optimizing svgs in the docs folder.

jtpio:
So when trying in a fresh env there should be some svg files around:

image

Not sure why they are not found on RTD.

bollwyvl:
Maybe rtd changed their build location, and it needs to be passed in or set via env var?

jobovy:
> Maybe rtd changed their build location, and it needs to be passed in or set via env var?

Yes, they changed the build location to `_readthedocs`: readthedocs/readthedocs.org/pull/9888.

bollwyvl:
This would probably work until there's a documented env var:

```py
# conf.py
def after_build(app: Sphinx, error):
    """sphinx-jsonschema makes duplicate ids. clean them"""
    os.environ.update("JLITE_DOCS_OUT", app.builder.outdir)  # <--- dodo.py already looking for this
    do_tasks("post", RTD_POST_TASKS)
```

bollwyvl:
Ah, maybe one is coming:

https://github.com/readthedocs/readthedocs.org/pull/9913

But the above is probably still the right play for now.

bollwyvl:
sry, of course my syntax bad:

```py
os.environ.update(JLITE_DOCS_OUT=app.builder.outdir) 
```

bollwyvl:
yay docs

Issue #914, opened 2022-12-13
Investigate further pyolite startup time improvements


bollwyvl:
### Problem

`pyolite` requires _many_ packages to be fully downloaded (if not already cached) and installed before giving the user the ability to do interactive computing. 

### Proposed Solution

Investigate additional packages which could be avoided, either by shim or patch.

### Additional context

With #913, we'd be down to 724 modules.

<details>
<pre>
>>> import sys
>>> sorted(sys.modules)
['IPython',
 'IPython.core',
 'IPython.core.alias',
 'IPython.core.application',
 'IPython.core.async_helpers',
 'IPython.core.autocall',
 'IPython.core.builtin_trap',
 'IPython.core.compilerop',
 'IPython.core.completer',
 'IPython.core.completerlib',
 'IPython.core.crashhandler',
 'IPython.core.debugger',
 'IPython.core.display',
 'IPython.core.display_functions',
 'IPython.core.display_trap',
 'IPython.core.displayhook',
 'IPython.core.displaypub',
 'IPython.core.error',
 'IPython.core.events',
 'IPython.core.excolors',
 'IPython.core.extensions',
 'IPython.core.formatters',
 'IPython.core.getipython',
 'IPython.core.history',
 'IPython.core.hooks',
 'IPython.core.inputtransformer2',
 'IPython.core.interactiveshell',
 'IPython.core.latex_symbols',
 'IPython.core.logger',
 'IPython.core.macro',
 'IPython.core.magic',
 'IPython.core.magic_arguments',
 'IPython.core.magics',
 'IPython.core.magics.auto',
 'IPython.core.magics.basic',
 'IPython.core.magics.code',
 'IPython.core.magics.config',
 'IPython.core.magics.display',
 'IPython.core.magics.execution',
 'IPython.core.magics.extension',
 'IPython.core.magics.history',
 'IPython.core.magics.logging',
 'IPython.core.magics.namespace',
 'IPython.core.magics.osm',
 'IPython.core.magics.packaging',
 'IPython.core.magics.pylab',
 'IPython.core.magics.script',
 'IPython.core.oinspect',
 'IPython.core.page',
 'IPython.core.payload',
 'IPython.core.prefilter',
 'IPython.core.profiledir',
 'IPython.core.pylabtools',
 'IPython.core.release',
 'IPython.core.shellapp',
 'IPython.core.splitinput',
 'IPython.core.ultratb',
 'IPython.core.usage',
 'IPython.display',
[...]

PR #978, opened 2023-02-11
Update to pyodide 0.22.1, use NodeJS 18 LTS

bollwyvl:
## References

- as suggested in https://github.com/jupyterlite/jupyterlite/issues/914#issuecomment-1352232342

## Code changes

- [x] bump to `nodejs 18`
- [x] add top pin for `jupyterlab >=3.5.3,<3.6` to avoid breaking changes from upstream (e.g collaboration, rust dependencies)
- [x] resolve examples
- [x] update to `pyodide 0.22.1`  

## User-facing changes

- users of the `[lab]` extra will get a compatible jupyterlab (not 3.6.x)

- ???


bollwyvl:
It looks like requiring `pyodide-build` would be a [_heavy_ build-time dependency](https://github.com/conda-forge/staged-recipes/pull/22011) for site builders, so we'll likely just want to vendor the code that [generates the `repodata.json`](https://github.com/pyodide/pyodide/blob/0.22.1/pyodide-build/pyodide_build/buildall.py#L548), even if we can have access to it during test (especially when updating pyodide versions).

bollwyvl:
The deeper stuff looks... involved. Going to change the intent of this PR (which is already useful) and do a follow-on.

jtpio (approved):
Thanks!

Issue #965, opened 2023-01-26
Execution of "%matplotlib notebook" before "import matplotlib" fails


boisgera:
## Description

```
%matplotlib notebook
```

generates the error

```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 1
----> 1 get_ipython().run_line_magic('matplotlib', 'notebook')

File /lib/python3.10/site-packages/IPython/core/interactiveshell.py:2369, in InteractiveShell.run_line_magic(self, magic_name, line, _stack_depth)
   2367     kwargs['local_ns'] = self.get_local_scope(stack_depth)
   2368 with self.builtin_trap:
-> 2369     result = fn(*args, **kwargs)
   2370 return result

File /lib/python3.10/site-packages/IPython/core/magics/pylab.py:99, in PylabMagics.matplotlib(self, line)
     97     print("Available matplotlib backends: %s" % backends_list)
     98 else:
---> 99     gui, backend = self.shell.enable_matplotlib(args.gui.lower() if isinstance(args.gui, str) else args.gui)
    100     self._show_matplotlib_backend(args.gui, backend)

File /lib/python3.10/site-packages/IPython/core/interactiveshell.py:3522, in InteractiveShell.enable_matplotlib(self, gui)
   3501 def enable_matplotlib(self, gui=None):
   3502     """Enable interactive matplotlib and inline figure support.
   3503 
   3504     This takes the following steps:
   (...)
   3520         display figures inline.
   3521     """
-> 3522     from matplotlib_inline.backend_inline import configure_inline_support
   3524     from IPython.core import pylabtools as pt
   3525     gui, backend = pt.find_gui_and_backend(gui, self.pylab_gui_select)

File /lib/python3.10/site-packages/matplotlib_inline/__init__.py:1
----> 1 from . import backend_inline, config  # noqa
      2 __version__ = "0.1.6"  # noqa

File /lib/python3.10/site-packages/matplotlib_inline/backend_inline.py:6
      1 """A matplotlib backend for publishing figures via display_data"""
      3 # Copyright (c) IPython Development Team.
      4 # Distributed under the terms of the BSD 3-Clause License.
----> 6 import matplotlib
      7 from matplotlib import colors
      8 from matplotlib.backends import backend_agg

ModuleNotFoundError: The module 'matplotlib' is included in the Pyodide distribution, but it is not installed.
You can install it by calling:
[...]

Issue #904, opened 2022-12-08
Loading a different version of a package included in `pyodide`


jobovy:
## Description

I embed a jupyterlite REPL to allow people to try my ``galpy`` code on its documentation site ([here](https://docs.galpy.org/en/latest/#try-galpy)), but I want to allow users to use the latest development version of the code there. But because ``galpy`` is included in the ``pyodide`` distribution, the version that is loaded whenever I include an `import galpy` statement in the cell where I try to load the latest version is the default ``pyodide`` version. 

I believe this happens because ``loadPackagesFromImports`` is called before any cell is run:
https://github.com/jupyterlite/jupyterlite/blob/31d91cacc86ef109eac558487f2efaf37acc1571/packages/pyolite-kernel/py/pyolite/pyolite/kernel.py#L86-L97

So even if the cell includes a command to install a different version (e.g., ``await piplite.install(...)``), the default version of the package is loaded first.

## Reproduce

Running
```
import pyodide_js
await pyodide_js.loadPackage(['numpy','scipy','matplotlib','astropy','future','setuptools','https://www.galpy.org/wheelhouse/galpy-latest-cp310-cp310-emscripten_3_1_14_wasm32.whl'])
import galpy; print(galpy.__version__)
```
prints `1.8.0`, while running
```
import pyodide_js
await pyodide_js.loadPackage(['numpy','scipy','matplotlib','astropy','future','setuptools','https://www.galpy.org/wheelhouse/galpy-latest-cp310-cp310-emscripten_3_1_14_wasm32.whl'])
```
and in the next cell
```
import galpy; print(galpy.__version__)
```
prints `1.8.2.dev0`.

I know I probably shouldn't be using `pyodide_js.loadPackage` here, but I believe the same will happen with ``micropip.install`` or ``piplite.install``. I use ``pyodide_js.loadPackage`` because my wheel does not have a valid name (I should probably change that!).

## Expected behavior

Load the requested version. Not sure how though! Options might include:
* Have a configuration parameter giving packages _never_ to load from the default distribution and then replace the ``_load_packages_from_imports`` call with something that uses [``pyodide.code.find_imports``](https://pyodide.org/en/stable/usage/api/python-api/code.html#pyodide.code.find_imports), pops packages never to load, and then runs ``loadPackage``.
* Search the code to be run for custom installation commands (like ``micropip.install`` etc.) and then pop any package that will be customly installed. That seems difficult though, because the wheel name doesn't necessarily include the package name.

[...]

Issue #925, opened 2022-12-20
Pipe the UI tests console messages to a file on CI


jtpio:
### Problem

As noticed in https://github.com/jupyterlite/jupyterlite/pull/917, the UI tests can sometimes fail and timeout even though trying with a "real" browser works fine. This might be because the versions of the browsers used in the UI tests can be different and not the latest.

In https://github.com/jupyterlite/jupyterlite/pull/917 the failure was because of an error in the dev tools console which caused a timeout.

Ideally there should be a way to get the console output piped to a file when the UI tests are running CI. At least when the tests are failing.

### Proposed Solution

Add logic to pipe the dev tools console logs (at least errors) to a file so they can be inspected more easily.

This might require adding some `page.on('console')` hooks.

### Additional context

Follow-up to https://github.com/jupyterlite/jupyterlite/pull/917

PR #979, opened 2023-02-12
Generate pyodide `repodata.json`

bollwyvl:
## References

- includes jupyterlite/jupyterlite#978
- as suggested in https://github.com/jupyterlite/pyodide-kernel/issues/14
- jupyterlite/jupyterlite#904
- jupyterlite/pyodide-kernel#60
- jupyterlite/jupyterlite#965 (as long as the `import` is in the same cell as the magic)

## Code changes

- [x] add pyodide-compatible `repodata.json` generation
  - [x] extract data from `top_level.txt` (`setuptools` wheels)
  - [x] detect from `RECORD` by
    - [x] `site-packages/<importable>.py`
    - [x] shortest unique `__init__.py` paths (curse you namespace packages)
  - > can be seen [on RTD](https://jupyterlite--979.org.readthedocs.build/en/979/_static/pypi/repodata.json) 
- [x] include repodata in build
- [x] use repodata to patch pyodide auto import functions
- [x] add settings to enable skipping/patching install of packages e.g. `jedi`, `parso` (saves 2mb before gzip)
- [x] mock `jupyterlab_widgets` (saves another 300kb before gzip) 
- [x] mock `prompt_toolkit` (saves another 1.1mb before gzip) 
- [x] remove most `%piplite` calls from examples 
  - [x] except for the one _about_ piplite for arbitrary wheels from the internet
- [x] hide all of the above behind an opt-in flag for now, 
  - `--piplite-install-on-import` 
  - `PyodideAddon.install_on_import`
- [x] ~~add `repodata_patches` to pyolite plugin config~~
  - > i couldn't make this work properly
- [x] tests
  - [x] use `127.0.0.1` everywhere instead of `localhost`, as some CI machines have started doing more ipv6 stuff that some of the tools (`karma`) don't know how to work with by default
- [x] docs

## User-facing changes

- viewers of JupyterLite sites with pre-configured metadata 
  - will not see non-portable magics in `pyolite` examples
  - kernels should start somewhat faster due to
    - making fewer API requests
    - loading shims for `jedi` and `prompt_toolkit` which pretty much don't work in the browser anyway)
- site builders will be able to 
  - replace `pyodide`-shipped modules (the demo currently shows this, shipping a newer `pygments`
    - but _might_ not work automagically for exotic wheels

- I'm still not entirely sure what is different between `micropip.install` and `pyodide_js.loadPackage`
  - hence this behavior is currently opt-in

## Notes

- `repodata.json` can't handle multiple versions of the same package
- finding exported names is trash, and maybe a fool's errand for the general case


bollwyvl:
the folium example is broken, as we're not finding `charset-normalizer` for some reason.

bollwyvl:
Some crude performance on: 
- opening `lab/index.html` on RTD
- wait for the launcher
- clear the network cache
- open a notebook

| task| #978 + #979 | main |
|-|-|-|
| start kernel | image | image
| load `ipywidgets`, draw slider | image | image |

<10mb is :yum: .

bollwyvl:
With `prompt_toolkit` and `jedi` out, the remaining heavy hitters:
image

Some ideas:
- 500kb: somehow patch out the need for `sqlite3`
  - we already have sqlite3-backed history disabled
    - but it's already unvendored from upstream and requires an explicit install
      - removing it would probably breaks other packages, too
- 200kb: convince `ipython` to stop shipping tests in the wheel
  - by naively deleting everything that looks like a test, this would save
  - as it no longer ships the `iptest` command, there's really no user-serviceable reason to keep them
- 20kb: convince `traitlets` to stop shipping tests in the wheel

With those we would get under 9mb, but below that would start to see diminishing marginal returns vs complexity.

bollwyvl:
aside from the UI test fails (which I'll look into, but likely on the weekend) I don't anticipate any additional changes, so the invite for review stands @jtpio.

bollwyvl:
Hm, UI test fails look real, I pine for #925.

bollwyvl:
ah-ha: absolutely can reproduce the issue. I moved the patches to before loading `micropip` and `ipykernel`, and the core `/app/lab/index.html` _doesn't_ ship the dependencies of `ipykernel`, so it fails (initially) on `wcwidth`. 

I had naively assumed this would be able to fall back to fetching with `micropip` if not available, but clearly this is not the case! Will move back to draft and think more.

bollwyvl:
Ok, back to ready for review!

bollwyvl:
@jtpio guess I'd like to show this at the next community call... the PR, as it stands, would be fine to demo, but would be lovely to get this in so we can do some more downstream work with it without building from source.

jtpio:
Thanks @bollwyvl.

Yeah I was hoping to get back to https://github.com/jupyterlite/jupyterlite/pull/854 and https://github.com/jupyterlite/pyodide-kernel/pull/2 before looking at this PR since it brings quite a lot of changes and could be tedious to port over :sweat_smile:

bollwyvl:
Welp, either way. If we're ready to start moving stuff over, that's fine, I can do the rework: really no complaints from me, as this of course would _improve_ the downstream situation.

app/repodata.schema.v0.json
```diff
@@ -0,0 +1,107 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
```
jtpio, line 2: Wondering how this file will be handled once the Pyodide kernel is moved to a separate repo? (https://github.com/jupyterlite/jupyterlite/pull/854)

bollwyvl: It only gets used by `jupyter lite check`, and wouldn't _strictly_ need to be shipped with the app (or, now, extension).

We'd also move the changed chunks of `litePLuginSettings` out of the main schema into the dedicated `check`... I _do_ think it's worth shipping the main schema with the as-deployed app.

bollwyvl:
Moved the parent issue to https://github.com/jupyterlite/pyodide-kernel.  

Humorously, the previous number (#914) became https://github.com/jupyterlite/pyodide-kernel/issues/14

jtpio:
Closing as this will more likely happen in the `pyodide-kernel` repo: https://github.com/jupyterlite/pyodide-kernel.

Thanks!

Issue #993, opened 2023-03-07
Reorganizing packages and dependencies


bollwyvl:
### Problem

With #854  close to merge, we've introduced some new complexity. We probably need to think more about both near- and further term packaging to avoid sacrificing flexibility _or_ compatibility.

### Proposed Solution

Definitely a strawman, but: targeting a notional `0.1.1b0` release, split the package on PyPI, but keep in this repo:

#### `jupyterlite-core 0.1.1b0`

This package would allow for really minimal or custom sites (such as kernels) to avoid features (e.g. kernels) they didn't ask for.

  - requires
    - `doit`
    - `jupyter-core`
  - provides 
    - `jupyter lite` CLI entrypoint
    - constants 
    - all actual addon implementations
    - core `app` bundle (with all `lab`, `retro` and `repl`)
      - splitting them is a pain, and probably not worth the squeeze
    - all the `[extras]` (still as `[extras]`)
    - tests
    - the JS kernel
    - maybe replace `python-libarchive-c` with something more portable/maintained

#### `jupyterlite 0.1.1b0`

This package would allow for better enabling the 80% case of:

> i got some python notebooks that pip install stuff, and want a site

   - provides
     - shims for `.constants` and `.addons.base`
       - toss a _reasonable_ number of warnings (ideally, one) when these are imported
       - don't shim `.tests`
   - requires
     - `jupyterlite-core[lab] 0.1.0b0`
       - this would fulfill `[contents,translation,check]`
     - `jupyterlite-pyodide-kernel`
     - `[extras]` that match all the existing ones
   - _maybe_ requires
     -  `jupyterlab-widgets` (just for completeness of what we've invested a lot of time into)

### Additional context

In making these changes, we should try to find in-the-wild, published downstreams that extend e.g. `addons.base.BaseAddon`, and take pains not to break their stuff... if they're well behaved, they have been pinning to a specific release, so it's likely we wouldn't be able to install them directly, but a `canary` repo in this site might be able to handle some short-lived submodules/patches.

A future version, number to be determined, might drop the `jupyterlab-widgets` and `jupyter-pyodide-kernel`.

jtpio:
Thanks for summarizing this in a new issue :+1: 

> targeting a notional 0.1.1b0 release

Why not just continue with `0.1.0b19` and then soon after start targeting rc and then finally 0.1.0. So we can finally use a proper version number.

[...]

PR #994, opened 2023-03-08
Add the `jupyterlite-core` Python package

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/993

## Code changes

- [x] Move `py/jupyterlite` to `py/jupyterlite-core`
- [x] Add `py/jupyterlite` metapackage
- [x] Adjustments
- [x] Try fixing `dodo.py`
- [x] Update CLI addon docs to mention `jupyterlite_core`
- [x] Update the releaser config to handle the new package (versioning and building of the dist files)
  - [x] Make sure it tries to upload both Python packages to the test PyPI Server
  - [x] Check handling of `PYPI_TOKEN` used by the releaser to be able to publish `jupyterlite-core` on PyPI

<details>

</details>

## User-facing changes

None

None


jtpio:
Opening early to get CI to run.

py/jupyterlite/src/jupyterlite/__init__.py
```diff
@@ -1,4 +1,7 @@
-"""tools for building JupyterLite sites
+"""re-export the core constants and addons from the core package.
 """
 
 __version__ = "0.1.0b18"
+
+from jupyterlite_core.addons.base import *  # noqa
```
bollwyvl, line 6: welp, as we've had to shim _a couple things_ around here... probably need individual files to properly emulate the modules. star imports are fine. probably _do_ want a warning.

dodo.py
```diff
@@ -423,31 +421,33 @@ def task_dist():
...
-            "--no-index",
-            "--find-links",
-            B.DIST,
-            C.NAME,
-        ]
-    else:
-        cwd = P.PY_SETUP_PY[C.NAME].parent
-        file_dep = [cwd / "src" / C.NAME / B.APP_PACK.name]
-        args = [*C.FLIT, "install", "--pth-file", "--deps=none"]
+    for py_name in [C.NAME, C.CORE_NAME]:
```
bollwyvl, line 424: doing this in a loop doesn't work very well for pip... perhaps generate a concrete `requirements.txt` that lists all of the things needed in a loop, and then have a single `pip` call that runs that?

jtpio: yeah I guess. Hoping we could just do `pip install -e .` at some point (#684) but that would be for later.

py/jupyterlite-core/pyproject.toml
```diff
@@ -91,7 +91,7 @@ all = [
     "tornado >=6.1",
 ]
 
-[project.entry-points."jupyterlite_core.addon.v0"]
+[project.entry-points."jupyterlite.addon.v0"]
```
bollwyvl, line 94: yeah, this would be _super_ breaking downstream

jtpio:
Looks fine at first glance on ReadTheDocs: https://jupyterlite--994.org.readthedocs.build/en/994/_static/lab/

Still need to re-enable `pyflakes`, and double check other details like the dependencies on the extra `[lab]`. We can also iterate more in separate PRs.

dodo.py
```diff
@@ -436,18 +431,30 @@ def task_dev():
...
+        )
     else:
-        cwd = P.PY_SETUP_PY[C.NAME].parent
-        file_dep = [cwd / "src" / C.NAME / B.APP_PACK.name]
-        args = [*C.FLIT, "install", "--pth-file", "--deps=none"]
+        for py_name in [C.NAME, C.CORE_NAME]:
+            py_name_pkg = py_name.replace("-", "_")
+            cwd = P.PY_SETUP_PY[py_name].parent
+            file_dep = [cwd / "src" / py_name_pkg / B.APP_PACK.name]
+            args = [*C.FLIT, "install", "--pth-file", "--deps=none"]
```
jtpio, line 451: I don't recall why we have two ways of installing the Python packages depending on whether it's running locally or on CI.

jtpio: We don't need to fix this in this PR but at some point it would be nice to enable more familiar for installing the packages like `python -m pip install -e .`.

bollwyvl: > two ways of installing the Python packages depending on whether it's running locally or on CI.

In CI, testing against as-built assets is a general way to avoid "surprises," which get amplified with more complex, multi-language packages and, of course, various build backends.

For a while `pip` broke editable installs by anything other than `setuptools-meta` (and somehow blamed the backends), so I haven't been too excited to switch back.

dodo.py
bollwyvl: If going with `pip`, with two inter-dependent packages it would still make sense to have a `requirements-editable.txt` with something like:

```
-e py_packages/jupyterlite_core
-e py_packages/jupyterlite
```

But then potentially turn on `--no-build-isolation` to avoid some of the "surprises" mentioned elsewhere.

dodo.py
jtpio: Sure. I mean we already have a dist task and this is for dev.

Anyway not gonna look into this now.

jtpio:
I think we can make a first release of `jupyterlite-core` after merging this PR, manually and probably without any tag on the repo for now . So it's easier to test the downstreams and help them switch to the new package.

For example starting with `jupyterlite-core==0.1.0a0` to leave some space for iterations. We could always yank these test releases on PyPI later if needed after re-aligning the version with the `jupyterlite` package.

jtpio:
> which is relatively hard to customize for humans to read.

The changelog should allow for manual edits. Meaning we can manually edit it to highlight some changes, for example under a specific section. The automated changelog entries should only affect the content between the start and end markers.

dodo.py
bollwyvl: presumably covered by below punt

bollwyvl (approved):
:shipit:

Issue #996, opened 2023-03-08
Provide more tools for downstream addons 


bollwyvl:
### Problem

Downstreams still have to do kind of a lot of work to do some things:
- find all `jupyter-lite.(ipynb|json)`
  - validate a specific `litePluginSettings` entry without reproducing the whole schema or digging through notebook meta
- ??? 

### Proposed Solution

- provide (and use) some more utility functions on `BaseAddon`
  - `LiteManager.all_lite_config_files`
    - should only return ones that exist
  - `LiteManager.all_output_config_files`
    - these may or may not exist
  - `LiteManager.get_lite_plugin_settings(path, plugin_id)`
  - `LiteManager.set_lite_plugin_settings(path, plugin_id, settings)`
  
### Additional context

- may prototype on https://github.com/jupyterlite/pyodide-kernel/pull/19
- probably blocked by
  - #994

bollwyvl:
Another solid tool (more strictly speaking) for downstreams would be a `py.typed` (but likely only in `jupyterlite-core`) and adding `mypy` to the CI mix: yep, it will make things slower, but will help a lot as we make more changes. As we have a relatively small number of actual dependencies, really only `doit` stands out as a particularly hard-to-typecheck API, though can be made pretty robust to consume with a `TypedDict` for task and a task generator.

Doing _good_ typing of `pytest` is a bit of a chore, but also generally worth it.

PR #998, opened 2023-03-09
Remove code references to pyolite, add BaseAddon methods

### LINKED ISSUE #995: Remove more of the `pyolite` hangover
### Problem

After #854, there are still some dangling references to `pyolite` that will cause confusion on the docs, etc.

### Proposed Solution

Remove the dangling references:

- constants ([renamed downstream](https://github.com/jupyterlite/pyodide-kernel/pull/19/files#diff-e90896b3695e1a7c57b585dc7c63ca9f80329d4ea5dc1935523c63fa30c8928aR14))
- schema entries (https://github.com/jupyterlite/pyodide-kernel/pull/19/commits/015e1a390e88a6012b66ee2a9a53989be285ee57)
- kernelspec/icon (https://github.com/jupyterlite/pyodide-kernel/pull/19#issuecomment-1462252729)

Update docs:

- use the new in-place schema from `jupyterlite-pyodide-kernel`
  - describe migration approaches

### Additional context

- probably blocked by 
  - #994 
  - https://github.com/jupyterlite/pyodide-kernel/pull/19

bollwyvl:
## References

- probably keep in draft until #994
- blocked by https://github.com/jupyterlite/pyodide-kernel/pull/19
- fixes #995
- first part of #996
  - pulls the new settings/paths methods from https://github.com/jupyterlite/pyodide-kernel/pull/19

## Code changes

- [x] `app`
  - [x] remove the `pyolite` stanza from schema
  - [x] remove the `piplite` schema
  - [x] remove the python kernel icon
- [x] `BaseAddon`
  - [x] adds `lite` and `output` config path finders
  - [x] adds getter/setter for lite plugins by id
- [x] `dodo.py`
  - [x] remove piplite references, task

## User-facing changes

- n/a (`pyodide-kernel` uses upstream branding)

- actually, kinda as these were previously part of the API


bollwyvl:
Yep, kernel icon definitely broken! No need to look at the UI report.

image


Just means screenshot-based tests in the docs site would be coupled with tightly pinned downstream versions...  certainly possible that e.g. the pyodide logo might change, if only to add a dark-mode-friendly version.

jtpio:
> Just means screenshot-based tests in the docs site would be coupled with tightly pinned downstream versions... certainly possible that e.g. the pyodide logo might change, if only to add a dark-mode-friendly version.

Right. Same goes with the p5 kernel. I think the screenshot helped catch a regression in the past with federated kernels not loaded.

I would say it's fine if the Pyodide kernel changes later and we need to update here. Normally it shouldn't change that often after reaching a good version.

jtpio:
Opened https://github.com/jupyterlite/jupyterlite/pull/1000 for now to help with this.

bollwyvl:
>  drop the Pyodide kernel from the UI tests

Even in another repo, and even leaving space for other repos, I think "breaks the latest (or a pre-)release of `pyodide-kernel`" is still pretty much "can't go to space today."

bollwyvl:
> JupyterLab that depends on

Dude, on LSP we try to keep _x_ kernels + _y_ language servers working under automated. It's awful. But that's how it goes to build tools not wrapped around "the one true" implementation.

bollwyvl:
This doesn't update any docs, but is probably still ready for looking at.

jtpio (approved):
Thanks!

jtpio:
Will publish a `jupyterlite-core==0.1.0a1` for testing so it can be used in https://github.com/jupyterlite/pyodide-kernel.

PR #1001, opened 2023-03-10
Update `jupyterlite` dependencies and docs

jtpio:
## References

Update the dependencies for the `jupyterlite` metapackage.

Follow-up to https://github.com/jupyterlite/jupyterlite/pull/994.

Part of #993 

This will be a follow-up to https://github.com/jupyterlite/pyodide-kernel/pull/23.

## Code changes

- [x] Update `jupyterlite-core` dependency
- [x] Depend on `jupyterlite-pyodide-kernel`
- [x] Streamline more CI with `jupyterlab/maintainer-tools/.github/actions/base-setup@v1` (remove some logic already handled by the action)
- [x] Test on Python 3.8
- [x] Update docs to mention `jupyterlite-core` and how to install the Pyodide kernel

## User-facing changes

This will ensure end users depending on `jupyterlite` will still get the Pyodide kernel by default for now.

None

### DIFF
--- a/.github/actions/build-dist/action.yml
+++ b/.github/actions/build-dist/action.yml
@@ -20,7 +20,6 @@ runs:
 
     - name: Install
       shell: bash
-      if: steps.cache-node-modules.outputs.cache-hit != 'true'
       run: doit setup:js
 
     - name: Build (js)
--- a/docs/howto/configure/advanced/extensions.md
+++ b/docs/howto/configure/advanced/extensions.md
@@ -33,18 +33,15 @@ site from an environment with already installed extensions.
 All third-party extensions, and some provided by JupyterLite and JupyterLab, can be
 disabled for one or all apps deployed in a site with the `disabledExtensions` option.
 
-For example, a site that doesn't use the Pyodide-based kernel can disable the
-[`ServiceWorker`](./service-worker.md) and the [Pyodide](../../python/pyodide.md) kernel
-with the following `jupyter-lite.json`:
+For example, a site that doesn't need the content to be available from the Python kernel
+can disable the [`ServiceWorker`](./service-worker.md) plugin with the following
+`jupyter-lite.json`:
 
 ```json
 {
   "jupyter-lite-schema-version": 0,
   "jupyter-config-data": {
-    "disabledExtensions": [
-      "@jupyterlite/pyodide-kernel-extension:kernel",
-      "@jupyterlite/server-extension:service-worker"
-    ]
+    "disabledExtensions": ["@jupyterlite/server-extension:service-worker"]
   }
 }
 ```
@@ -82,7 +79,7 @@ If detected, [`libarchive-c`](https://pypi.org/project/libarchive-c) will be use
 better performance, especially when working with archives with many/large assets,
 espcially [pyodide](../../python/pyodide.md).
 
-If not `libarchive-c` is not detected, python's built-in `zipfile` and `tarfile` modules
+If `libarchive-c` is not detected, Python's built-in `zipfile` and `tarfile` modules
 will be used.
 
 ```{warning}
--- a/docs/howto/configure/simple_extensions.md
+++ b/docs/howto/configure/simple_extensions.md
@@ -13,7 +13,7 @@ also work with JupyterLite.
 
 ### Creating a new environment
 
-The easiest way to add new extensions is to use the `jupyterlite` CLI in a Python
+The easiest way to add new extensions is to use the `jupyterlite-core` CLI in a Python
 environment where extensions have already been installed.
 
 You can choose the tool of your choice to manage these dependencies, such as `pip`,
--- a/docs/howto/configure/simple_extensions.md
+++ b/docs/howto/configure/simple_extensions.md
@@ -13,7 +13,7 @@ also work with JupyterLite.
 
 ### Creating a new environment
 
-The easiest way to add new extensions is to use the JupyterLite CLI in a Python
+The easiest way to add new extensions is to use the `jupyterlite-core` CLI in a Python
 environment where extensions have already been installed.
 
 You can choose the tool of your choice to manage these dependencies, such as `pip`,
--- a/docs/howto/configure/simple_extensions.md
+++ b/docs/howto/configure/simple_extensions.md
@@ -13,7 +13,7 @@ also work with JupyterLite.
 
 ### Creating a new environment
 
-The easiest way to add new extensions is to use the `jupyterlite-core` CLI in a Python
+The easiest way to add new extensions is to use the JupyterLite CLI in a Python
 environment where extensions have already been installed.
 
 You can choose the tool of your choice to manage these dependencies, such as `pip`,
--- a/docs/howto/extensions/frontend.md
+++ b/docs/howto/extensions/frontend.md
@@ -82,7 +82,7 @@ If you iterate and make new changes to the extension:
   https://user-images.githubusercontent.com/591645/171583522-f5677259-b91a-4ab0-8812-9770807a088e.gif
 
 ```{note}
-The `jupyterlite-core` package does not install any kernel by default.
+By default the `jupyterlite-core` package only includes a JavaScript kernel running in a Web Worker.
 If you would like to have a Python kernel available in your test JupyterLite deployment, you can install the `jupyterlite-pyodide-kernel` package with `pip`, or add it to your list of dependencies.
 ```
 
--- a/docs/howto/extensions/frontend.md
+++ b/docs/howto/extensions/frontend.md
@@ -51,9 +51,10 @@ JupyterLab.
 By default JupyterLite is able to find JupyterLab extensions installed in the same
 environment under `PREFIX/share/labextensions`.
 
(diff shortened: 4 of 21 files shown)


jtpio:
Pending https://github.com/jupyterlite/pyodide-kernel/pull/23 and a release of `jupyterlite-pyodide-kernel`.

jtpio:
Since we're requiring Python 3.8+ in `jupyterlite-pyodide-kernel` (https://github.com/jupyterlite/pyodide-kernel/blob/c3f1d74a48510f4fa39f52b6c04312e946090be7/pyproject.toml#L25), I would say we should just to the same here as well.

jtpio:
OK I think this should be good to go now.

Planning to make a new 0.1.0b19 release of both `jupyterlite-core` and `jupyterlite` from the repo after the merge.

Some follow-ups items for later:

- another pass on the docs
- remove the `0.1.0-beta` temporary tag and revert https://github.com/jupyterlite/jupyterlite/pull/997
- finish updating the downstreams to depend on `jupyterlite-core`

dodo.py
```diff
@@ -425,29 +425,32 @@ def task_dev():
...
+            "-e",
+            "./py/jupyterlite",
+            "--no-deps",
+        ]
         yield dict(
-            name=f"py:{py_name}",
-            actions=[U.do(*args, cwd=cwd)],
-            file_dep=file_dep,
+            name="py:jupyterlite",
+            actions=[U.do(*metapackage_args, cwd=P.ROOT)],
```
bollwyvl, line 453: removing and existing `file_dep` is a recipe for having to re-run stuff over and over again with no reason, and probalby out of order.

docs/howto/extensions/frontend.md
```diff
@@ -80,6 +81,11 @@ If you iterate and make new changes to the extension:
 [apod-tutorial]:
   https://user-images.githubusercontent.com/591645/171583522-f5677259-b91a-4ab0-8812-9770807a088e.gif
 
+```{note}
+The `jupyterlite-core` package does not install any kernel by default.
```
bollwyvl, line 85: still does JS?

docs/howto/configure/simple_extensions.md
```diff
@@ -13,7 +13,7 @@ also work with JupyterLite.
 
 ### Creating a new environment
 
-The easiest way to add new extensions is to use the `jupyterlite` CLI in a Python
+The easiest way to add new extensions is to use the `jupyterlite-core` CLI in a Python
```
bollwyvl, line 16: are we actually renaming the CLI? that will break a lot of things.

.github/actions/build-dist/action.yml
```diff
@@ -20,7 +20,6 @@ runs:
 
     - name: Install
       shell: bash
-      if: steps.cache-node-modules.outputs.cache-hit != 'true'
```
bollwyvl, line 23: :crying_cat_face: caching `node_modules` (provided `.eslint` and `.prettier` don't muck about in there, as they do by default for some reason) has been really stable, and saves an appreciable real amount of time...  meanwhile, without a way to clean out the yarn cache, it just gets bigger, and misses a lot.

it _does_ opens us up for e.g. the `svgo` issues, which seem to happen _anyway_ when the runners change their `apt` deps, i guess.

docs/howto/extensions/frontend.md
jtpio: Right, will add it back to the docs.

docs/howto/configure/advanced/extensions.md
```diff
@@ -82,7 +79,7 @@ If detected, [`libarchive-c`](https://pypi.org/project/libarchive-c) will be use
 better performance, especially when working with archives with many/large assets,
 espcially [pyodide](../../python/pyodide.md).
 
-If not `libarchive-c` is not detected, python's built-in `zipfile` and `tarfile` modules
+If `libarchive-c` is not detected, Python's built-in `zipfile` and `tarfile` modules
```
bollwyvl, line 82: we should move away from `libarchive-c`: outside of `conda/mamba`, it is a pain to install... i don't know what to replace it with, though... the _whole idea_ is to be able to use binary tools for vetted performance and security, but every one of them carries complexity risks.

docs/howto/configure/simple_extensions.md
jtpio: no it was just to avoid confusion about the package name.

We could say "the JupyterLite CLI" instead to remove ambiguity.

py/jupyterlite/pyproject.toml
```diff
@@ -9,9 +9,7 @@ authors = [
 ]
 dependencies = [
     "jupyterlite-core",
-    # TODO
-    # "jupyterlite-core[lab]",
-    # "jupyterlite-pyodide-kernel",
+    "jupyterlite-pyodide-kernel >=0.0.4",
```
bollwyvl, line 12: why not also the compatible lab?

dodo.py
jtpio: yeah probably we can add it back. This is likely a leftover of fighting with the doit tasks

.github/actions/build-dist/action.yml
jtpio: I don't think this one was used anyway, the `jupyterlab/maintainer-tools/.github/actions/base-setup` action above should already handle caching.

py/jupyterlite/pyproject.toml
jtpio: I guess we could indeed include more dependencies as mentioned in https://github.com/jupyterlite/jupyterlite/issues/993 (like ipywidgets). Maybe we can first see how it goes and add them later.

dodo.py
jtpio: Actually I don't know, this task is just here so we can simply install the packages in editable mode...

.github/actions/build-dist/action.yml
jtpio: For `svgo` maybe we should remove it from the CI / dodo tasks. I don't know if we need to optimize the images all the time, this could be done occasionally when needed. And also setting up a fresh new env almost always leaves the git repo in a dirty state with some icon files being modified.

dodo.py
bollwyvl: Right: but it would _at least_ depend on:
- `pyproject.toml` (changing `entry_points`, `data_files`, and a raft of other things invalidates an editable install)
- anything indirectly consumed _by_ `pyproject.toml` (e.g. `_version.py` or wherever the source of truth is these days)

.github/actions/build-dist/action.yml
bollwyvl: >  should already handle caching.

... of the ever-expanding, never-pruned `~/.cache/yarn`, which then _still_ requires compiling binaries from source, sometimes, and invoking the solver.

> for `svgo`

My point was "_any_ crazy `node_modules`," but `svgo` in specific _could_ be replaced with the stable, but slightly less efficient pure-python `scour`. But maybe more closed-form, and would cause less thrash.

Anything that "only gets run sometimes,".... doesn't.

py/jupyterlite/pyproject.toml
bollwyvl: My thinking here: the "happy path" of one of the most complex dependency issues, relying on the users' package manager of choice to make `sys.prefix` whole, will start getting less happy as extensions don't install/solve correctly because of jupyterlab 4... and `[extras]` are basically broken. 

A future `jupyterlab-core` can then depend on `jupyterlab >=4,<4.1`, even when we maintain the `3.5` compatible one.

I like widgets for this same reason: we can support `ipywidgets >=7.6,<8`, but there's no guarantee all future versions will have compatible `jupyterlab_widgets` releases, and getting kernels up to supporting widgets represents a _huge_ amount of effort (right next to arbitrary python packages, and iteratively reducing friction there).

docs/howto/configure/simple_extensions.md
bollwyvl: Maybe

> _the `jupyter lite` CLI_

?

py/jupyterlite/pyproject.toml
jtpio: Let's track that in a separate issue about which package to include in `jupyterlite` by default.

jtpio: Opened https://github.com/jupyterlite/jupyterlite/issues/1003

docs/howto/configure/advanced/extensions.md
jtpio: Let's track that in an issue?

.github/actions/build-dist/action.yml
jtpio: Anyway that's off topic, I opened https://github.com/jupyterlite/jupyterlite/issues/1004 to discuss this more.

dodo.py
jtpio: I'm planning to have another look at this as part of https://github.com/jupyterlite/jupyterlite/issues/684

jtpio:
Getting this one to unblock other work and we can address follow-ups in other PRs.

PR #1013, opened 2023-03-16
Distribute `jupyterlite-javascript-kernel` as a federated kernel

### LINKED ISSUE #881: Migrate to `hatch`
### Problem

Most Jupyter projects have now switched to using `hatch` for packaging: https://hatch.pypa.io/latest/

To keep things consistent and limit the number of different tools, we could consider switching to `hatch` here as well. And use its plugin system to extend functionality if needed.

### Proposed Solution

- [ ] Switch from `flit` to `hatch`
- [ ] Use `hatch` for versioning instead of `tbump` (if possible)

### Additional context

There is also a post of this on the Jupyter blog: https://blog.jupyter.org/packaging-for-jupyter-in-2022-c7be64c38926

jtpio:
## References

Follow-up to #994 and https://github.com/jupyterlite/jupyterlite/issues/993

Fixes #881 for consistency with other packages and https://blog.jupyter.org/packaging-for-jupyter-in-2022-c7be64c38926.

Following up the creation of a `jupyterlite-core` package and the extraction of the `jupyterlite-pyodide-kernel` package, it would make sense to also do the same for the JavaScript kernel.

We can keep it in the repo for now to ease development and have at least one kernel available when working on JupyterLite core.

## Code changes

- [x] Bootstrap `py/jupyterlite-javascript-kernel` from https://github.com/jupyterlite/serverlite-cookiecutter-ts
- [x] Cleanup files and config generated by the cookiecutter
- [x] Move the extension code to `py/jupyterlite-javascript-kernel/src`, similar to [`jupyterlab-widgets`](https://github.com/jupyter-widgets/ipywidgets/tree/main/python/jupyterlab_widgets/src) ? Otherwise setting the `outputDir` to `py/jupyterlite-javascript-kernel/jupyterlite_javascript_kernel/labextension` might be enough and simpler.
- [x] Add `jupyterlite-javascript-kernel` as a dependency of the `jupyterlite` metapackage, at least for now but can probably be removed in 0.2.0 (to be tracked in a separate issue).
- [x] Update the user facing docs
- [x] Streamline the Python package to be similar to other Jupyter projects (no nested `src`). Also this removes confusion with the layout of the `py/jupyterlite-javascript-kernel` package which uses `src` for the TypeScript source (like the cookiecutter)
- [x] Update the dev docs
- [x] Fix dev setup to be able to develop the kernel
- [x] Include the JS kernel on the demo site
- [x] Check releaser assets

## User-facing changes

Site deployers using `jupyterlite-core` will not get the JavaScript kernel by default anymore.

This will avoid having to disable the `@jupyterlite/javascript-kernel-extension` in a `jupyter-lite.json` for many deployments.

No JavaScript kernel available by default with `jupyterlite-core`.


bollwyvl:
This is a great step. 

With JS independently installable, we could potentially hoist more stuff here from `pyodide-kernel`, as, strictly speaking, if it's in the browser, it's _all_ JS at the end of the day, and could nail down more things like contents/IDrive files, comlink boilerplate, and egonomics (see https://github.com/jupyterlite/pyodide-kernel/issues/32).

While we're _not_ in nodejs, having a shim for _parts of_ the `path` nodejs stdlib module that made it easy to get to contents/IDrive shared with the worker would be good... not _everything_ needs to be `fetch`, and not everything _can_ be `import`ed (though imports _should_ be made to work for relative paths). No magic WebPack syntax, though, please! 

For `jyve`, i looked real hard at repurposing the [ijavascript](https://github.com/n-riesco/ijavascript) stack, but that might be a bridge too far. But, either way, getting to a `jupyterlite-typescript-kernel` would be a _big_ deal, as well as other "almost JS" languages like [civet](https://github.com/DanielXMoore/Civet) that are "easy" to compile in the browser.

Again, as a federated module (which could potentially move out of `packages` into `py_packages`, it would be in a closer state to serve as the baseline for being a wrapper kernel, and would have some of the same concerns as https://github.com/jupyterlite/pyodide-kernel/issues/28. 

Doing an in-tree `copier` in an extensible way, launched from the `jupyter lite` CLI, but shipped in `jupyterlite-core`, could have some "template" sites, and help site owners migrate versions.

As a standalone package, `jupyterlite-javascript-kernel` could add a `copier` that helped start/update/manage/migrate a wrapper kernel at the JS level, and this could be extended by `pyodide-kernel`, with the ability to extend all the "base" labextension boilerplate from `jupyterlite-javascript-kernel`.

requirements-docs.txt
```diff
@@ -5,6 +5,7 @@
 # build
 docutils >=0.18
 jupyter-server-mathjax
+jupyterlab >=3.5.3,<3.6
```
bollwyvl, line 8: already handled by `-> test -> build`? More lab specs in more places is harder to reason about...

jtpio: also all these `requirements-*` and `environment.yml` are quite confusing. Hoping we can do something about it as part of #684.

jtpio:
> With JS independently installable, we could potentially hoist more stuff here from `pyodide-kernel`, as, strictly speaking, if it's in the browser, it's _all_ JS at the end of the day, and could nail down more things like contents/IDrive files, comlink boilerplate, and egonomics (see [jupyterlite/pyodide-kernel#32](https://github.com/jupyterlite/pyodide-kernel/issues/32)).

Thanks these are good ideas.

Indeed streamlining the Web Worker approach would make things simpler for downstreams (#463). This is also what is currently happening in the xeus stack to make it easier to have more kernels compiled to Wasm and run in lite.

---

For now I think we can limit the scope of this PR to just getting `jupyterlite-javascript-kernel` out of the main bundle, to unblock 0.1.0 and other work. We can look more into this after as part of 0.2.0 or later.

Let's track these in separate issues.

requirements-docs.txt
bollwyvl: > all these requirements-* and environment.yml

At the expense of more downloads in CI, we could presumably go down to:

- `requirements.txt`
- `.binder/environment.yml`
  - really only `diffoscope` is not available on all platforms

It really would be nice to have slightly more control over the `pip` cache behavior than the current setup, as it doesn't seem to clean up very efficiently, so these just keep getting bigger.

> do something 

i sure hope the _something_ is not "use hatch for everything," as shoehorning a single-package build tool into a monorepo manager is not very enticing. Similarly, i continue to be unimpressed with `nx`... maybe it's awesome if for teams that use all the distributed cache stuff, but mostly just seems to rework stuff a lot, and get confused about its cache state.

jtpio: > "use hatch for everything,"

Not really no. We don't use `hatch` in a fancy way here, and we are also not using `hatch-jupyter-builder` (and not planning to). Just so that's it the same tool as in other Jupyter projects.

Probably more to be discussed in https://github.com/jupyterlite/jupyterlite/issues/684 and follow-ups, but I was more thinking about leveraging top-level scripts for building packages and run tasks. So folks could contribute without having to know `doit` or understand / debug the whole `dodo.py`.

jtpio:
Quick check on ReadTheDocs looking fine:

image

bollwyvl:
At the wrong machine: will take a look, but will be much later today, or
more likely or this weekend.

I think we need to hit a few of the known, but probably solvable, bugs like
upload >1mb, high Unicode paths, Json corruption, etc. before going out of
prerelease.

jtpio:
> bugs like upload >1mb, high Unicode paths, Json corruption,

Since those are bugs and have been around for quite a while, we don't need to block on them and they can be addressed in a 0.1.x.

.binder/environment.yml
```diff
@@ -54,9 +54,10 @@ dependencies:
   - python >=3.10,<3.11
   # build
   - doit >=0.34,<1
-  - flit >=3.7.1,<4
+  - hatch >=1.6.3,<2
   - jsonschema >=3
   - jupyter-server-mathjax >=0.2.3
+  - jupyterlab >=3.5.3,<3.6
```
bollwyvl, line 60: perhaps remove the pin above in the `demo` section so it isn't duplicated (without refining)

.github/workflows/build.yml
```diff
@@ -200,6 +200,10 @@ jobs:
         run: python3 -m pip install -r requirements-docs.txt
       - name: Install (py)
         run: doit dev:py:jupyterlite-core
+      - name: Build the JavaScript kernel
+        run: |
+          yarn
```
bollwyvl, line 205: this is where caching the in-tree `node_modules` can save 30s-1min, depending on the os.

but also: should have doit task.

.binder/environment.yml
```diff
@@ -54,9 +54,10 @@ dependencies:
   - python >=3.10,<3.11
   # build
   - doit >=0.34,<1
-  - flit >=3.7.1,<4
+  - hatch >=1.6.3,<2
```
bollwyvl, line 57: this seems like a larger-than-needed change. no hate on `hatch`, but `flit` can ship shared data just fine.

docs/howto/extensions/frontend.md
```diff
@@ -82,7 +82,7 @@ If you iterate and make new changes to the extension:
   https://user-images.githubusercontent.com/591645/171583522-f5677259-b91a-4ab0-8812-9770807a088e.gif
 
 ```{note}
-By default the `jupyterlite-core` package only includes a JavaScript kernel running in a Web Worker.
+By default the `jupyterlite-core` package does not include any kernel.
```
bollwyvl, line 85: Perhaps

```
any kernels.
```

docs/howto/extensions/frontend.md
```diff
@@ -82,7 +82,7 @@ If you iterate and make new changes to the extension:
...
 If you would like to have a Python kernel available in your test JupyterLite deployment, you can install the `jupyterlite-pyodide-kernel` package with `pip`, or add it to your list of dependencies.
```
bollwyvl, line 86: probably need to rep the JS kernel here

dodo.py
```diff
@@ -349,25 +349,27 @@ def task_build():
...
         if py_name == "jupyterlite-core":
             file_dep += [B.PY_APP_PACK]
 
-        pyproj_toml = py_pkg / "pyproject.toml"
-
         targets = [wheel, sdist]
+        actions = []
+        # make sure to build the lab extension before building the wheel
+        if py_name == C.JS_KERNEL_NAME:
+            actions += [U.do("yarn", cwd=py_pkg)]
```
bollwyvl, line 368: if keeping this as a single job, would need to add all of the `.ts`, `.json`, `.css`, etc.

ideally in CI it would use `--frozen-lockfile`.

dodo.py
```diff
@@ -819,7 +820,7 @@ class C:
     TESTING_IN_CI = json.loads(os.environ.get("TESTING_IN_CI", "0"))
     WIN_DEV_IN_CI = json.loads(os.environ.get("WIN_DEV_IN_CI", "0"))
     PYM = [sys.executable, "-m"]
-    FLIT = [*PYM, "flit"]
+    HATCH = [*PYM, "hatch"]
```
bollwyvl, line 823: is there special `hatch` things, or can we go to `pyproject-build`?

py/jupyterlite-javascript-kernel/pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+keywords = [
+    "jupyterlite",
+    "jupyter",
+    "jupyterlab",
+    "notebook",
+    "browser",
+    "javascript",
+]
+license = { file = "LICENSE" }
+requires-python = ">=3.7"
```
bollwyvl, line 22: 3.8

tsconfig.typedoc.json
```diff
@@ -24,9 +24,6 @@
     {
       "path": "./packages/javascript-kernel"
     },
-    {
```
bollwyvl, line 27: this seems like a big loss, especially if we're considering making this a more prominent part of the dependency tree.

bollwyvl:
I will again stand up for having python modules not importable by default in the root of a package: just because other jupyter projects do it, doesn't mean it hasn't caused those same projects issues during test and shipping time which then users get the pleasure of finding the bugs in missing files, etc.

Same tack on apparently starting to move away from `doit`: while one can complain about it, please point me to the one thing I run at the CLI that builds all of jupyterlab, its documentation, all its tests, and can restart in the middle of that, then switch branches, and have it fix itself. While the individual _tasks_ don't have to be in `dodo.py`, not having them integrated into a task DAG is basically nuts at scale... when we have something that works.

And, as mentioned, we can do more lifting of task content from well-known places in `package.json`, `pyproject.toml`... but `yarn this && jlpm that` and `cd there && jlpm that && hatch that` requires an order of magnitude more documentation.

dodo.py
jtpio: No `hatch` specific things. For now it's really just a matter of aligning with the other Jupyter projects as mentioned in the top comment:

> Fixes https://github.com/jupyterlite/jupyterlite/issues/881 for consistency with other packages and https://blog.jupyter.org/packaging-for-jupyter-in-2022-c7be64c38926.

So at least the configuration in `pyproject.toml` looks similar as for the other projects and one does not need to learn about a new tool. I guess we could consider `pyproject-build` if that simplifies even further, something to track in another issue?

py/jupyterlite-javascript-kernel/pyproject.toml
jtpio: Will do in a follow-up PR.

.binder/environment.yml
jtpio: See https://github.com/jupyterlite/jupyterlite/pull/1013#discussion_r1141895631

jtpio:
> I will again stand up for having python modules not importable by default in the root of a package: just because other jupyter projects do it, doesn't mean it hasn't caused those same projects issues during test and shipping time which then users get the pleasure of finding the bugs in missing files, etc.

What kind of issues exactly?

tsconfig.typedoc.json
jtpio: well for now it's just about extracting the JS kernel out of the main bundle.

Let's track that separately since this will likely take more time and thoughts to know what we are considering.

.github/workflows/build.yml
jtpio: > but also: should have doit task.

Not sure I'll have the bandwith and energy for this right now.

Also as mentioned before some doit tasks might or might not be executed depending on outside environment variable, which makes it very difficult to work with. Normally there should already be a task for building it already (should be `build:py:jupyterlite-javascript-kernel` I think) but it won't get triggered in the docs because of this environment variable.

jtpio:
Moving on with this to unblock follow-up work and pending draft PRs.

We can track the remaining items in other issues.

asd585140-glitch:
تحسين

.binder/environment.yml
asd585140-glitch: ضبط

Issue #1020, opened 2023-03-23
Adding custom extension to the demo site does not seem to include/activate it


stevejpurves:
## Description

I am trying to build [a demo site based on the jupyterlite demo template repo](https://github.com/stevejpurves/jupyterlab-myst-quickstart-lite). The aim is to include a prebuilt jupyterlab only frontend extension (`jupyterlab-myst`) and some demo notebooks to accompany it.

I started out by forking the template and then:

- added `jupyterlab-myst` to the `requirements.txt` file
- added notebooks to the `content` folder
- pushed to `main` and redeployed

when accessing the site https://stevejpurves.github.io/jupyterlab-myst-quickstart-lite/lab/index.html I can see the notebooks I added and open them - however he myst content does not render. I am not sure how to check that the `jupyterlab-myst` extension was loaded, or how to debug further.

The `juptyerlab-myst` extension overrides the default markdown renderer and is functional in `jupyterlab` -- should this be expected to work out of the box in jupyterlite?

## Reproduce
Check the deployed demo site here: https://stevejpurves.github.io/jupyterlab-myst-quickstart-lite/lab/index.html

## Expected behavior

I would expect the markdown cells that contain myst in the demo notebooks to render myst directives correctly.

## Context

- JupyterLite version: forked from latest demo repo
- Operating System and version: MacOS
- Browser and version: Chrome

<details><summary>Browser Output</summary>
On opening the notebook `01-jupyterlab-myst.ipynb` the follow appears in the JS console
<pre>
VM356 react_devtools_backend.js:2655 TypeError: Cannot set properties of undefined (setting 'text')
    at m._setValue (jsoneditor.ts:331:19)
    at set source [as source] (jsoneditor.ts:134:10)
    at e.NotebookMetadataEditorTool._update (notebooktools.ts:637:19)
    at e.NotebookMetadataEditorTool.onActiveNotebookPanelChanged (notebooktools.ts:623:12)
    at e.NotebookMetadataEditorTool.processMessage (notebooktools.ts:345:16)
    at c (index.ts:530:15)
    at Object.e [as sendMessage] (index.ts:220:7)
    at notebooktools.ts:172:7
    at f (iter.ts:227:9)
    at P._onActiveNotebookPanelChanged (notebooktools.ts:171:5)
overrideMethod @ VM356 react_devtools_backend.js:2655
s @ index.ts:467
c @ index.ts:532
e @ index.ts:220
[...]

PR #1021, opened 2023-03-23
Add `@jupyterlab/attachments` to the build

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1020

## Code changes

Add `@jupyterlab/attachments` so it ends up in the `resolutions` and the webpack shared scope.

## User-facing changes

This should prevent some JupyterLab extensions depending on the `@jupyterlab/attachments` package to fail to activate. For example `jupyterlab-myst`:

image


bollwyvl:
As noted before: adding a test that compares all the upstream
pins/singletons with ours (plus a known list of do-not-wants) would help
catch this, and is likely appropriate on this pr.

jtpio:
> As noted before: adding a test that compares all the upstream pins/singletons with ours (plus a known list of do-not-wants) would help catch this, and is likely appropriate on this pr.

Opened https://github.com/jupyterlite/jupyterlite/issues/1022 to track this.

Issue #1011, opened 2023-03-15
JSON error when opening from URL


JasonWeill:
## Description

Moved from https://github.com/jupyterlab/jupyterlab/issues/14194, opened by @oeway.

When opening a notebook via URL, JupyterLite displays a JSON error but JupyterLab does not.

## Reproduce

1. Launch JupyterLite here: https://jupyter.org/try-jupyter/lab/
2. Choose File > Open from URL…
3. Open https://raw.githubusercontent.com/InsightSoftwareConsortium/itkwidgets/00a7ae180f3abf3f90222ce514ad17a8e5ba0402/docs/jupyterlite/files/Hello3DWorld.ipynb
4. See error: 

>    Upload Error
>   JSON.parse: unterminated string at line 48 column 470833 of the JSON data

## Expected behavior

Notebook opens. (This happens with JupyterLab 4.0.0a35, when installed locally.)

## Context

- JupyterLite version: 0.1.0-beta.19
- Operating System and version: macOS 12.6.3 "Monterey"
- Browser and version: Firefox 102.8.0esr (64-bit)

vaddisrinivas:
Is this issue still open? I'd like to contribute here!

JasonWeill:
I was able to reproduce it just now. Please free to submit a pull request when you're ready! You can also keep the conversation going on this issue or on our other social channels, like [Gitter](https://gitter.im/jupyterlab/jupyterlab#) and [Discourse](https://discourse.jupyter.org/).

Issue #888, opened 2022-12-01
Data uploaded to the system gets trimmed, even if its bellow 2MB.


dgcmain:
## Description

Upload a .csv file and the data gets trimmed to a certain row, instead of show the full file like its supposed.

File sizes tested: 1.6mb, 1.8mb, 1.4mb

## Reproduce

1. Go to Upload Data, upload a csv file of more than 1.5mb
2. Open the file with the previewer (or a notebook) and see how the file is trimmed.

## Expected behavior

The file should load completely.

dgcmain:
@bollwyvl @jtpio do you consider this bug relevant?

jtpio:
Thanks @dgcmain for reporting.

Yes this seems like an issue with how JupyterLite handles file uploads.

rafpyprog:
I am facing the same problem. There's a work around if you are working with csv files and Pandas. Compress the file using zip to reduce the file size. Pandas is able to read the zip file using the _pd.read_csv_ method.

Issue #741, opened 2022-07-18
Can't upload CSV files > 1MB to Jupyterlite file system by drag and drop


tangocat73:
## Description

I'm trying to see if Jupyterlite demo site can process some of my local CSV files with pandas. I've noticed that when dragging and dropping these files in Jupyterlite file system UI, any file that larger than ~1MB would be truncated by the system; while the smaller ones <1MB can be processed correctly.   

Attached please see the screenshots for this issue: 

Screenshot 1. query.csv is a 1. 9MB earthquake dataset I have, but it was NOT read correctly by file system; a lot of rows were dropped during uploading process, and thus column headers were not processed correctly. 

Screenshot 2. query_2.csv is a 0.9 MB partial dataset I edited from query.csv, by deleting nearly half of the data; and Jupyterlite file system took it without any issue, with column heads processed correctly.  

Screenshot 3. reading these CSV files using Pandas.

I wonder if this is a limitation of the Jupyterlite demo website, so that users are now allowed to upload large files (but I'm under the impression that these local files are uploaded to users' browser storage...); or maybe there is some site setting I need to change. 

(note: I've also noticed that if I serve these csv files as static assets using my own Github Jupyterlite deployment, then file size would not matter). 

## Context

- JupyterLite version: v0.1.0b11
- Operating System and version: Windows 10
- Browser and version:  version 103.0.5060.114

## Screenshots

![query csv 1 9MB](https://user-images.githubusercontent.com/109368994/179566970-43ceedce-2ba0-435f-8668-fdadf1cc0707.JPG)
![query_2 csv 0 9MB](https://user-images.githubusercontent.com/109368994/179567010-1292fcfe-deb3-44c2-827d-4d1ded026bb7.JPG)
![Jupyterlite](https://user-images.githubusercontent.com/109368994/179567049-cd8bf8c8-2a86-4c4b-b60c-77fbfcafb27f.JPG)

bollwyvl:
> see the screenshots 

It would also be interesting to have some browser logs (e.g. revealed by `F12`) as well, as this might tell us where the problem is.

Recall: everything  in this free software, hosted on free servers, is still very much in beta, and the in-browser storage stuff is all best-effort, especially with respect to the less-than-month-old "magic" mirroring of files between kernels and contents.

Yep: to fix this for real, it would need to be under test in a real browser, and I haven't had to time to start building up a suite. Last a checked, galata took some shortcuts...

> serve these csv files as static assets using my own Github

[...]

PR #1024, opened 2023-03-24
Support chunked uploads

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/888
Fixes https://github.com/jupyterlite/jupyterlite/issues/741
Fixes https://github.com/jupyterlite/jupyterlite/issues/1011

## Code changes

JupyterLab defines a `CHUNK_SIZE` here:

https://github.com/jupyterlab/jupyterlab/blob/a3a0e3e7abf75e53810ed7ff8037c439cbb07697/packages/filebrowser/src/model.ts#L34

```ts
/**
 * The size (in bytes) of the biggest chunk we should upload at once.
 */
export const CHUNK_SIZE = 1024 * 1024;
```

This change adds support for chunked uploads in the JupyterLite contents manager.

## User-facing changes

Users should be able to upload files larger than 1MB.

None


jtpio:
Checking with a ~20MB CSV file is looking good on ReadTheDocs:

https://user-images.githubusercontent.com/591645/227598931-05228ea6-2749-4872-852e-6570395735f8.mp4

bollwyvl:
Awesome stuff 🤠

jtpio:
Did a bit more testing on ReadTheDocs and it seems to be fine.

Getting that one in so it can be released. Opened https://github.com/jupyterlite/jupyterlite/issues/1027 to track putting that under test.

PR #1031, opened 2023-03-31
Update docs site python dependencies, pyodide 0.23

bollwyvl:
## References

- companion PR to https://github.com/jupyterlite/pyodide-kernel/pull/39

## Code changes

- [x] use `jupyterlite-pyodide-kernel` from RTD
- [x] restore docs dependency discovery and wheel sync 
- [x] update federated extensions and pyodide wheels 

## User-facing changes

- docs site viewers
  - should see pyodide 0.23, python 3.11
- maintainers
  - should be able to keep the example extensions in sync with python packages

- n/a


dodo.py
```diff
@@ -1183,6 +1248,52 @@ def get_deps(has_deps, dep_file):
 
         dep_file.write_text(out, **C.ENC)
 
+    def deps_to_wheels(all_deps):
```
jtpio, line 1251: Maybe we should try to move the handling of the Pyodide examples to the `jupyterlite-pyodide-kernel` repository at some point. But still include them on the demo site (https://jupyterlite.rtfd.io/en/latest/).

Hoping we could avoid growing this `dodo.py` at some point, it feels more complex the more I look at it :(

bollwyvl: > move the handling

Sure, packaging the `depfinder` stuff is fine, and while it would be separately-released, also be co-developed in the kernel repo.

> avoid growing this dodo.py

Sure, that's fine: but it's _still_ a complex, long build, and relying on docs just isn't worth the time keeping the docs up-to-date, and telling people where to look for them when they hit a problem which we _know_ is a problem. And not all these things can be solved by more 

The few, non-pyodide-related `file_dep`s restored in this PR (e.g. "do a dev install when `pyproject.toml` changes", or "build the app bundle before running the unit tests"), mean one can run `doit check` and get a fully built docs site, which rebuilds correctly when things change. _That_, day-in, day-out simplicity is worth the complication in this one file. Heck, we should have the ui-tests in the loop as well, and use the output of those tests in documentation. More complex, but _breaks the build_ in a good way if things change too much.

I've been making some progress on a tool that moves a lot of the grunt work of defining tasks and dependencies to some declarative `pyproject.toml`, which could eventually be employed here.

dodo.py
```diff
@@ -1506,6 +1650,13 @@ def _ensure_resolutions(app_name):
 
         return all_up_to_date
 
+    def fetch_pyodide_repodata():
+        url = "https://cdn.jsdelivr.net/pyodide/v0.23.0/full/repodata.json"
```
jtpio, line 1654: Hmm does that means we also need to know the Pyodide version here? Which is already handled in https://github.com/jupyterlite/pyodide-kernel.

Looks like one more piece of configuration that will need to be updated next time we update the pyodide kernel.

bollwyvl:
Yeah should move those closer.

Or we actually install it, and grab it from the schema.

jtpio:
Closing as the pyodide kernel has been updated to 0.7.0 since.

Issue #1033, opened 2023-03-31
"e.cells is undefined" in uploaded ipynb files


chrysn:
## Description

When uploading a ipynb file (which was freshly downloaded) and opening it, a "File Load Error for (filename)" is shown, with the message "e.cells is undefined". The tab for the file flickers open only briefly, and is gone by the time the error is shown.

I've observed this both with Python (Pyodide) and JavaScript (Web Worker) notebooks.

## Reproduce

1. Go https://jupyterlite.readthedocs.io/en/latest/_static/lab/index.html
2. In the launcher, select a new Python or JavaScript notebook.
3. Right click on the file in the file browser tab, download it. (You could enter something and/or save it, makes no difference).
4. Close the file's tab.
5. Upload the file through the upload button in the file browser tab, confirm overwriting.
6. Double click on the (now re-uploaded) file.
7. See the error message.

## Context

I haven't observed this until today, and searching around in the issue tracker indicates this is new.

I've tried it with a freah browser instance.

- JupyterLite version: Whatever is currently running on jupyterlite.readthedocs.io (Help / About JupyterLite Examples shows 0.1.0-eta.22, but I don't think that's the relevant piece of info)
- Operating System and version: Debian sid (but I don't run anything locally, so it shouldn't matter)
- Browser and version: Firefox 111.0. uBlock Origin is running in the session that generated the output below, but I also tested with adblocking disabled.

<details><summary>Browser Output</summary>
<pre>
Unsatisfied version 6.0.2 from @jupyter-widgets/jupyterlab-manager of shared singleton module @jupyter-widgets/base (required ^1.1.10 || ^2 || ^3 || ^4.0.0) remoteEntry.03390b18ed36b245b514.js:1:6189
Unsatisfied version 6.0.2 of shared singleton module @jupyter-widgets/base (required ^1 || ^2 || ^3 || ^4) remoteEntry.8a3514e7c1c3c4a80f2e.js:1:5546
Unsatisfied version 6.0.2 of shared singleton module @jupyter-widgets/base (required ^1 || ^2 || ^3 || ^4) remoteEntry.0b6a37985538f1f2120b.js:1:5345
Registering new JupyterLite ServiceWorker https://jupyterlite.readthedocs.io/en/latest/_static/service-worker-b2fb40a.js service-manager.ts:50:16
JupyterLite ServiceWorker was sucessfully registered service-manager.ts:52:16
[...]

PR #1034, opened 2023-03-31
Fix chunk handling when uploading notebooks

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1033

Follow-up to #1024 

## Code changes

Fix handling of the `chunk` option before storing a notebook.

## User-facing changes

This should allow for uploading notebooks without error.

None


chrysn:
Thanks! Yes, works in my tests as well.

jtpio:
Thanks @chrysn for checking!

PR #1061, opened 2023-04-21
DOC Make documentation about source maps more accurate

lesteve:
## Code changes

I believe the correct config is something like below so I used `LiteBuildConfig/no_sourcemaps` similarly to no_unused_shared_packages above.

```json
{
    "LiteBuildConfig": {
        "no_sourcemaps": true
    }
}
```

The previous sentence led me to believe that the following would work (it doesn't):
```json
{
    "no_sourcemaps": true
}
```


jtpio (approved):
Thanks!

Issue #1074, opened 2023-05-09
JupyterLite sometimes crashes in broadcast.ts with a message without path


andeplane:
## Description

I cannot reproduce this at all times, but it is a problem occurring when JupyterLite is being embedded in an iframe. In short, we are planning to use JupyterLite to embed a notebook at my work. We are embedding the notebook as an iframe inside our webapp. 

This normally works very well, but sometimes the kernel crashes (see attached screenshot) because a message is being sent and picked up in [broadcast.ts](https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/broadcast.ts#L61), sometimes with `event.data = null`, and sometimes an object missing the `path` property. 

Sometimes, I also get `e.data` being [this object](https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/broadcast.ts#L114) which could make sense since we are doing `this._channel.postMessage(response);` [here](https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/broadcast.ts#LL157C5-L157C41).

<img width="1920" alt="Screenshot 2023-05-09 at 10 17 46" src="https://user-images.githubusercontent.com/16174/237039913-9a530154-d468-464a-b192-c4871623a80f.png">

<img width="1496" alt="Screenshot 2023-05-09 at 10 50 49" src="https://user-images.githubusercontent.com/16174/237045371-10f3866e-4c1d-4725-ae1d-675be487f481.png">

I have also gotten this problem with JupyterLite with no modifications, but it only happens inside an iframe, and not always (but when it happens, I can reproduce it 100% until it suddenly works again).

## Reproduce
I don't know a stable way to do it, but have JupyterLite embedded in an iframe is the only way I know, and it mostly works. No code change happens when it suddenly doesn't work anymore. This happens if I embed https://jupyterlite.readthedocs.io/en/latest/_static/lab/?path=intro.ipynb into the iframe as well.

## Expected behavior

JupyterLite kernel should not crash, but it sometimes does.

## Context

- JupyterLite version: latest as of today.
- Operating System and version: MacOS 13.0
- Browser and version: Chrome 113

So a couple of questions:
1) This is not deterministic, how can I even debug this?
2) Why is this only happening in an iframe?
3) Can a workaround be to ignore events without `path` in the `e.data` [here](https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/broadcast.ts#L57)?
4) Why is this happening?

[...]

PR #1075, opened 2023-05-09
Only handle broadcasted messages meant to be received

andeplane:
## References

Attempt to solve https://github.com/jupyterlite/jupyterlite/issues/1074 (also https://github.com/jupyterlite/jupyterlite/issues/1126 and https://github.com/jupyterlite/jupyterlite/issues/1089)

## Code changes

Ref https://github.com/jupyterlite/jupyterlite/issues/1074, sometimes `_onMessage` in [broadcast.ts](https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/broadcast.ts#L61) gets a message that does not meet the expected structure of the function. This PR makes sure these messages are ignored by this receiever.

### DIFF
--- a/packages/contents/src/drivefs.ts
+++ b/packages/contents/src/drivefs.ts
@@ -39,8 +39,6 @@ export type TDriveMethod =
   | 'get'
   | 'put';
 
-const SENDER = 'drivefs.ts';
-
 /**
  * Interface of a request on the /api/drive endpoint
  */
@@ -51,9 +49,9 @@ export interface IDriveRequest {
   method: TDriveMethod;
 
   /**
-   * The sender of the request ('drivefs.ts' or 'broadcast.ts')
+   * The expected receiver of the request
    */
-  sender: 'drivefs.ts' | 'broadcast.ts';
+  receiver?: 'broadcast.ts';
 
   /**
    * The path to the file/directory for which the request was sent
@@ -308,6 +306,7 @@ export class ContentsAPI {
   request(data: IDriveRequest): any {
     const xhr = new XMLHttpRequest();
     xhr.open('POST', encodeURI(this.endpoint), false);
+
     try {
       xhr.send(JSON.stringify(data));
     } catch (e) {
@@ -322,26 +321,17 @@ export class ContentsAPI {
   }
 
   lookup(path: string): DriveFS.ILookup {
-    return this.request({
-      sender: SENDER,
-      method: 'lookup',
-      path: this.normalizePath(path),
-    });
+    return this.request({ method: 'lookup', path: this.normalizePath(path) });
   }
 
   getmode(path: string): number {
-    const { mode } = this.request({
-      sender: SENDER,
-      method: 'getmode',
-      path: this.normalizePath(path),
-    });
-
-    return Number.parseInt(mode);
+    return Number.parseInt(
+      this.request({ method: 'getmode', path: this.normalizePath(path) })
+    );
   }
 
   mknod(path: string, mode: number) {
     return this.request({
-      sender: SENDER,
       method: 'mknod',
       path: this.normalizePath(path),
       data: { mode },
@@ -350,7 +340,6 @@ export class ContentsAPI {
 
   rename(oldPath: string, newPath: string): void {
     return this.request({
-      sender: SENDER,
       method: 'rename',
       path: this.normalizePath(oldPath),
       data: { newPath: this.normalizePath(newPath) },
@@ -358,30 +347,21 @@ export class ContentsAPI {
   }
 
   readdir(path: string): string[] {
-    const { contents } = this.request({
-      sender: SENDER,
+    const dirlist = this.request({
       method: 'readdir',
       path: this.normalizePath(path),
     });
-    contents.push('.');
-    contents.push('..');
-    return contents;
+    dirlist.push('.');
+    dirlist.push('..');
+    return dirlist;
   }
 
   rmdir(path: string): void {
-    return this.request({
-      sender: SENDER,
-      method: 'rmdir',
-      path: this.normalizePath(path),
-    });
+    return this.request({ method: 'rmdir', path: this.normalizePath(path) });
   }
 
   get(path: string): DriveFS.IFile {
-    const response = this.request({
-      sender: SENDER,
-      method: 'get',
-      path: this.normalizePath(path),
-    });
+    const response = this.request({ method: 'get', path: this.normalizePath(path) });
 
     const serializedContent = response.content;
     const format: 'json' | 'text' | 'base64' | null = response.format;
@@ -415,7 +395,6 @@ export class ContentsAPI {
       case 'json':
       case 'text':
         return this.request({
-          sender: SENDER,
           method: 'put',
           path: this.normalizePath(path),
           data: {
@@ -429,7 +408,6 @@ export class ContentsAPI {
           binary += String.fromCharCode(value.data[i]);
         }
         return this.request({
-          sender: SENDER,
           method: 'put',
           path: this.normalizePath(path),
           data: {
@@ -443,7 +421,6 @@ export class ContentsAPI {
 
   getattr(path: string): IStats {
     const stats: IStats = this.request({
-      sender: SENDER,
       method: 'getattr',
       path: this.normalizePath(path),
     });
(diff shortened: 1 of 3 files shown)


andeplane:
Tested the PR server in our environment now, and it seems to work, as a message is being ignored with this change. However, a better solution would be to figure out why the message is sent there in the first place.

image

martinRenou:
Thanks a lot for opening a PR, this was on my TODO list for weeks!

> However, a better solution would be to figure out why the message is sent there in the first place

I agree. Were you able to see which message was sent in the first place? This request should be built and sent in https://github.com/jupyterlite/jupyterlite/blob/2ee9cba529cb514c00130924c355d8696fd4562c/packages/contents/src/drivefs.ts#L301, but I don't see a reason for the content to be `null`. We may want to figure out what the message originated from.
We may also want to add there the check for data==null before sending the message. If we do the check in the service worker, I fear the kernel may fail and get stuck later on.

andeplane:
Yeah! I have some insight:

If you look at this screenshot, you see that the `e.data` is an object that matches the `response` being created [here](https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/broadcast.ts#L114), which seems to be sent later in the same `_onMessage` function [here](https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/broadcast.ts#L157).

I don't understand why it doesn't always fail after reading it, but I am definitely not understanding this properly yet.

image

martinRenou:
Ok 😕  Would you be able to put a `console.log("message", JSON.stringify(data))` in here: https://github.com/jupyterlite/jupyterlite/blob/2ee9cba529cb514c00130924c355d8696fd4562c/packages/contents/src/drivefs.ts#L304 ?

I would test it myself, I am not able to reproduce this unfortunately, I've only seen it on other people's env.

dlcognite:
When debugging this, what we noticed is that
- we always (no matter in iframe, or as standalone) get an event where the request.method is  `getattr` in the `_onMessage` of the broadcast.ts file.
- at the end of the file it will `_postMessage` to the `_channel` this body
```
{
        dev: 1,
        nlink: 1,
        uid: 0,
        gid: 0,
        rdev: 0,
        size: model.size || 0,
        blksize: BLOCK_SIZE,
        blocks: Math.ceil(model.size || 0 / BLOCK_SIZE),
        atime: model.last_modified, // TODO Get the proper atime?
        mtime: model.last_modified,
        ctime: model.created,
        timestamp: 0,
      }
      ```
      
      notice here that the body here doesn't have a `method`
      
     when in iframe, however the same channel would receive this new message missing "method", and fail. but when running this not in iframe, it works fine... so somehow in the iframe, the `_postMessage` comes back to the same broadcast channel...

**Comment by andeplane** [comment]
Ok, we think we know what it is. When two or more tabs are open, messages sent from `broadcast.ts` are being received in the other tab's `BroadcastChannelWrapper._onMessage`, which does not have the expected format. Normally you have one sender and one receiver on the `/api/drive.v1` channel, but it gets mixed up with multiple tabs.

The right solution (I think) is to have a [session id](https://github.com/jupyterlite/jupyterlite/blob/main/packages/kernel/src/kernel.ts#L18) that is being attached to all messages so the other receivers can ignore messages.

But on the other hand, since the file system is shared across all tabs it actually does make sense to let tabs communicate, in which case a slight modification this PR might be the right solution.

What do you think @martinRenou?

File: packages/contents/src/drivefs.ts [hunk 1]
```diff
@@ -343,6 +357,7 @@ export class ContentsAPI {
 
   readdir(path: string): string[] {
     const dirlist = this.request({
+      sender: 'drivefs.ts',
```
martinRenou, line 360: Could this be a `const SENDER = 'drivefs.ts'` at the top of the file?

martinRenou (commented):
Thanks for continuing exploring this!

It's unclear to me where the other message comes from.

andeplane:
> Thanks for continuing exploring this!
> 
> It's unclear to me where the other message comes from.

So if you have two tabs A and B, both running the notebook.
`A.broadcast.ts` is sending a message on the `/api/drive.v1` channel that is meant to be received by `A.drivefs.ts`. `A.drivefs.ts` receives it and all is good. But here is the thing, since this channel is shared across tabs, `B.drivefs.ts` receives it (which is ok), but `B.broadcast.ts` also receives it, and that's why it breaks because `B.broadcast.ts` expects messages with a different payload.

packages/contents/src/drivefs.ts
andeplane: Fixed that now

martinRenou (approved):
LGTM! @jtpio if you're fine with this PR, it would be great to make a patch release with it

jtpio:
Thanks all for looking into this!

Looks like the UI tests failure might be relevant though. 

Otherwise yes it would be great to get this fix released after.

andeplane:
Ok great, I’ll fix the ui tests.

jtpio:
Looking at the report, the UI tests seem to be stuck while executing the cell: 

image

jtpio:
So trying with the ReadTheDocs preview for this PR (https://jupyterlite--1075.org.readthedocs.build/en/1075/_static/lab/index.html) in a fresh new browser tab on Chromium gives the following error:

image

andeplane:
Yep, I was testing a thing that is in the latest commit. Will do some work tomorrow and fix UI tests and make sure it works.

andeplane:
Seems to be working, please verify :)

jtpio:
Thanks @andeplane for your work on this!

Looks like the ReadTheDocs preview is still giving the same error? (trying in a Chromium private window)

https://jupyterlite--1075.org.readthedocs.build/en/1075/_static/lab/index.html

andeplane:
> 

Yep I see that, will be sure to test with incognito. It seems like it breaks when we have the
`if (event.data?.sender !== 'broadcast.ts')` test, which means some messages are being sent elsewhere. Testing without it now, and will find out who that sender is.

andeplane:
@martinRenou back working on this now, seems like I need your approval to run workflows!

andeplane:
@jtpio I think I need some help here. I know for sure that UI tests start failing with the test on `event.data?.sender` which could make sense if the tests somehow communicate with the file system using the broadcast channel and doesn't set the sender. However the UI tests seems to use functions provided by JupyterLab?
```
const name = await page.notebook.createNew();

    await page.notebook.setCell(0, 'markdown', '## This is a markdown cell');
    await page.notebook.addCell('raw', 'This is a raw cell');
    await page.notebook.addCell('code', '2 + 2');

    await page.notebook.run();
    await page.notebook.save();
```

(I can't find e.g. `getCellTextOutput` anywhere in the source code for JupyterLite. 

Question is then why this change does anything to the tests (I can manually open notebooks and run them, so doesn't happen on _my_ browser, hehe).

jtpio:
@andeplane not sure it was only the tests failing. According to the comment above (https://github.com/jupyterlite/jupyterlite/pull/1075#issuecomment-1549085438) the failure seemed relevant and also happening on the ReadTheDocs demo.

Just kicked the CI, let's see if it passes.

andeplane:
Right, forgot it after all this time, sorry. I see that it is true on readthedocs. I can't reproduce locally (works when I build and run using gitpod), so the only reproducible environment I have is building here, so expect some noise while I debug through readthedocs 😂

But that is great news tbh, because then it isn't something weird like I was trying to claim above. I'm still confident the underlying issue is the messaging like explained earlier, so I'll keep digging. Thanks!

andeplane:
I figured out why it didn't work earlier (me being stupid was one good reason!), and changed the logic so we mark the messages being sent using serviceWorker to broadcast.ts with `receiver: 'broadcast.ts'`, so all messages being sent to itself will be ignored. 

This now works locally, and I can open multiple tabs with [ReadTheDocs](https://jupyterlite--1075.org.readthedocs.build/en/1075/_static/lab/index.html), and the tests pass. 

Would love for you folks to give it a go as well.

jtpio (approved):
Thanks!

jtpio:
Planning to make a patch release with this change on Monday 21.

ogrisel:
Feel free to ping me once this is done so that I can trigger a CI update for the environment used to reproduce https://github.com/jupyterlite/jupyterlite/issues/1126 so that we can confirm that this can close it.

jtpio:
@ogrisel `0.1.2` has been released: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.2

PR #1084, opened 2023-05-15
Trigger CI

martinRenou:
Trying to reproduce the UI-test issue in #1075


martinRenou:
Yeah, I was verifying that master was not impacted as well. I'm surprised #1075 breaks it.

PR #1109, opened 2023-07-17
Resolve the null getter issue in the _onMessage method

sko-kr:
## References
Fixes null getter error as can be seen below;
image
image

## Code changes
Modify the sections of code in the _onMessage function that disregard the possibility of `event.data` being `undefined`.

## User-facing changes
This update doesn't bring about any visible changes for the regular user, it simply eliminates the error log that was previously visible, as depicted in the images above.

None, as it only adds a nullish checks.


martinRenou (approved):
Thanks!

Issue #1069, opened 2023-04-27
Spaces in Files Folder Name Causes Build Failure


theonewolf:
## Description

In my `files` folder, if I have subfolders with spaces in the names, by `jupyter lite build` fails and says it can't find the folder with what look like URL-encoded space values.

Example:

![Screenshot 2023-04-26 at 9 46 41 PM](https://user-images.githubusercontent.com/522248/234739141-680c2524-d941-402b-8ecf-1b547eb7d933.png)

The copy succeeds, as the above screenshot looks like.

![Screenshot 2023-04-26 at 9 46 51 PM](https://user-images.githubusercontent.com/522248/234739101-37f21236-f531-4c15-9781-4d36667f89b2.png)

But it fails to find the folder later in the build.

## Reproduce

1. Create a `files` folder
2. Create a `files\spaces in name` subfolder
3. Try a `jupyter lite build`

## Expected behavior

In older versions of JupyterLite, I was able to build this project fine.

## Context

- JupyterLite version: 0.1.0
- Operating System and version: macOS and GitHub Actions
- Browser and version: N/A

<details><summary>Browser Output</summary>
<pre>
N/A
</pre>
</details>

jtpio:
Thanks @theonewolf for reporting :+1: 

Looks like this is indeed an issue. Which might be related to https://github.com/jupyterlite/jupyterlite/issues/984.

Issue #984, opened 2023-02-17
post_build:contents:contents fails if there is a folder with multibyte characters


katotetsuro:
## Description

post_build:contents:contents fails if there is a folder with multibyte characters in the name

## Reproduce

Build jupyterlite as follows

```
mkdir -p contents/multibyte_character_folder_あ 
touch contents/multibyte_character_folder_あ/foo.txt
jupyter lite build --contents contents
``` 

outputs the following error

```
.  post_build:contents:contents:multibyte_character_folder_あ
Couldn't fetch //multibyte_character_folder_%E3%81%82 as Jupyter contents.  HTTP 404: Not Found (file or directory does not exist: 'multibyte_character_folder_%E3%81%82')
                If this folder, or one of its parents, starts with a `.`, you can
                enable indexing hidden files with a `jupyter_lite_config.json` such as:

                    "ContentsManager": {
                        "allow_hidden": true
                    }

                Alternately, to skip it:

                    "LiteBuildConfig": {
                        "extra_ignore_contents": [
                            "/\.<the offendings path name>"
                        ]
                    }
                
TaskFailed - taskid:post_build:contents:contents:multibyte_character_folder_あ
Python Task failed: '<bound method ContentsAddon.one_contents_path of <jupyterlite.addons.contents.ContentsAddon object at 0x105ab6b80>>' returned False
```

## Expected behavior

The post_build process completes normally.

## Context

I think this problem has occurred since the following commit to use `as_uri`

https://github.com/jupyterlite/jupyterlite/commit/a9372f178accf2aa4b0523de32256a4e6e19e23b

I don't know what kind of problem exists in a Windows environment (avoided the conversion of \ to / in as_posix?), but if you want to convert to a path relative to self.output_files_dir, it may be replaced by the following code

```
listing_path = str(output_file_dir.relative_to(self.output_files_dir))
```

- JupyterLite version: 0.1.0b18
- Operating System and version: MacOS 12.6
- Browser and version: N/A

katotetsuro:
This appears to be the same problem
https://github.com/jupyterlite/jupyterlite/issues/1069

theonewolf:
@katotetsuro I agree, it is related to URL encoding the folder names, but not decoding them for file system ops.

PR #1116, opened 2023-07-31
stop using as_uri() for listing paths

katotetsuro:
## References

#984 #1069

## Code changes

- Avoid encoding with `as_uri` and use `relative_path`
- add unit test

## User-facing changes

none

none


jtpio:
Thanks @katotetsuro :+1: 

Looks like it will need a lint pass according to the CI failure.

jtpio (approved):
Thanks!

Issue #559, opened 2022-03-16
mock-socket WebSocketServer should use the mock: false flag to not stub out the global WebSocket 


shibbas:
### Problem

The latest version of mock-socket now provides a way to specify that the global WebSocket should not be stubbed out. Currently any following Websocket operations after starting a kernel also end up using mock-socket unless you set it back. It's best to set this option to false to avoid this. 

### Proposed Solution

Solution provided here: https://github.com/thoov/mock-socket/issues/308

jtpio:
Thanks @shibbas for the heads up :+1: 

Looks like the relevant code is here: 

https://github.com/jupyterlite/jupyterlite/blob/fa985055885cb1d53cccbf6e46ced7d14d38ea06/packages/kernel/src/kernels.ts#L152

Would you like to submit a PR to add the new flag? Thanks!

oeway:
Would be really useful if we can fix this, the current mock-socket used in the jupyterlite kernel is problematic when I try to use it with the native websocket.

PR #1118, opened 2023-08-02
Do not stub out the native WebSocket

oeway:
## References
Fixes https://github.com/jupyterlite/jupyterlite/issues/559
## Code changes

add `{ mock: false }` when starting the websocket server as suggested here: https://github.com/thoov/mock-socket/issues/308

## User-facing changes

None

None

~~Note that I haven't tested it locally, see if we pass the CI first.~~
I tested with the ReadTheDocs link, and the native websocket is preserved, the jupyterlite seems normal.


jtpio (approved):
Thanks!

PR #1019, opened 2023-03-22
Update to JupyterLab 4 and Notebook 7

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/826
Fixes https://github.com/jupyterlite/jupyterlite/issues/925
Fixes https://github.com/jupyterlite/jupyterlite/issues/262
Closes https://github.com/jupyterlite/jupyterlite/pull/457

## Code changes

- [x] Update `@jupyterlab` and `@lumino` packages
- [x] Remove a few extensions in the demo environment that are not compatible with JupyterLab 4 yet
- [x] Adjust message passing based on https://github.com/jupyterlab/jupyterlab/pull/11841
- [x] Update to Yarn 3: use `jlpm` for development
- [x] Update to `@jupyter-notebook` (Notebook 7) packages
- [x] Package Notebook 7 pre-releases on conda: https://github.com/conda-forge/notebook-feedstock/issues/109
- [x] Reuse upstream Playwright config, fix TODO added in https://github.com/jupyterlite/jupyterlite/pull/923
- [x] Fix Galata activation: https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#testing-with-galata
- [x] Closes https://github.com/jupyterlite/jupyterlite/pull/457
- [x] Check https://github.com/jupyterlite/jupyterlite/issues/509
- [x] Add `@jupyterlab/lsp-extension`
- [x] Add `@jupyterlab/metadataform-extension`
- [x] Disable `@jupyterlab-apputils-extension:announcements`
- [x] Mock `ws://localhost:8000/api/events/subscribe` or disable events for now?
- [x] Mock `lsp/status` endpoint
- [x] Notebook 7:
  - [x] Follow same plugin templating as in https://github.com/jupyter/notebook
  - [x] Build multiple apps for each endpoint? 
    - [x] `edit`
    - [x] `consoles`
    - [x] `notebooks`
    - [x] `tree`
- [x] Fix TODO and use a custom document opener plugin:

https://github.com/jupyterlite/jupyterlite/blob/fa5f44ff3f0eec2c75cada62fa659f13c9f26ab0/packages/retro-application-extension/src/index.ts#L46-L77  

- [x] Fix UI tests:
  - [x] Looks like some failures are related to https://github.com/jupyterlab/jupyterlab/pull/14945. Updating to the latest `4.0.5` packages fixes it
  - [x] Update Firefox snapshots
  - [x] Update Chromium snapshots
  - [x] Fix some tests
[...]

Issue #1008, opened 2023-03-15
Add a lower bound for the `jupyterlite-core` dependency in `jupyterlite`


jtpio:
Currently `jupyterlite` declares its dependency on `jupyterlite-core` here: 

https://github.com/jupyterlite/jupyterlite/blob/8a3bce63ec8e5bd97d132d60012c8a0378d11698/py/jupyterlite/pyproject.toml#L11

Since releases of `jupyterlite` and `jupyterlite-core` are done at the same time with the releaser, we might want to ensure a lower bound for `jupyterlite-core`.

For example `jupyterlite==0.1.0` would depend on `jupyterlite-core>=0.1.0`.

The `bump-version` script would need to be updated to keep the version in sync: https://github.com/jupyterlite/jupyterlite/blob/main/scripts/bump-version.py

Issue #1003, opened 2023-03-14
List of packages to include with the `jupyterlite` metapackage


jtpio:
### Problem

With https://github.com/jupyterlite/jupyterlite/pull/994 the `jupyterlite` package will be used as a metapackage. Which means it can include additional packages by default.

### Proposed Solution

Come up with a list of packages that should be included by default when installing `jupyterlite`.

With https://github.com/jupyterlite/jupyterlite/pull/1001 it is for now:

- `jupyterlite-core`
- `jupyterlite-pyodide-kernel`

### Additional context

Follow-up to https://github.com/jupyterlite/jupyterlite/pull/994 and https://github.com/jupyterlite/jupyterlite/issues/993

Also discussed in https://github.com/jupyterlite/jupyterlite/pull/1001#pullrequestreview-1337895709

In #1001 @bollwyvl writes:

> My thinking here: the "happy path" of one of the most complex dependency issues, relying on the users' package manager of choice to make `sys.prefix` whole, will start getting less happy as extensions don't install/solve correctly because of jupyterlab 4... and `[extras]` are basically broken.
> 
> A future `jupyterlab-core` can then depend on `jupyterlab >=4,<4.1`, even when we maintain the `3.5` compatible one.
> 
> I like widgets for this same reason: we can support `ipywidgets >=7.6,<8`, but there's no guarantee all future versions will have compatible `jupyterlab_widgets` releases, and getting kernels up to supporting widgets represents a _huge_ amount of effort (right next to arbitrary python packages, and iteratively reducing friction there).

PR #1142, opened 2023-09-13
[0.1.x] Add bounds to the `jupyterlite-core` dependency

jtpio:
## References

Follow-up to #1019 

In line of https://github.com/jupyterlite/jupyterlite/issues/1008

## Code changes

Add bounds to the `jupyterlite-core` dependency. 

## User-facing changes

Should help make upgrade scenarios smoother for those using the `jupyterlite` package instead of `jupyterlite-core`.

None

### DIFF
--- a/py/jupyterlite/pyproject.toml
+++ b/py/jupyterlite/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
 dependencies = [
     "jupyterlite-core >=0.1.2,<0.2.0",
     "jupyterlite-pyodide-kernel >=0.1.1",
-    "jupyterlite-javascript-kernel",
+    "jupyterlite-javascript-kernel >=0.1.1",
 ]
 keywords = [
     "browser",
--- a/py/jupyterlite/pyproject.toml
+++ b/py/jupyterlite/pyproject.toml
@@ -9,7 +9,7 @@ authors = [
 ]
 dependencies = [
     "jupyterlite-core >=0.1.2,<0.2.0",
-    "jupyterlite-pyodide-kernel >=0.0.4",
+    "jupyterlite-pyodide-kernel >=0.1.1",
     "jupyterlite-javascript-kernel",
 ]
 keywords = [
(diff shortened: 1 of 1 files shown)


py/jupyterlite/pyproject.toml
```diff
@@ -8,7 +8,7 @@ authors = [
     {name = "JupyterLite Contributors"},
 ]
 dependencies = [
-    "jupyterlite-core",
+    "jupyterlite-core >=0.1.2,<0.2.0",
     "jupyterlite-pyodide-kernel >=0.0.4",
```
bollwyvl, line 12: I think as we also happen to know this kernel would need to be updated to `>=0.1.1` for `DriveFS`, probably worth doing here. I don't think the JS kernel has a similar requirement, so is probably fine.

jtpio: Wondering if it's still actually relevant to include the Pyodide and JS kernels by default. Initially they were included to avoid disrupting workflows too much, but now that it's been a couple of months and there is a migration guide, we can also consider removing them.

The `jupyterlite` metapackage is still useful if we want to include other packages like `ipywidgets` for example, as discussed in https://github.com/jupyterlite/jupyterlite/issues/1003.

bollwyvl: Sure, `0.2.0` is a good opportunity to do this, as it would break the downstream dependency cycle. This would allow, for example, merging the `jupyterlite` and `jupyterlite-core` feedstocks on `conda-forge`... no harm in the JS kernel getting shipped from there.

bollwyvl: > ipywidgets for example

Well, strictly speaking, it would just be `jupyterlab_widgets`, no? Probably really only interested in the labextension side of the house, weird as that is.

Of course, the Real Deal would be to finally do the labextension (or widget) that visually built sites. Still a few missing pieces to self-hosting this in a browser-side kernel, but as a bootstrapping environment, wouldn't be awful.

jtpio: > Well, strictly speaking, it would just be `jupyterlab_widgets`, no?

Yes I meant that package :+1:

Issue #1138, opened 2023-09-11
Undo `"trailingComma": "es5"` prettier option


jtpio:
https://github.com/jupyterlite/jupyterlite/pull/1019 added that setting to make the diff simpler to read.

As a follow-up to https://github.com/jupyterlite/jupyterlite/pull/1019 we should remove that option to stick to `prettier` defaults.

PR #1143, opened 2023-09-13
Move prettier config to `package.json`, drop `trailingComma`

jtpio:
## References

Actually we could leave the `trailingComma` setting set to `es5` for now and see how it goes. And remove it later if we want to.

Closes https://github.com/jupyterlite/jupyterlite/issues/1138

## Code changes

Move the prettier config to `package.json`

## User-facing changes

None

None

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,0 @@
-{
-  "singleQuote": true,
-  "printWidth": 88,
-  "proseWrap": "always",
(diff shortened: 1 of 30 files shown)


bollwyvl:
At the end of this juggling, I think we _do_ want the trailing comma, as it leads to cleaner diffs, generally.

jtpio:
Fine either way.

PR #1159, opened 2023-09-20
Allow config worker url in ServiceWorkerManager

### LINKED ISSUE #1158: service worker named with hash makes it hard to use in downstream.
## Description

The service worker file is named at build-time by Webpack and it contains the hash of the file content. For our use case of Voici, we still use the worker provided by Jupyterlite but the worker file name is still computed by the Voici build process. This leads to a mismatch between the file name provided by Webpack and the file provided by Jupyterlite.

## Expected behavior

Can we remove the hash from the worker file name?

## Context

- JupyterLite version:
- Operating System and version:
- Browser and version:

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

trungleduc:
## References

Closes #1158 

## Code changes

Allow users to provide service worker URL in `ServiceWorkerManager` constructor

## User-facing changes

N/A

The service worker file name has been changed. In `0.1.0`, it was `service-worker-[hash].js` with the `hash` computed by webpack, in `0.2.0` the hash is removed and the new file name is `service-worker.js`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,7 +127,6 @@ app/test/data
 # do not include kernels such as Pyodide
 app/kernels
 app/extensions
-app/service-worker-*.js
 app/service-worker.js
 
 # generated html
--- a/.gitignore
+++ b/.gitignore
@@ -128,6 +128,7 @@ app/test/data
 app/kernels
 app/extensions
 app/service-worker-*.js
+app/service-worker.js
 
 # generated html
 app/index.app.template.html
--- a/packages/server/src/service-manager.ts
+++ b/packages/server/src/service-manager.ts
@@ -6,8 +6,10 @@ import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 import { IServiceWorkerManager, WORKER_NAME } from './tokens';
 
 export class ServiceWorkerManager implements IServiceWorkerManager {
-  constructor() {
-    void this.initialize().catch(console.warn);
+  constructor(options?: IServiceWorkerManager.IOptions) {
+    const workerUrl =
+      options?.workerUrl ?? URLExt.join(PageConfig.getBaseUrl(), WORKER_NAME);
+    void this.initialize(workerUrl).catch(console.warn);
   }
 
   /**
@@ -31,9 +33,9 @@ export class ServiceWorkerManager implements IServiceWorkerManager {
     return this._ready.promise;
   }
 
-  private async initialize(): Promise<void> {
+  private async initialize(workerUrl: string): Promise<void> {
     const { serviceWorker } = navigator;
-    const workerUrl = URLExt.join(PageConfig.getBaseUrl(), WORKER_NAME);
+
     let registration: ServiceWorkerRegistration | null = null;
 
     if (!serviceWorker) {
--- a/packages/server/src/tokens.ts
+++ b/packages/server/src/tokens.ts
@@ -2,6 +2,8 @@ import { Token } from '@lumino/coreutils';
 
 import { ISignal } from '@lumino/signaling';
 
+import SW_URL from './service-worker?text';
+
 /**
  * The token for the ServiceWorker.
  */
@@ -47,4 +49,4 @@ export namespace IServiceWorkerManager {
   }
 }
 
-export const WORKER_NAME = 'service-worker.js';
+export const WORKER_NAME = `${SW_URL}`.split('/').slice(-1)[0];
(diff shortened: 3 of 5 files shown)


martinRenou:
I *think* that does not fix the case for jupyterlite-xeus-python, or does it?

martinRenou:
I understand it only fixes the Voici case, as Voici would be able to provide its own Service Worker with a specific name.

But in the case of jupyterlite-xeus-python, the issue is that it builds against a specific built version of jupyterlite, with a specific hash name for the service worker. And whenever a jupyterlite release comes in, jupyterlite-xeus-python now has an invalid information about the service worker name. Until we build again jupyterlite-xeus-python.

So I think it would be better to remove the hash entirely.

martinRenou:
I don't recall the details exactly, maybe @jtpio does?

trungleduc:
I removed the content hash, but I think there is no harm to allow configuring the worker URL, so I keep the initial changes.

jtpio:
> Out of curiosity, `"@jupyterlite/server` is marked with `"bundled": false` in `jupyterlite-xeus-python`, I don't understand why there still is a worker file in the static assets of this extension.

hmm good question. Maybe digging in the git history will help.

packages/server/src/tokens.ts
```diff
@@ -34,4 +34,16 @@ export interface IServiceWorkerManager {
...
 
+/**
+ * A namespace for `ServiceWorkerManager` class.
+ */
+export namespace IServiceWorkerManager {
+  /**
+   * An options object for initializing a worker manager.
+   */
+  export interface IOptions {
+    workerUrl?: string;
```
jtpio, line 45: Maybe we can add a docstring for `workerUrl` for consistency?

jtpio:
It's a bit flaky sometimes. Seems to be passing after restarting it.

trungleduc:
Thanks!

.gitignore
```diff
@@ -128,6 +128,7 @@ app/test/data
 app/kernels
 app/extensions
 app/service-worker-*.js
```
martinRenou, line 130: Do we still need this rule?

martinRenou (commented):
Just a minor comment, otherwise 🚀 !

.gitignore
trungleduc: Nope

jtpio:
Also maybe we can log this change in the migration guide under a new `Service Worker` section, since this affects downstreams: https://jupyterlite.readthedocs.io/en/latest/migration.html#to-0-2-0

jtpio (approved):
Thanks!

Following up with a release so you can test it downstream @trungleduc @martinRenou.

trungleduc:
> Thanks!
> 
> Following up with a release so you can test it downstream @trungleduc @martinRenou.

Great! Thank you very much.

packages/server/src/service-manager.ts
```diff
@@ -6,8 +6,10 @@ import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 import { IServiceWorkerManager, WORKER_NAME } from './tokens';
 
 export class ServiceWorkerManager implements IServiceWorkerManager {
-  constructor() {
-    void this.initialize().catch(console.warn);
+  constructor(options?: IServiceWorkerManager.IOptions) {
+    const workerUrl =
+      options?.workerUrl ?? URLExt.join(PageConfig.getBaseUrl(), WORKER_NAME);
```
stevejpurves, line 11: @trungleduc @jtpio is there a way to set this option? from what I can see here: https://github.com/jupyterlite/jupyterlite/blob/bde12e5209c37e278c5b84af9ebe40fc17696475/packages/server/src/app.ts#L65-L66

there's no way to pass this in when creating a new instance of the JupyterLiteServer, and no way to load options onto the page

stevejpurves:
Thanks for these changes @trungleduc @martinRenou @jtpio 🙏🏼 the fact that the service worker filename is stable alone, means we can override this in thebe at build time. having control over the workUrl would be nice though -- but I don't think there's a way to do that yet. is it appropriate for that to be an option directly on the `JupyterLiteServer` constructor?

packages/server/src/service-manager.ts
trungleduc: you can disable the `"@jupyterlite/server-extension:service-worker"` plugin and provide your own `IServiceWorkerManager`
https://github.com/jupyterlite/jupyterlite/blob/bde12e5209c37e278c5b84af9ebe40fc17696475/packages/server-extension/src/index.ts#L200-L207

stevejpurves: cool - I take it that you mean adding my own plugin with this `id: '@jupyterlite/server-extension:service-worker'` will automatically override / disable the default plugin, and of course replace it with the new one?

PR #1176, opened 2023-09-28
Add missing singleton packages, fix extra newlines, fix duplicated content

jtpio:
## References

More follow-ups to #1019 

https://github.com/jupyterlite/jupyterlite/issues/1141

## Code changes

- [x] Add missing `singletonPackages` (align with JupyterLab)
- [x] Fix duplicated content on edit
- [x] Bump to a newer `@jupyter/doc` that would include https://github.com/jupyter-server/jupyter_ydoc/pull/186 to fix the extra newlines issue.
  - Using `jlpm up @jupyter/ydoc -R` 

## User-facing changes

Might help fix some JupyterLab 4 extensions compatibility issues.

Also fixes duplicated file on save as reported in https://github.com/jupyterlite/jupyterlite/issues/1141

https://github.com/jupyterlite/jupyterlite/assets/591645/0c6e2441-b0c5-4d3a-8521-3183cc25722b

## Backwards-incompatible changes

None

Issue #1141, opened 2023-09-13
JupyterLab 4 / Notebook 7 follow-ups


jtpio:
Tracking follow-ups of https://github.com/jupyterlite/jupyterlite/pull/1019 so they can be fixed before a `0.2.0` final.

Listing them in a single issue, but can be split into multiple if needed.

## Service Worker

Check if the following behavior is still happening:

- [x] ~after opening a bunch of notebooks, the service worker basically dies~
    - `Uncaught (in promise) DOMException: Failed to execute 'put' on 'Cache': Quota exceeded.`
    - at that point, you can't open/save anything anymore, even after reloading the page
      - requires a new (private) browser

-> can be deferred to a separate issue if needed: https://github.com/jupyterlite/jupyterlite/issues/1141#issuecomment-1743072720

## Demo site

- [x] Fix app redirect issues on RTD (and elsewhere?): https://github.com/jupyterlite/jupyterlite/pull/1019#issuecomment-1711733062
  - Might not be an issue anymore? https://github.com/jupyterlite/jupyterlite/pull/1170#issuecomment-1738642971 
- [x] all notebooks are rendering with extra newlines: https://github.com/jupyterlite/jupyterlite/pull/1176
  - this actually breaks notebooks like ipycanvas, where there are long multiline strings
- [x] sometimes saving (maybe just opening?) notebooks creates duplicates in the file browser: https://github.com/jupyterlite/jupyterlite/pull/1178
- [x] altair notebook throws a bunch of deprecation warnings: https://github.com/jupyterlite/jupyterlite/pull/1181
- [x] the `ipyvue` widgets don't have matching kernel/extension packages and appear about 50% broken
  -  but still cost a few mb of downloads on _page load_, plus the 11mb of donwload at kernel start
- [x] leaflet can't find mathjax
  -  this appears to fire pretty early, so breaks more than just mathjax
  - Investigating in https://github.com/jupyterlite/jupyterlite/pull/1184

## Console errors

- [x] mathjax missing fonts for rendered outputs (js notebook)

Fixed in https://github.com/jupyterlab/jupyterlab/pull/15230

```
_static/lab/js/output/chtml/fonts/tex-woff-v2/MathJax_Zero.woff
```

- [ ] cell toolbar appears to be throwing _more_ errors than before

```
celltoolbartracker.js:239 Uncaught TypeError: Cannot read properties of undefined (reading 'node'
```

- toc is throwing errors

```
[...]

PR #1178, opened 2023-09-29
Fix file indexing that caused content duplication

jtpio:
## References

Partial backport of https://github.com/jupyterlite/jupyterlite/pull/1176 to `0.1.x`.

See https://github.com/jupyterlite/jupyterlite/issues/1141#issuecomment-1739346299 for more information.

## Code changes

Normalize the listing path to avoid creating an index with `./` as prefix for top-level `path` fields.

## User-facing changes

This should fix duplicated files issues on save as reported in https://github.com/jupyterlite/jupyterlite/issues/1141

None


jtpio:
Will make a `0.1.3` release right after.

bollwyvl:
Good find!

bollwyvl (approved):
Rekicked win38, but LGTM

PR #1182, opened 2023-09-29
Bump `ipyvue` and `ipyvuetify`

jtpio:
## References

Investigate if this would fix the issue reported in https://github.com/jupyterlite/jupyterlite/issues/1141

## Code changes

Bump `ipyvue` and `ipyvuetify` dependencies in the examples.

None


bollwyvl:
Appears to not be quite as janky, though still:
- loads 3mb of assets at application launch time
- `TypeError: this.hasBanana is not a function`
- makes a lot of icky calls to external domains

jtpio:
We can also remove them from the examples if it's too much trouble maintaining them.

bollwyvl:
Nah, I think it's fine if we can keep them running with minor updates now and again. Anything that stresses some of the known regressable features (e.g. `bqplot`'s use of nested buffers) are worth having around, and us finding rather than downstream users, especially during major upgrade cycles.

bollwyvl:
also #1183 should help with the windows flake

jtpio:
RTD preview for reference: 

image

Then let's keep them for now in the examples.

PR #1193, opened 2023-10-05
Cleanup Mathjax

### LINKED ISSUE #1192: [0.2.0] Update for removed out-of-band MathJax assets
### Problem

JupyterLab 4 no longer uses `mathjaxUrl`, and pre-bundles the font assets. We still have a fair amount of code hanging around for it, but it won't do anything useful.

### Proposed Solution

- [ ] determine which set of fonts (e.g. in `js/output/chtml/` or `build/`) is _correct_ (see below)
  - > pre-compressed `.woff2` would be preferable, but not supported upstream
- [ ] update code paths to only use the correct font URLs and avoid `404`
- [ ] remove the MathJax addon, CLI flags, etc.
- [ ] remove the `[mathjax]` extra
- [ ] remove `jupyter-server-mathjax` from any development deps
- [ ] remove `mathJaxUrl` from the schemas
- [ ] update the docs
  - [ ] remove most references to mathjax
  - [ ] add to migration guide
    - > I guess if you want a custom mathjax build now, you need a custom labextension)
- [ ] update ui tests
  - [ ] if we do end up changing how anything works in a meaningful way from upstream

### Additional context

While MathJax _works_, we're apparently not handling startup quite right:

| lite | lab `main` |
|--|--|
| image | image

The lab ones pull from `js/output/chtml/`, but this is... pretty weird:

```
http://127.0.0.1:8889/lab/tree/js/output/chtml/fonts/tex-woff-v2/MathJax_Zero.woff
```

As MathJax is... large, we very likely shouldn't keep around a copy for every app.

jtpio:
## References

Fixes #1192.

Follow-up to #1019

Upstream related PR: https://github.com/jupyterlab/jupyterlab/pull/13877

## Code changes

Items from #1192 :

- [x] Remove the MathJax addon, CLI flags, etc.
- [x] Remove the `[mathjax]` extra
- [x] Remove `jupyter-server-mathjax` from any development deps
- [x] Remove `mathJaxUrl` from the schemas
- [x] Remove most references to mathjax from the documentation
- [x] Add to migration guide
- [x] ~Update code paths to only use the correct font URLs and avoid `404`~ -> https://github.com/jupyterlab/jupyterlab/pull/15230
- [x] ~determine which set of fonts (e.g. in `js/output/chtml/` or `build/`) is _correct_ (see below)~ -> https://github.com/jupyterlab/jupyterlab/pull/15230
  - pre-compressed `.woff2` would be preferable, but not supported upstream
- [x] ~UI test to check there is no Mathjax related 404s?~  -> might consider adding one after https://github.com/jupyterlab/jupyterlab/pull/15230

## User-facing changes

None

Removed from `jupyter-lite.json`:

- `mathjaxConfig`
- `fullMathjaxUrl`

Removed from `jupyterlite-core`:

- `mathjax` addon
- `mathjax` optional dependencies


app/jupyterlite.schema.v0.json
```diff
@@ -86,12 +86,6 @@
           "default": "./extensions",
           "$ref": "#/definitions/non-trailing-slash-uri"
         },
-        "fullMathjaxUrl": {
```
bollwyvl, line 89: i guess removing this doesn't _strictly_ require a version number bump, as `additionalProperties: true`....

docs/migration.md
```diff
@@ -57,6 +57,12 @@ removed and the new file name is `service-worker.js`.
 
 ### API changes
 
+#### `jupyterlite-core`
+
+- The Mathjax addon was removed from the `jupyterlite-core` package. As a consequence
```
bollwyvl, line 62: Could reference the upstream change, and maybe offer https://pypi.org/project/jupyterlab-mathjax2/ if someone wants the older behavior. Not sure if it can be configured...

py/jupyterlite-core/pyproject.toml
```diff
@@ -124,9 +119,3 @@ artifacts = [
 include = [
     "/jupyterlite_core",
 ]
-
-[tool.pytest.ini_options]
-filterwarnings = [
-    # https://github.com/jupyter-server/jupyter_server_mathjax/pull/11
-    "ignore:.*Use `_i18n\\(\\)`.*:FutureWarning",
```
bollwyvl, line 131: :tada:

jtpio:
https://github.com/jupyterlab/jupyterlab/pull/14726 added some logic to ignore some of the Mathjax 404s in the examples. So there might be something more involved in JupyterLab.

docs/migration.md
bollwyvl: Also probably `MathJax`.

> because the RTD demo uses `jupyterlab-myst`

Given it breaks a fair amount of CSS, still, maybe it's not worth keeping that extension it if it hides (problems in) core, out-of-the-box features.

bollwyvl:
Yep: it looks like upstream might need to overload/intercept `CHTML.stylesheet` where it's created in the upstream extension: https://github.com/jupyterlab/jupyterlab/blob/v4.0.6/packages/mathjax-extension/src/index.ts#L70

image

bollwyvl:
Started https://github.com/jupyterlab/jupyterlab/pull/15230 ... presumably could be backported if there's going to be a 4.0.7...

docs/migration.md
jtpio: Maybe. Although in this particular case the ignored 404s in https://github.com/jupyterlab/jupyterlab/pull/14726 should probably have been tracked in a separate issue (would have been easier to find the issue at least).

jtpio:
Opened https://github.com/jupyterlite/jupyterlite/issues/1195 to keep track of it,

jtpio:
Will release `0.2.0b1` with this change.

Issue #1211, opened 2023-10-13
Opening a notebook from a URL with an encoded space character in it is broken


tomjakubowski:
## Description

I wanted to try my friend's ipynb in Jupyterlite. It was published on GitHub, so I added it from the File -> Open from URL menu.

When I did that, a file was added to my local Jupyterlite filesystem. But I wasn't able to open it - instead, I got this error message.

<img width="468" alt="Screenshot 2023-10-13 at 12 46 42 PM" src="https://github.com/jupyterlite/jupyterlite/assets/667901/d2862a86-e797-4e78-944d-1b8aead555a7">

## Reproduce

1. Go to https://jupyterlite.readthedocs.io/en/stable/_static/lab/index.html
2. File -> Open from URL
3. Open my friend's notebook https://raw.githubusercontent.com/mds2/rocketry/main/rocket%20equations.ipynb 

## Expected behavior

The notebook is added and I can open it.

## Context

- JupyterLite version: Version 0.1.3
- Operating System and version: macOS 13.3.1
- Browser and version: Chrome 117

<details><summary>Browser Output</summary>
<pre>
dialog.tsx:43 Showing error: Error: Could not find file with path rocket%20equations.ipynb
    at m.createCheckpoint (contents.ts:513:13)
    at async index.ts:120:15
    at async _.createCheckpoint (index.ts:1308:11)
D @ dialog.tsx:43
_handleError @ context.ts:769
(anonymous) @ context.ts:714
Promise.catch (async)
_revert @ context.ts:711
initialize @ context.ts:269
await in initialize (async)
(anonymous) @ manager.ts:590
Promise.then (async)
_createOrOpenDocument @ manager.ts:590
open @ manager.ts:365
openOrReveal @ manager.ts:401
handleOpen @ listing.ts:1021
evtDblClick @ listing.ts:1116
handleEvent @ listing.ts:666
manager.ts:605 Failed to initialize the context with 'notebook' for rocket%20equations.ipynb Error: Could not find file with path rocket%20equations.ipynb
    at m.createCheckpoint (contents.ts:513:13)
    at async index.ts:120:15
    at async _.createCheckpoint (index.ts:1308:11)
(anonymous) @ manager.ts:605
Promise.catch (async)
_createOrOpenDocument @ manager.ts:604
open @ manager.ts:365
openOrReveal @ manager.ts:401
handleOpen @ listing.ts:1021
evtDblClick @ listing.ts:1116
[...]

PR #1216, opened 2023-10-18
[WIP] Contents: avoid double-URIdecoding paths

tomjakubowski:
Hello! We're working on a jupyterlite integration over at https://prospective.co. I'm thrilled to contribute a bug fix back to jupyterlite! 🎉 

## References

Addresses https://github.com/jupyterlite/jupyterlite/issues/1211

## Code changes

The paths coming in to /api/contents need to be URI decoded, because jupyterlab's frontend URI encodes them.

The existing Contents API would sometimes double-decode the filenames. For example, `Contents.createCheckpoint()` would decode the path once, then call `this.get()` on the path, which would decode it again.

To prevent double-decoding, the URI decode is moved out of the Contents class and into the route definitions.

## User-facing changes

Users can now add both files from URLs containing percent-encoded characters, and files from their filesystem with names like `hello%20world.txt`.

This has a side effect of changing the Contents class API to accept "regular" paths, and not URI-encoded ones. For example, `Contents.get("hello%20world.txt")` will now return a file named `hello%20world.txt` on the filesystem, instead of a file named `hello world.txt`.

### DIFF
--- a/ui-tests/test/contents.test.ts
+++ b/ui-tests/test/contents.test.ts
@@ -111,20 +111,26 @@ test.describe('Contents Tests', () => {
     expect(output2![0]).toBe('4');
   });
 
-  test('Open a URL with a percent-encoded filename', async ({ page }) => {
-    await page.menu.clickMenuItem('File>Open from URL…');
-    // This test relies on JupyterLab's behavior of using the response as the file contents when it is a 404.
-    // We can't mock a response with page.route() because the requests come from JupyterLite's service worker
-    const name = 'test file.txt';
-    const url = encodeURIComponent(name);
-    await page.locator('#jp-dialog-input-id').fill(url);
-    await page.locator('.jp-mod-accept').click();
+  // These tests rely on JupyterLab's behavior of using a 404's response body as the file contents.  In this
+  // case the 404 is from the test server.
+  // We can't mock a response with page.route() because Playwright doesn't support intercepting the network
+  // for a service worker
+  const names = ['test file.txt', 'test%20file.txt'];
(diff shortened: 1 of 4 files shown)


tomjakubowski:
My coworker and I started on a ui-test for this, a new test which ran through "Open from URL". 

Now that the fix is ready, the bug seems larger in scope than that ui test would suggest. Some of the interactions are also a bit clunky to test with Playwright. Would it be acceptable to instead write a unit/regression test for this in the `contents` package?

tomjakubowski:
still some weirdness on this branch! so very much a wip, which I have now marked

example of some remaining weirdness:

1. Create a folder on your filesystem with files called `hello%20world.txt` and `hello world.txt`. See [foo.zip](https://github.com/jupyterlite/jupyterlite/files/12972870/foo.zip)
2. Drag and drop `hello world.txt` into Jupyterlite.
3. Drag and drop `hello%20world.txt` into Jupyterlite. Instead of adding a new file, it overwrites `hello world.txt` and renames it to `hello%20world.txt`

ui-tests/test/contents.test.ts
```diff
@@ -22,7 +22,7 @@ test.use({
   waitForApplication: firefoxWaitForApplication,
 });
 
-test.describe('Contents Tests', () => {
+test.describe.only('Contents Tests', () => {
```
jtpio, line 25: You can also use the `-g` Playwright command line option to only run a subset of the test suite.

tomjakubowski: thanks! nice tip

Issue #1213, opened 2023-10-16
Issue when loading settings from federated extensions


jtpio:
## Description

Currently the JupyterLab app is missing a menu entry to open Notebook 7 in a new browser tab.

## Reproduce

On https://jupyterlite.readthedocs.io/en/latest/_static/lab/index.html :

![image](https://github.com/jupyterlite/jupyterlite/assets/591645/639e7961-7d89-4cda-98a2-1a4957df54b8)

## Expected behavior

It should be possible to open Notebook 7 from JupyterLab like with JupyterLite 0.1.x using RetroLab:

![image](https://github.com/jupyterlite/jupyterlite/assets/591645/49350f8f-768b-4c0e-bb21-9f278c06b81b)

But using the existing Notebook 7 extension that adds the following menu entry:

![image](https://github.com/jupyterlite/jupyterlite/assets/591645/6180c087-6e65-4541-a388-29c4a5162a10)

## Context

Follow-up to #1019.

jtpio:
Installing the `notebook` package and running `jupyter lite build` correctly includes `@jupyter-notebook/lab-extension` in `jupyter-lite.json`:

```json
{
  "extension": "./extension",
  "liteExtension": false,
  "load": "static/remoteEntry.749fd42299b69250a475.js",
  "name": "@jupyter-notebook/lab-extension",
  "style": "./style"
}
```

The commands are available in the command palette, but the menu and toolbar items are missing: 

![image](https://github.com/jupyterlite/jupyterlite/assets/591645/b1b29288-371e-497c-96b9-112059e3524e)

So there might be a more general issue of loading settings from a federated extension (to be confirmed).

jtpio:
> So there might be a more general issue of loading settings from a federated extension (to be confirmed).

Looking more into it, there does seem to be a more general issues with toolbar items and menu entries (defined via the settings). This is for example reproducible with https://github.com/jupyterlab/extension-examples/tree/main/toolbar-button.

For now not sure if it's related to a upstream change in JupyterLab 4 w.r.t how the settings are loaded. Next step would be confirm that settings based items and menu entries used to work fine in JupyterLite 0.1.0 built on JupyterLab 3.

jtpio:
Maybe related to https://github.com/jupyterlab/jupyterlab/pull/14195 ?

jtpio:
JupyterLab 4 now seems to be expecting **all** settings to be returned in one call at application startup.

[...]

PR #1220, opened 2023-10-19
Load all settings from federated extensions at startup

jtpio:
## References

JupyterLab 4 seems to now be expecting all settings (including the ones from federated extensions) to be returned in one go on the first call to `/api/settings`.

This is likely related to the change in https://github.com/jupyterlab/jupyterlab/pull/14195 and/or other upstream PRs as well.

Fixes https://github.com/jupyterlite/jupyterlite/issues/1213

## Code changes

- [x] Aggregate settings from federated extensions in a `all_federated.json` file by adding a step in the `federated_extensions` addon `post_build`
- [x] UI test to check the "Launch Notebook File Browser" item is available when the `notebook` package is installed in the build environment
- [x] Handle redirects to `/doc/tree`
- [x] UI test to check going from Notebook to Lab with the toolbar button
- [x] Mention installing the `notebook` package in the docs to get the UI elements

An alternative could have been to create an `all.json` file with all the settings for a federated extension, in each federated extension directory. And query this file for each extension. However many extensions don't provide settings, so these queries would lead to many 404s.

## User-facing changes

End users should be able to see menu entries or toolbar items defined in third-party extensions via the settings system.

image

https://github.com/jupyterlite/jupyterlite/assets/591645/63e7cbad-e4eb-4df4-b6c4-8d92e2610812

None

### DIFF
--- a/packages/settings/src/settings.ts
+++ b/packages/settings/src/settings.ts
@@ -1,13 +1,14 @@
 import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 
+import { PromiseDelegate } from '@lumino/coreutils';
+
 import * as json5 from 'json5';
 
 import type localforage from 'localforage';
 
-import { IFederatedExtension } from '@jupyterlite/types';
+// import { IFederatedExtension } from '@jupyterlite/types';
 
 import { IPlugin, ISettings } from './tokens';
-import { PromiseDelegate } from '@lumino/coreutils';
 
 /**
  * The name of the local storage.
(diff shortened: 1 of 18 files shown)


packages/settings/src/settings.ts
```diff
@@ -133,37 +131,15 @@ export class Settings implements ISettings {
...
-    return Private.override({
-      id: pluginId,
-      raw,
-      schema,
-      settings,
-      version: packageJson.version || '3.0.8',
-    });
+  private async _getAll(type: 'core' | 'federated' = 'core'): Promise<IPlugin[]> {
+    const settingsUrl = PageConfig.getOption('settingsUrl') ?? '/';
+    const file = type === 'core' ? 'all.json' : 'all_federated.json';
```
bollwyvl, line 138: maybe the `type` could just be the file name... and hoisted to a named type.

py/jupyterlite-core/jupyterlite_core/addons/federated_extensions.py
```diff
@@ -284,6 +286,53 @@ def post_build(self, manager):
...
+
+    def get_federated_settings(self, extension):
+        """get the settings for a federated extension"""
+        pkg_json = extension / PACKAGE_JSON
+        pkg_data = json.loads(pkg_json.read_text(**UTF8))
+        settings_dir = extension / "schemas"
+        if not settings_dir.is_dir():
+            # bail if there is no settings for that extension
+            return []
+        setting_files = settings_dir.rglob("*.json")
```
bollwyvl, line 317: probably sort these for stable output

py/jupyterlite-core/jupyterlite_core/addons/federated_extensions.py
```diff
@@ -284,6 +286,53 @@ def post_build(self, manager):
...
+        yield self.task(
+            name="settings",
+            doc=f"ensure {ALL_FEDERATED_JSON} includes the settings of federated extensions",
+            file_dep=[*lab_extensions],
+            actions=[(self.ensure_federated_settings, [manager, lab_extensions])],
+        )
+
+    def ensure_federated_settings(self, manager, lab_extensions):
+        """ensure settings from federated extensions are aggregated in a single file"""
+        all_federated_settings = [
```
bollwyvl, line 298: probably sort these, too

jtpio:
Thanks @bollwyvl for the review, I pushed a couple more commits to address the suggestions :+1:

jtpio:
Getting this one in and following up with a new release.

Issue #942, opened 2023-01-08
For Notebook in Local Files, /drive points to JupyterLite Storage when Notebook is in the top level directory.


tanaga9:
### Problem

Unstable references from /drive in JupyterLite with jupyterlab-filesystem-access extension
https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/issues/48
![211135747-b3e4c502-cdd8-4128-aa11-9b8dd2b59481](https://user-images.githubusercontent.com/6501421/211285715-f51edae3-0400-4a4e-bd80-23c5a423fcc5.png)

- Sometimes /drive refers to JupyterLite Storage
- Sometimes /drive refers to Local Files

I don't know the conditions for either behavior. (when Notebook is in the top level directory ?)

I guessed the [Storage Settings](https://jupyterlite.readthedocs.io/en/latest/howto/configure/storage.html#settings-storage) in JupyterLite is involved.

Because it seemed that JupyterLite and jupyterlab-filesystem-access extension conflicted on /drive.
(Can you see what's going on?)

I tried setting `jupyter-lite.json#jupyter-config-data/contentsStorageDrivers` to `["FileSystem"]`.
([FileSystem is a Drive Name](https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/blob/043521aeb090ccb67aa9ca867082a7ef042a4467/src/drive.ts#L7))

```
{
    "jupyter-config-data": {
      "contentsStorageDrivers": ["FileSystem"]
    },
    "jupyter-lite-schema-version": 0
}
```

This attempt failed.

### ~~Suggested Improvement~~

~~Document how to stabilize /drive references with jupyterlab-filesystem-access extension.~~

### Expected

For local file notebooks, /drive always points to JupyterLite storage.

martinRenou:
The current behavior is that any Notebook started from the local file system extension would use this local directory as `/drive` and any Notebook started from the JupyterLite storage would use the JupyterLite storage.

The reasoning behind this is that notebooks from a local folder would sometimes load local data files (`.csv`, etc.) so they need access to those.

For now this is not configurable but we could make it configurable.

What kind of behavior would you like to see? Do you want to always have `/drive` point to the jupyterlite storage?

tanaga9:
Thanks a lot @martinRenou

> What kind of behavior would you like to see? Do you want to always have /drive point to the jupyterlite storage?

[...]

PR #1227, opened 2023-10-23
Fix handling of files in custom drives

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/942

As noticed in https://github.com/jupyterlab/jupyterlab/pull/14519#issuecomment-1774055091, JupyterLab 4 does not store the drive prefix (for example `FileSystem:`) in the session path anymore.

But JupyterLite is relying on this information to start a new kernel:

https://github.com/jupyterlite/jupyterlite/blob/0ce857b1aab8db3253d51d6a126e8cd6a0260d9d/packages/session/src/sessions.ts#L114-L118

Which is then used in the kernel to store the name of the drive, and pass it to the `DriveFS` used in the kernels:

**Pyodide**

https://github.com/jupyterlite/pyodide-kernel/blob/143526c9e11589b03c8d6844be5471d527b76255/packages/pyodide-kernel/src/worker.ts#L23-L30

https://github.com/jupyterlite/pyodide-kernel/blob/143526c9e11589b03c8d6844be5471d527b76255/packages/pyodide-kernel/src/worker.ts#L117-L124

**Xeus Python**

https://github.com/jupyter-xeus/xeus-lite/blob/07df211bc73c1002558017b86499d63806d133d6/share/xeus-lite/web_worker_kernel.ts#L164-L171

https://github.com/jupyter-xeus/xeus-lite/blob/07df211bc73c1002558017b86499d63806d133d6/share/xeus-lite/worker.ts#L119-L126

As a result, the update to JupyterLab 4 broke accessing files from a custom drive, as the `driveName` would always be empty since it would never be known to the kernel.

See the following discussions for more information:

- https://github.com/jupyterlab/jupyterlab/pull/14519#issuecomment-1774055091
- https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/pull/55#issuecomment-1772879905
- https://github.com/jupyterlab/jupyterlab/issues/15143
- https://github.com/jupyterlab/jupyterlab/issues/15289

## Code changes

To restore this functionality, we add a small plugin to patch `sessionContext['_path']` and `sessionContext['_name']` when a document context is created and opened.

This also allows fixing the behavior without having to introduce breaking changes (and updating kernels) in 0.2.0. However this is more of a workaround, and a better approach should be found in 0.3.0, which will require quite some work in JupyterLab directly to allow for such use cases.

- [x] Fix accessing contents of custom drives from kernels
- [x] Fix previous issue where trying to access content from a notebook created at the top-level of a custom drive would default to accessing the content from the default browser
- [x] ~Add `jupyterlab-filesystem-access` extension to the RTD demo site?~ -> postponing to a later time since there is already `jupyterlab-github`
  - Would need https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/pull/55, https://github.com/jupyterlab-contrib/jupyterlab-filesystem-access/issues/58 and a new release of the extension first
  - There is already `jupyterlab-github` that we can use to test by opening the `virtual_filesystem` notebook from within the GitHub drive
- [x] ~UI test with a simple custom drive?~ -> tracked in https://github.com/jupyterlite/jupyterlite/issues/1231

## User-facing changes

This should allow accessing files from a custom drive from a kernel in JupyterLite 0.2.0:

**jupyterlab-filesystem-access**

https://github.com/jupyterlite/jupyterlite/assets/591645/a0ab40d1-1a94-482b-8323-1b2e50c0528f

**jupyterlab-github**

image

None

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -453,6 +453,8 @@ const sessionContextPatch: JupyterFrontEndPlugin<void> = {
       // with a kernel.
       // This was changed in JupyterLab 4 in https://github.com/jupyterlab/jupyterlab/pull/14519
       // and is needed for the kernel to be aware of the drive it is associated with.
+      // This is a temporary fix until a better solution is found upstream in JupyterLab ideally.
+      // This also avoid having to patch the downstream kernels (e.g. xeus-python and pyodide)
       sessionContext['_name'] = context?.path;
       sessionContext['_path'] = context?.path;
     });
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -8,7 +8,13 @@ import {
   ILabShell,
 } from '@jupyterlab/application';
 
-import { Clipboard, ICommandPalette, Dialog, showDialog, SessionContext } from '@jupyterlab/apputils';
+import {
+  Clipboard,
+  ICommandPalette,
+  Dialog,
+  showDialog,
+  SessionContext,
+} from '@jupyterlab/apputils';
 
 import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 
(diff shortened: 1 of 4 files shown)


jtpio:
On RTD with the `jupyterlab-github` drive:

image

Issue #751, opened 2022-07-27
Serviceworker re-registers every time


joemarshall:
The servicewkrker registration should really only register if it isn't already registered. The reason being that in Chrome at least, repeatedly reregistering the service worker creates multiple registrations in the application tab and is generally not a good idea even if it works. 

I'd suggest combining this with moving the serviceworker into the server package, so one can use cache busting hashing in the url to make it work on updates.

jtpio:
The handling of the service worker was improved in the latest 0.6.0 release:

- https://github.com/jupyterlite/jupyterlite/pull/1627
- https://github.com/jupyterlite/jupyterlite/pull/1621

So closing this issue as likely fixed. But don't hesitate to re-open or open a new one if there are still issues with the service worker, thanks!

PR #1251, opened 2023-11-07
Fixed so hard refresh loads service worker properly

andeplane:
Thanks @eiriklv who helped me figuring this out.

## References
See https://github.com/jupyterlite/pyodide-kernel/issues/76 

## Code changes
Coming 

## User-facing changes
Nothing except fixing the bug.

### DIFF
--- a/packages/server/src/service-manager.ts
+++ b/packages/server/src/service-manager.ts
@@ -40,27 +40,30 @@ export class ServiceWorkerManager implements IServiceWorkerManager {
 
     if (!serviceWorker) {
       console.warn('ServiceWorkers not supported in this browser');
-    } else if (serviceWorker.controller) {
-      registration =
-        (await serviceWorker.getRegistration(serviceWorker.controller.scriptURL)) ||
-        null;
-      // eslint-disable-next-line no-console
-      console.info('JupyterLite ServiceWorker was already registered');
+      this._ready.reject(void 0); // Reject the promise if service workers are not supported.
+      return;
     }
 
-    if (!registration && serviceWorker) {
-      try {
-        // eslint-disable-next-line no-console
-        console.info('Registering new JupyterLite ServiceWorker', workerUrl);
-        registration = await serviceWorker.register(workerUrl);
-        // eslint-disable-next-line no-console
-        console.info('JupyterLite ServiceWorker was sucessfully registered');
-      } catch (err: any) {
-        console.warn(err);
-        console.warn(
-          `JupyterLite ServiceWorker registration unexpectedly failed: ${err}`,
-        );
-      }
+    // Unregister any existing service workers before registering the new one.
+    const registrations = await serviceWorker.getRegistrations();
+    for (const registration of registrations) {
+      await registration.unregister();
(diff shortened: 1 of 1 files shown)


andeplane:
@jtpio this PR is ready for review and seems to fix the issue both myself and @tgourdel have encountered. I can't run the linter locally for some reason and the CI doesn't tell me what's wrong so would love help fixing that :D

jtpio:
Thanks @andeplane and @eiriklv!

I'll have a look at the lint failure.

cc @martinRenou for awareness

packages/server/src/service-manager.ts
```diff
@@ -40,35 +40,38 @@ export class ServiceWorkerManager implements IServiceWorkerManager {
...
-        (await serviceWorker.getRegistration(serviceWorker.controller.scriptURL)) ||
-        null;
+      this._ready.reject('Not supported'); // Reject the promise if service workers are not supported.
+      return;
+    }
+
+    // Unregister any existing service workers before registering the new one.
+    const registrations = await serviceWorker.getRegistrations();
+    for (const registration of registrations) {
+      await registration.unregister();
```
jtpio, line 50: Curious, does that mean the service worker is unregistered and registered again at each page reload, not just a hard refresh?

This seems to be the case when testing on the preview: https://jupyterlite--1251.org.readthedocs.build/en/1251/_static/lab/index.html

image


Not sure if that's the desired behavior, as noted in https://github.com/jupyterlite/jupyterlite/issues/751.

andeplane: Yeah I think you're right. Question is if we can detect hard refresh (that's when you have to unregister + register). I'll look into it.

I'm curious: that issue from 2022 (still open) indicates that we already do that, but I don't think we do?

andeplane: See the updated PR which is much simplified. Back to the old detection logic on existence, but added the `unregister` part before registering a new service worker. I'll verify also with the PR server that the reported bug is fixed with this change.

jtpio: > I'm curious: that issue from 2022 (still open) indicates that we already do that, but I don't think we do?

Indeed. From that issue there is also https://github.com/jupyterlite/jupyterlite/pull/871 which seems to have changed a bit the logic of the service worker.

andeplane: Are the tests flaky, or can you rerun? I'm not sure what fails with Firefox (I can successfully switch between lab and notebook in firefox myself)

jtpio: Yes the firefox UI test is still a bit flaky.

andeplane: Yeah, I see latest master is failing on this as well. Suppose I consider this PR ready then.

jtpio (approved):
Thanks!

I think this looks good and seems to be behaving fine on the ReadTheDocs preview.

Leaving this PR opened until tomorrow in case @martinRenou would like to have a quick look as well. Otherwise we can go ahead and make a new patch release that would include this change.

PR #1264, opened 2023-12-08
javascript-kernel: use string repr of eval result

### LINKED ISSUE #1255: Javascript kernel writes non-string text/plain outputs in mimebundle
## Description

Noticed this one playing with the Javascript kernel. The .ipynb files produced with this kernel sometimes have `text/plain` outputs that aren't strings, which are rejected by nbformat.

Screenshot 2023-11-27 at 7 06 56 PM

JS kernel on left, Pyodide kernel on right. In the JS notebook, cells[0].outputs[0].data['text/plain'] is an integer when Pyodide renders it a string

This causes `nbformat.reads` to error out in validation

Screenshot 2023-11-27 at 7 08 40 PM

## Reproduce

1. Create a ipynb for the Javascript kernel
2. Type a non-string expression into a cell, like `1+1`, and evaluate it
3. Save the notebook and run the .ipynb file through `nbformat.reads`

Screenshot 2023-11-27 at 7 08 40 PM

## Expected

The `text/plain` output should be a string, like `"2"`, not a number like `2`.

A similar test in a Pyodide notebook gives a string result, which is accepted by `nbformat.reads`.

## Context

The relevant line in the kernel is here, the output of `self.eval()` in the worker is used directly: https://github.com/jupyterlite/jupyterlite/blob/main/packages/javascript-kernel/src/worker.ts#L52

- JupyterLite version: v0.1.2 (also affects master)
- Operating System and version: macOS 13
- Browser and version: Chrome 119

tomjakubowski:
## References

Closes #1255

## Code changes

Adds the `object-inspect` library to the Javascript kernel, and uses that to represent values evaluated in the Javascript kernel.

## User-facing changes

* In Javascript notebooks, objects print key/value instead of `[object Object]`, arrays are printed with enclosing brackets.
* .ipynb files made by the Javascript kernel meet nbformat's schema

N/A

### DIFF
--- a/packages/javascript-kernel/src/worker.ts
+++ b/packages/javascript-kernel/src/worker.ts
@@ -1,4 +1,6 @@
 import { IJavaScriptWorkerKernel } from './tokens';
+import { KernelMessage } from '@jupyterlab/services';
+import objectInspect from 'object-inspect';
 
 export class JavaScriptRemoteKernel {
   /**
@@ -44,13 +46,17 @@ export class JavaScriptRemoteKernel {
   async execute(content: any, parent: any) {
     const { code } = content;
     try {
-      const result = self.eval(code);
+      const result = self.eval(code) as unknown;
       this._executionCount++;
 
-      const bundle = {
-        data: {
-          'text/plain': result,
-        },
+      const textPlain = this._inspect(result);
+      const data: { ['text/plain']?: string } = {};
+      if (typeof textPlain === 'string') {
+        data['text/plain'] = textPlain;
+      }
+
+      const bundle: KernelMessage.IExecuteResultMsg['content'] = {
+        data,
         metadata: {},
         execution_count: this._executionCount,
       };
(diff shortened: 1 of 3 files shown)


packages/javascript-kernel/src/worker.ts
```diff
@@ -44,16 +45,16 @@ export class JavaScriptRemoteKernel {
...
-          'text/plain': result,
+          'text/plain': inspect(result),
         },
         metadata: {},
         execution_count: this._executionCount,
-      };
+      } satisfies KernelMessage.IExecuteResultMsg['content'];
       postMessage({
         bundle,
         type: 'execute_result',
```
tomjakubowski, line 60: there is a typesafe generic factory function `KernelMessage.createMessage()` which would be useful throughout this module. They won't work directly with the language currently spoken over `postMessage()`, but we control both ends and can change it to accommodate them

jtpio:
Thanks @tomjakubowski for looking into this!

Trying it on the ReadTheDocs preview, this also seems to be adding an extra `undefined` when printing text with `console.log` or `console.error`: https://jupyterlite--1264.org.readthedocs.build/en/1264/_static/lab/index.html

image

tomjakubowski:
Ah, good catch! Probably makes sense to just give no output when the result evals to undefined, like an ipython notebook when a cell evals to None, yeah?

tomjakubowski:
I updated it to not include a `text/plain` in the bundle when the cell evals to `undefined`

Screenshot 2023-12-13 at 4 54 58 PM

jtpio (approved):
Thanks!

Issue #1284, opened 2024-01-08
Provide the server router from a token rather than an app attribute


fcollonval:
### Problem

In https://github.com/jupyterlite/jupyterlite/pull/506 (related issue https://github.com/jupyterlite/jupyterlite/issues/390), the router has been exposed from the app for extension to register new route.

The work on notebook 7 demonstrates that a robust approach for providing conditional feature on the application is through token being provided or not.

### Proposed Solution

I propose to expose the server router through a new token `IServerRouter`.

The interesting consequence is that the `app` argument for `activate` function will keep the consistent `JupyterFrontend` and `JupyterFrontendPlugin` typing removing a customization.

jtpio:
> The interesting consequence is that the `app` argument for `activate` function will keep the consistent `JupyterFrontend` and `JupyterFrontendPlugin` typing removing a customization.

Currently the `app` would still need to be a `JupyterLiteServer` because it's a different app from the ones regular JupyterLab plugins get (`Application<never>`):

https://github.com/jupyterlite/jupyterlite/blob/839741ede0f35e5e1d939b8857bd92405c74bc4b/packages/server/src/app.ts#L52

Also the JupyterLite server extensions are currently discovered with the `jupyterlite` metadata in `package.json` so the plugins can be registered on the proper app at page load.

Ideally this additional application could / should be removed in the long run. For now its main purpose has been to keep things clearly separate between "frontend" plugins and "server" plugins in lite. Maybe something like this upstream proposal could help in the long run: https://github.com/jupyterlab/jupyterlab/issues/15329

jtpio:
> The service manager customization proposed in [jupyterlab/jupyterlab#13558](https://github.com/jupyterlab/jupyterlab/pull/13558) is enough as a first step - and I have an idea to improve the security point raised in that attempt.

Would you mind sharing more thoughts about that idea in https://github.com/jupyterlab/jupyterlab/issues/15329, since it's now the closest issue tracking this enhancement?

As mentioned in https://github.com/jupyterlab/jupyterlab/issues/15329, I was hoping we could flatten most of the underlying services in `ServiceManager` so they become easily swappable via extensions. And as mentioned in that issue, one challenge may be backwards compatibility. Thinking more about it, backwards compatibility (so usage of `app.serviceManager` is not broken) should be a hard requirement to avoid extension authors fatigue.

Thanks!

PR #1285, opened 2024-01-08
Provide the server router from a token rather than an app attribute

fcollonval:
## References

Fixes partly #1284. The next step is to make the `JupyterFrontEnd` more composable to allow dropping the namespacing separation between JupyterLite and JupyterLab/Notebook.

The goal is for third-party developer to be able to hook part of their Jupyter extension into the JupyterLiteServer without additional burden of namespacing plugins (what can be lifted if the Lumino `Application` is changed to a composition of plugins).

## Code changes

Make the server request router available from a token.

## User-facing changes

None

None - the `router` attribute has been marked as deprecated but not removed.

### DIFF
--- a/docs/howto/extensions/kernel.md
+++ b/docs/howto/extensions/kernel.md
@@ -19,7 +19,7 @@ const kernel: JupyterLiteServerPlugin<void> = {
   id: 'my-custom-kernel:plugin',
   autoStart: true,
   requires: [IKernelSpecs],
-  activate: (app: JupyterLiteServer, kernelspecs: IKernelSpecs) => {
+  activate: (app: IJupyterLiteServer, kernelspecs: IKernelSpecs) => {
     kernelspecs.register({
       spec: {
         name: 'custom',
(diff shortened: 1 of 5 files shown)


jtpio:
Thanks @fcollonval for looking into this!

> The goal is for third-party developer to be able to hook part of their Jupyter extension into the JupyterLiteServer without additional burden of namespacing plugins (what can be lifted if the Lumino `Application` is changed to a composition of plugins).

Curious if being able to swap the router can already be useful to third-party extensions? Compared to being able to swap the in-browser services (sessions, kernels, contents...), which should already be possible? (just wondering about potential use cases).

fcollonval:
> Curious if being able to swap the router can already be useful to third-party extensions?

You mean accessing the JupyterLite router to extend it, not providing an alternative implementation?

If so, yes it will answer the question. But the challenge that we should lift is the namespacing of plugins. For example, how would you provide the router that is part of JupyterLite server plugins space? Going that road seems more complex and hacky than removing the namespacing following the ability to customize upstream the service manager in the lab application.

PR #1291, opened 2024-01-15
Include src in published library for sourcemap resolution

fcollonval:
## References

Similar to https://github.com/jupyterlab/jupyterlab/pull/14028

## Code changes

Includes files in ./src in npm package dist, so that the relative paths in the built JS source maps can be resolved. Adds an integrity check for the same, so that it will also be enforced for packages in the future.

## User-facing changes

None (slightly larger dist on disk, but not loaded in browser unless debugging).

None


jtpio (approved):
Thanks!

Issue #1103, opened 2023-07-04
Aggressive caching by the service worker


stevejpurves:
# Context

It seems like the Jupyterlite service worker aggressively caches all get requests, which can cause problems when jupyterlite-core assets are used in a different web contexts. The web context I'm looking at in particular is mystmd websites, and other sites that use the `thebe` library.

from [the code here](https://github.com/jupyterlite/jupyterlite/blob/f98c5c0f90f25466aea6beeb4f2bd2ce8549aa68/packages/server/src/service-worker.ts#L111) like the SW will cache all GET requests, on http[s] that does not have the `/api/' in its path.

```
/**
 * Whether the fallback behavior should be used
 */
function shouldDrop(request: Request, url: URL): boolean {
  return (
    request.method !== 'GET' ||
    url.origin.match(/^http/) === null ||
    url.pathname.includes('/api/')
  );
}
```

This causes a few issues for our usage in mystmd, namely:

* the interactive editing experience for authors needs hard refreshes to get content at times
* we get issues when we update styles, and styles are sometimes not applied correctly.
* Deployed websites created by mystmd user can end up being agreessiely cached and their content, styling or js assets don;t update without a hard reload or unregistering of the service worker.

We're looking at how to improve this situation while keeping the service worker in place to support the file system functionality.

# Questions

* are there any options or configuration available to us that would allow us to disable caching, while maintaining the message passing features needed for filesystem?
* [abit of an aside but out off interest...] is the caching behaviour of the service worker essential to the operation of the filesystem, or is it just enabled as a nice to have?
* could the service worker responsibilities be split into two service workers? one looking after the filesystem broadcast channel and a second providing the caching, with an option to disable the caching?
* from tests so far it seems like setting `"disabledExtensions": ["@jupyterlite/server-extension:service-worker"]` has no effect and the service worker still installs and activates, this was set both programmatically and on page as per [the example here](https://jupyterlite.readthedocs.io/en/latest/howto/configure/advanced/extensions.html#disabling-extensions-at-runtime) - is disabling still possible?

jtpio:
Thanks @stevejpurves for opening the issue :+1: 

Right, the service worker has been causing a couple of issues related to caching.

Agree it would be useful to provide options or way to disable some of its features. Maybe as page config options, or by splitting functionality into separate plugins so they can be disabled.

[...]

Issue #1294, opened 2024-01-19
Make the service worker an opt-in, not an opt-out


martinRenou:
cc. @trungleduc @DerThorsten @jtpio 

### Why the service worker?

The service worker fills two purposes as of today:
- file-system access for Emscripten-based kernels
- caching (with the by-product of being able to install jupyterlite as a PWA)

### What's the issue?

The service worker is also the cause of many issues in JupyterLite:
- aggressive caching causing unintended behaviors (seeing a jlite webpage whereas the server is not serving such page, outdated content etc). See https://github.com/jupyterlite/jupyterlite/issues/1103
- kernel crashing due to the flaky filesystem-service worker communication. https://github.com/voila-dashboards/voici/issues/92 
- outdated service worker file cached by the browser

### What could we do in the future?

At some point, it would be nice to ditch the service worker for file-system access, and use a probably more robust approach using shared array buffers. In that case, file-system access support will be an opt-in (behind a browser flag).

### What can we do now?

Aggressive caching may be better as an opt-in, as it's causing lots of issues for normal users who may not know how/what to debug. Also, given that we may have file-system access behind a flag at some point (not enabled in the normal usage) we can consider doing this now, even before we make use of shared array buffers. 

For those reasons, I would like to suggest putting the service worker behind a flag:
```bash
jupyter lite build . --enable-service-worker
```

jtpio:
Thanks @martinRenou for opening the opening, this looks like a good summary of the current service worker issues :+1: 

> At some point, it would be nice to ditch the service worker for file-system access, and use a probably more robust approach using shared array buffers. In that case, file-system access support will be an opt-in (behind a browser flag).

Right. As an alternative to ditching the service worker completely, maybe have a system of priorities based on feature availability? For example:

- if `SharedArrayBuffer` is available, use it
- if not, check if service workers are available and use them (current behavior)
- if not, then contents is not sync (current behavior in a Firefox private window)

This would also be useful for https://github.com/jupyterlite/jupyterlite/pull/460.

> For those reasons, I would like to suggest putting the service worker behind a flag:

Just to clarify, are you suggesting putting the caching behavior behind a flag, not the whole service worker registration? Which would be similar to the idea expressed in https://github.com/jupyterlite/jupyterlite/issues/1103#issuecomment-1691751425.

Otherwise users might be surprised if by default they can't access their files from the kernel.

martinRenou:
> - if SharedArrayBuffer is available, use it
[...]

Issue #969, opened 2023-01-31
Loading demo in Chrome crashes due to fetching non-existent URL, apparently due to stale browser cache


jameshfisher:
## Description

As of today, the hosted demo does not load for me, due to a 404 when initializing. This seems to _only_ happen in my specific Chrome instance, and so seems to be tied to some hidden stale browser state.

## Reproduce

0. Be me on my Chrome instance
1. Go to https://jupyterlite.github.io/demo/repl
2. Open JS console
3. See error:

```
Uncaught (in promise) PythonError: Traceback (most recent call last):
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 161, in _fetch_bytes
    return await fetch_bytes(self.url, fetch_kwargs)
  File "/lib/python3.10/site-packages/micropip/_compat_in_pyodide.py", line 30, in fetch_bytes
    result_bytes = await (await pyfetch(url, **kwargs)).bytes()
  File "/lib/python3.10/pyodide/http.py", line 185, in bytes
    self._raise_if_failed()
  File "/lib/python3.10/pyodide/http.py", line 112, in _raise_if_failed
    raise OSError(
OSError: Request for https://jupyterlite.github.io/demo/build/pypi/./pyolite-0.1.0b17-py3-none-any.whl failed with status 404: 
```

Indeed, https://jupyterlite.github.io/demo/build/pypi/./pyolite-0.1.0b17-py3-none-any.whl is a 404.

However, all other browser instances fetch https://jupyterlite.github.io/demo/build/pypi/./pyolite-0.1.0b18-py3-none-any.whl, which loads normally.

## Expected behavior

REPL loads normally.

If broken due to some stale state, clearing all site data should fix the problem.

## Context

This only started a few hours ago.

This does not happen in incognito, or any other browser.

The only way I found to fix the issue was to open the Network tab, check "Disable cache", then reload.

The following did NOT fix the issue:

* Ctrl-Shift-R ("hard refresh").
* Clearing site data in the Application > Storage tab.
* Unregistering service workers.
* Restarting browser.

- Operating System and version: Windows
- Browser and version: Chrome Version 109.0.5414.120 (Official Build) (64-bit)

jameshfisher:
Its complaint seems valid, this is a 404: https://jupyterlite.github.io/demo/build/pypi/./pyolite-0.1.0b17-py3-none-any.whl

jameshfisher:
[...]

Issue #815, opened 2022-09-27
JupyterLite version caching in Google Chrome.


kolibril13:
For https://github.com/Octoframes/jupyter_compare_view  I've just updated the JupyterLite from version `0.1.0-beta.11` to 
`Version 0.1.0-beta.13` (See https://github.com/Octoframes/jupyter_compare_view/commit/f87c300aaa5268c4b16a4876ae321233364a3d30). 
The last weeks, I've used the jupyter_compare_view demo with JupyterLite version `0.1.0-beta.11` in Google Chrome and not in other browsers.
15 minutes after updating, when I  opened the demo in chrome, it still showed JupyterLite version `0.1.0-beta.11` in Help-> About JupyterLite.
When I press  `⌘ Cmd + ⇧ Shift + R` to refresh the browser and clear the cache (https://bacreative.com.au/how-to-hard-refresh-browser-and-clear-cache-using-a-mac/#:~:text=To%20hard%20refresh%20on%20Google,key%20and%20then%20press%20R) it still showed the old version.
When I opened the demo in Microsoft Edge or Safari, then the correct new version `Version 0.1.0-beta.13` was shown.

After opening another JupyterLite instance from another JupyterLite project in Google Chrome with `Version 0.1.0-beta.13`, the jupyter_compare_view project showed the new version number.

I think this is might be a caching problem in Google Chrome.

kolibril13:
Hi @jtpio !
Thanks for the release of Jupyter beta.14.
I've just tested this and here are two reports:
First: When I type 
```py
import pathlib
p = pathlib.Path("greeting.txt")
p.exists()
```
In Microsoft Edge and in Safari, that returns "False "as expected.
In Google Chrome, the same script returns the below error message.(even after pressing ⌘ Cmd + ⇧ Shift + R to refresh the browser and clear the cache). However, when I open a private tab in Google Chrome that problem does not occur, so it must be somehow cache related. Or maybe cookie related?
```
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Cell In [2], line 3
      1 import pathlib
      2 p = pathlib.Path("greeting.txt")
----> 3 p.exists()

File /lib/python3.10/pathlib.py:1288, in Path.exists(self)
   1284 """
   1285 Whether this path exists.
   1286 """
   1287 try:
-> 1288     self.stat()
   1289 except OSError as e:
   1290     if not _ignore_error(e):

[...]

PR #1316, opened 2024-02-19
Add an option to enable the service worker cache

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1103

Also related to:
- https://github.com/jupyterlite/jupyterlite/issues/1294
- https://github.com/jupyterlite/jupyterlite/issues/969
- https://github.com/jupyterlite/jupyterlite/issues/815

The service worker caching is probably one of the biggest frustration points for JupyterLite users. Making it an opt-in might help make the user experience smoother.

## Code changes

- [x] Add an (opt-in) option to enable the Service Worker cache.
- [x] Read from page config
- [x] Add to migration guide

## User-facing changes

Hopefully less issues about missing files and after updating JupyterLite versions.

More of a change in behavior than a breaking change. Still need to be documented in the migration guide though.

### DIFF
--- a/app/jupyterlite.schema.v0.json
+++ b/app/jupyterlite.schema.v0.json
@@ -187,6 +187,11 @@
           "description": "Whether to enable collaborative editing over WebRTC. Should be paired with the ``?room=<room name>`` URL parameter",
           "type": "boolean",
           "default": false
+        },
+        "enableServiceWorkerCache": {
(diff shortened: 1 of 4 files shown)


jtpio:
Need to do a bit more testing to make sure this actually has a real effect.

jtpio:
Getting this in so it can be tested in the next rc.

Issue #1335, opened 2024-03-04
Fix browser caching of existing service workers


andeplane:
## Description
I haven't seen this problem explicitly be mentioned, but maybe https://github.com/jupyterlite/jupyterlite/issues/815 is the same one.

When there is a new version of the service worker, there is no information that prevents caching in the browser. The URL fetched [here](https://github.com/jupyterlite/jupyterlite/blob/main/packages/server/src/service-manager.ts#L45) is e.g. https://jupyterlite.readthedocs.io/en/stable/_static/service-worker.js, e.g. no information about the version. 

This means the preloaded service worker will be the one installed, and there is thus a potential change in the API between client and service worker.

## Reproduce
1) load jupyterlite
2) deploy a breaking change in service worker
3) load jupyter lite again, it crashes

We are using JupyterLite and have seen problems when bumping to a new version due to this. 

## Expected behavior
Old service worker should be unregistered to get the latest version.

## Context

- JupyterLite version: 0.2.3
- Operating System and version:
- Browser and version: Chrome 122.0.6261.94

jtpio:
> When there is a new version of the service worker, there is no information that prevents caching in the browser. The URL fetched [here](https://github.com/jupyterlite/jupyterlite/blob/main/packages/server/src/service-manager.ts#L45) is e.g. https://jupyterlite.readthedocs.io/en/stable/_static/service-worker.js, e.g. no information about the version.

There used to be hash in the file name of the service worker, but it was removed in https://github.com/jupyterlite/jupyterlite/pull/1159 because it was creating issue for downstream applications trying to reuse it, since the hash would depend on the JupyterLite version used for building the application.

Maybe this idea of hash should be kept, and instead but passed as a query argument? So requesting the service worker file would become https://jupyterlite.readthedocs.io/en/stable/_static/service-worker.js?v=123456789

andeplane:
Yep, that is a god alternative to the proposal in #1336. I'm not sure about the life cycle of these service workers then though, if the old versions keep existing and maybe it is cleaner to unregister + register.

jtpio:
> maybe it is cleaner to unregister + register.

Yeah, since the main motivation for having the service worker in the first place was to support accessing files from a kernel.

PR #1336, opened 2024-03-04
Unregistering old service workers when there is a new version 

andeplane:
## References
Closes https://github.com/jupyterlite/jupyterlite/issues/1335

## Code changes

I added a test to see which service worker version that is installed and will unregisted any old ones if there are a new version, or if no version is registered.

I would like to read the version from a global place rather than a const variable in the top of the file, so input in this is appreciated. 

Alternatively, adding a version (hash) to the service worker URL (which today is without version e.g. https://jupyterlite.readthedocs.io/en/stable/_static/service-worker.js ) would also work.

## User-facing changes
Breaking changes between service workers will no long lead to kernel death.

### DIFF
--- a/packages/server/src/service-manager.ts
+++ b/packages/server/src/service-manager.ts
@@ -35,10 +35,11 @@ export class ServiceWorkerManager implements IServiceWorkerManager {
     return this._ready.promise;
   }
 
-  private unregisterOldServiceWorkers = () => {
+  private unregisterOldServiceWorkers = (scriptURL: string) => {
+    const versionKey = `${scriptURL}-version`;
     // Check if we have an installed version. If we do, compare it to the current version
     // and unregister all service workers if they are different.
-    const installedVersion = localStorage.getItem('jupyterlite-version');
+    const installedVersion = localStorage.getItem(versionKey);
 
     if ((installedVersion && installedVersion !== VERSION) || !installedVersion) {
       // eslint-disable-next-line no-console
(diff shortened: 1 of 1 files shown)


packages/server/src/service-manager.ts
```diff
@@ -33,7 +35,32 @@ export class ServiceWorkerManager implements IServiceWorkerManager {
     return this._ready.promise;
   }
 
+  private unregisterOldServiceWorkers = () => {
+    // Check if we have an installed version. If we do, compare it to the current version
+    // and unregister all service workers if they are different.
+    const installedVersion = localStorage.getItem('jupyterlite-version');
```
jtpio, line 41: Not sure how this approach would work if multiple JupyterLite deployments are hosted under the same host? Could there be an issue accessing the same local storage, if for example there is a JupyterLite site on 0.2.2 and another one on 0.2.3?

jtpio: An example is a GitHub organization having multiple repo with a JupyterLite deployment on GitHub Pages.

andeplane: Yeah good point. this can be a problem if they share the domain (thus localstorage) like you say.

If site A has 0.2.2 and is loaded, then site B with 0.2.3 is loaded will override version stored in local storage to 0.2.3. Now if site A is updated to new version (0.2.3) it will wrongly _not_ unregister its old version (0.2.2).

What we can do is switch out the localstorage key to include e.g. the full path to the service worker which thus will be unique across the two deployments even if they are on the same domain?

andeplane:
Updated with a new suggestion that should solve this problem. What do you think?

jtpio:
> Updated with a new suggestion that should solve this problem. What do you think?

Thanks for looking into this :+1: 

Sounds like it should work? Also cc @martinRenou who may be interested

jtpio:
I guess we could also make use of the current `0.3.0` pre-releases to try this: releasing `0.3.0rc0` and check whether upgrading from `0.3.0b0` to `0.3.0rc0` works as expected.

jtpio (approved):
Thanks!

Let's get it in so it can be tested more easily.

Issue #1371, opened 2024-04-15
Trying to prevent "FileNotFoundError: [Errno 44]"


pierrepo:
This issue is both a bug report and a documentation improvement, so I report it as a feature request :-)

### Problem

I'm using JupyterLite for teaching because it could be easily deployed on GitHub pages.

My students are sometimes reporting the error message `FileNotFoundError: [Errno 44] No such file or directory` documented in the [Troubleshooting](https://jupyterlite.readthedocs.io/en/stable/troubleshooting.html) section. Here is my attempt to better understand how and when this error occurs:

From all the tests I performed, it looks like this error could be reproduced:
- with Mozilla Firefox (both on Windows and Linux)
- when the user executes a Python code cell to access/read a file **before** the notebook is fully loaded. By this, I mean the Kernel status is *Unknown* 
![image](https://github.com/jupyterlite/jupyterlite/assets/733861/e5659073-9297-4293-b13a-225321085f9f)
instead of *Iddle*
![image](https://github.com/jupyterlite/jupyterlite/assets/733861/91043d3d-c31b-41e9-9e2b-a7373c144931)

Once the `FileNotFoundError: [Errno 44] No such file or directory` is obtained, reloading the page or the kernel, does not help. Files can't be read even if the notebook is properly loaded and the kernel status is *Iddle*. The only solution is then to clear the browser cache.

### Proposed Solution

1. Documentation-based solution: mention in the documentation that the notebook should be fully loaded (and the kernel Idlle) before trying to access any local files.

2. Code-based solution: prevent the user to perform any interaction with the notebook until the kernel is iddle. This could be an opt-in option has this solution may not fit all users.

Carreau:
Thanks for the report, 

I never tried on Firefox for mac (124.0.2), but jupyterlite seem to not work for me:

```
Error
Repository is read only
```

Which is weird. 

On chrome (123.0.6312.107) I have no issues with executing, it will just wait until the kernel is ready.

But we can indeed add a troubleshooting section warning about this problem.

ArturoAmorQ:
> From all the tests I performed, it looks like this error could be reproduced:
   with Mozilla Firefox (both on Windows and Linux)

This issue also seems to appear in chrome-based browsers.

ogrisel:
I can trigger `FileNotFoundError` but only if I open the same jupyterlite site in 2 concurrent browser tabs as reported here: #1582. But I don't think it's related to executing cells when Kernel status is Unknown.

[...]

PR #1372, opened 2024-04-15
[DOCS]: Suggest to wait for kernel starting to prevent to prevent "Fi…

Carreau:
…leNotFoundError: [Errno 44]"

See #1371, but it would be good to find a code workaround.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -12,6 +12,13 @@ FileNotFoundError: [Errno 44] No such file or directory: 'data/iris.csv'
 
 ![a screenshot showing an error while trying to access a local file from the kernel](https://github.com/jupyterlite/jupyterlite/assets/591645/3edffdc3-77ef-45fe-8a4a-8cd7147dd235)
 
+This seem to happen when code try to get executed before a kernel is fully
+ready. See [issue #1371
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks!

> it would be good to find a code workaround.

Yes. This is likely still related to the various caching issues folks have been reporting, for example: https://github.com/jupyterlite/jupyterlite/issues/969

docs/troubleshooting.md
```diff
@@ -12,6 +12,13 @@ FileNotFoundError: [Errno 44] No such file or directory: 'data/iris.csv'
 
 ![a screenshot showing an error while trying to access a local file from the kernel](https://github.com/jupyterlite/jupyterlite/assets/591645/3edffdc3-77ef-45fe-8a4a-8cd7147dd235)
 
+This seem to happen when code try to get executed before a kernel is fully
+ready. See [issue #1371
+](https://github.com/jupyterlite/jupyterlite/issues/1371). If this regularly
+happen, please try to wait until the kernel indicator is ready before starting
```
jtpio, line 18:
```suggestion
happens, please try to wait until the kernel indicator is ready before starting
```

Carreau:
Thanks, suggestions applied.

Issue #1041, opened 2023-04-05
Switch to `Contents.IDrive` for the JupyterLite in-browser contents manager


jtpio:
### Problem

Currently the JupyterLite contents manager is built into the JupyterLite Server: https://github.com/jupyterlite/jupyterlite/tree/main/packages/contents

This could be an issue when site deployers want to have better control on where the files are stored.

### Proposed Solution

Move the contents manager to be a `Contents.IDrive` instead. This has a couple of benefits:

- it should (hopefully) be easier to swap the contents manager so site deployers could develop their own drive and use it in JupyterLite. For example for storing the files on a distant server
- help use multiple drives in JupyterLite
- follows the new RTC implementation, which is now also an `IDrive`: https://github.com/jupyterlab/jupyter_collaboration/blob/eb8456d81a60a2db6cbb14249df31f6e08493c02/packages/docprovider/src/ydrive.ts#L24
- the extension could be used in stock JupyterLab, Binder or other lab-based applications

There could be a new `jupyterlab-browser-contents` (or similar name) package that is basically a normal JupyterLab extension.

The extension could live in this repo under `py/jupyterlab-browser-contents`, or https://github.com/jupyterlite/jupyterlab-browser-contents.

### Additional context

Questions to investigate:

- the standalone extension should probably just focus on managing content in `IndexedDB` via `localforage`
- check mounting the content in the kernel works with an `IDrive`
- check if the current layering of "server" files would work with that approach. Some files visible in the file browser might be the ones provided at build time with the deployment

**Edit**: linking to some upstream issues for more context:
- https://github.com/jupyterlab/jupyterlab/issues/7755
- https://github.com/jupyterlab/jupyterlab-github/issues/75
- https://github.com/jupyterlab/jupyterlab/issues/15143
- https://github.com/jupyterlab/jupyterlab/pull/15291

jtpio:
Adding to `0.2.0` for consideration.

bollwyvl:
`IDrive` is fine API, but the upstream UX of multiple, mutually-exclusive sidebars is pretty rough, making things like drag-and-drop pretty useless. Really probably needs to be something more like a tree, as some extensions have demonstrated.

getting the lazy layering is _clutch_, of course: not sure what to do about that. at present, i don't know if idrive stuff works much _at all_ inside the kernels.

[...]

PR #1383, opened 2024-05-16
Allow providing your own ContentsAPI implementation for the emscripten DriveFS

martinRenou:
This PR allows https://github.com/jupyterlite/xeus/pull/87.

- Allow easily overwriting the `ContentsAPI` implementation in the DriveFS
- Extract the logic from `broadcast.ts` which processes the driveFS request, so that it's easily reusable in another implementation


packages/contents/src/drivefs.ts
```diff
@@ -460,10 +486,26 @@ export class ContentsAPI {
...
+    this.ERRNO_CODES = ERRNO_CODES;
+  }
+
+  request(data: IDriveRequest): any {
+    const workerAPI = coincident(self) as typeof globalThis;
+
+    // Assuming the kernel provides a processDriveRequest method
+    // that processes the request with utils.processDriveRequest
+    return (workerAPI as any).processDriveRequest(data);
+  }
```
martinRenou, line 508: I should probably redesign this to allow the kernel author using whichever tool they want for the blocking communication (coincident, synclink, raw Atomics.wait calls etc...).

packages/contents/src/utils.ts
```diff
@@ -0,0 +1,110 @@
+import { PathExt } from '@jupyterlab/coreutils';
```
bollwyvl, line 1: Moving this to a standalone function, outside of a class, means it becomes rather harder to overload... and the original `switch` implementation wasn't very extensible anyway.

It's worth considering this as a new class, advertised as a full interface, and moving the  `switch` monolith into individual, well-typed, documented methods:

```ts
export class DriveContentsProcessor implments IDriveContentsProcessor {
  contentsManager: Contents.IManager
  async processDriveRequest(request) {
    switch(request.method) {
      // ...
    }
  }
  async onReaddir(request: IReadDirRequest): IReadDirResponse {...}
  // ... 
  async onPut(request: IPutRequest): IPutResponse {...}
}
```

This would then be an argument to `BroadCastChannelWrapper`, or a new plugin, depending on whether it's important something else get _exactly_ this instance.

martinRenou: I like the idea of splitting the switch into multiple methods of a class 👍🏽 Thanks for the suggestion

martinRenou: Though it does not have to be a class, it does not need to keep a state whatsoever. It could simply be free functions

```typescript
export async function readdir(request: IReadDirRequest): IReadDirResponse {...}
  // ... 
export async function put(request: IPutRequest): IPutResponse {...}
```

But I don't have a strong feeling about this.

martinRenou: At least the class could keep the content manager as a property

bollwyvl: ... _and_ be subclassable, vs a pile-of-functions, which puts us right back in the "can't be overloaded without extreme feats of monkeypatching" place.

If nothing else, the class-based approach provides better-discoverable, granular documentation for a logical set of tightly-coupled behaviors.

martinRenou: I updated the PR introducing the new class.

It's not a singleton provided by a plugin, we could potentially do that in a separate PR

martinRenou:
Converting into draft. I want to make extra testing on this before continuing.

martinRenou:
It looks like making backward incompatible changes to the DriveFS will break the UI-tests here because it's using pyodide-kernel which builds against a specific DriveFS version.

It's fine in this PR because I'd like it to be backward compatible, but it may be annoying some day if we want to make intentional breaking changes.

martinRenou:
This is ready for review. The UI-tests failure appear on the main branch too, fixing it in https://github.com/jupyterlite/jupyterlite/pull/1390

martinRenou:
CI is passing now @jtpio

packages/contents/src/drivecontent.ts
```diff
@@ -0,0 +1,235 @@
...
+        return this.lookup(request);
+      case 'mknod':
+        return this.mknod(request as IDriveMknodRequest);
+      case 'getattr':
+        return this.getattr(request);
+      case 'get':
+        return this.get(request);
+      case 'put':
+        return this.put(request as IDrivePutRequest);
+    }
```
bollwyvl, line 119: would consider `console.error`ing here in a `default`.

packages/contents/src/drivecontent.ts
```diff
@@ -0,0 +1,235 @@
...
+ * Class for processing a drive request from the DriveFS.
+ */
+export class DriveContentsProcessor implements IDriveContentsProcessor {
+  private contentsManager: Contents.IManager;
+
+  constructor(options: DriveContentsProcessor.IOptions) {
+    this.contentsManager = options.contentsManager;
+  }
+
+  async processDriveRequest(request: IDriveRequest): Promise<any> {
```
bollwyvl, line 99: Sooo... there's almost certainly a way to use a type discriminator to get well-typed values out, removing the need for `!` in a few places elsewhere.

Basically:

```ts
// somewhere, maybe in `tokens`?
export type TDriveMethod = 'readdir' | /* ... */ | 'put';
// somewhere, else maybe _also_ in `tokens`?
export interface IDriveRequest {
  method: TDriveMethod;
  // ...
}
```

Then each of the requests would refine it down to _exactly_ a member;

It's also possible to make wacky types that work as lookup tables, but I can't pull it out off the top of my head..

martinRenou: Sorry, I don't understand your comment, can you elaborate?

The `IDriveRequest` is already defined that way on the main branch: https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/drivefs.ts#L31

bollwyvl: Ok, cool, we have that. My point is we should not need `!` in new places elsewhere, as the return types should be inferrable from the inputs, based on `method`... seeing if i can gin up something more concrete...

martinRenou: Ah ok I see what you mean, I'll have a look

bollwyvl: Ooof... I _have_ done something like this:

https://github.com/jupyter-lsp/jupyterlab-lsp/pull/278/files#diff-7c7804a4b99d4b67b07ab518d033a6de85bdbf291e5e03400106901ff2501d5aR30

(most of the types in that signature are in that PR, but it's still a beast)

We'd basically want types for each of the methods (that actually overload the `method` with the actual value), then would be able to reference them with `<T, U, V>`. It's a lot, but adding _new_ APIs with `any` feels bad.

martinRenou: Thanks for the reference! I just pushed an attempt to implement that. It seems to compile?

packages/contents/src/drivecontent.ts
```diff
@@ -0,0 +1,244 @@
...
+  constructor(options: DriveContentsProcessor.IOptions) {
+    this.contentsManager = options.contentsManager;
+  }
+
+  async processDriveRequest<T extends TDriveMethod>(
+    request: TDriveRequest<T>,
+  ): Promise<TDriveResponse<T>> {
+    switch (request.method) {
+      case 'readdir':
+        return this.readdir(request as TDriveRequest<'readdir'>) as Promise<
```
martinRenou, line 95: I understand we have no choice to cast here. It seems you're casting to `any` in https://github.com/jupyter-lsp/jupyterlab-lsp/pull/278/files#diff-7c7804a4b99d4b67b07ab518d033a6de85bdbf291e5e03400106901ff2501d5aR30.

martinRenou: From outside calls of `processDriveRequest`, the response type is correctly computed.

martinRenou:
@bollwyvl please tell me what you think now!

afshin:
Since "contents" in our API language is typically plural, I would suggest renaming `drivecontent.ts` to `drivecontents.ts`

martinRenou:
I just rebased (and squashed to make my life easier during the rebase).

Thanks @afshin I resolved it.

packages/contents/src/drivefs.ts
```diff
@@ -41,13 +43,42 @@ export type TDriveMethod =
   | 'put';
 
 /**
- * Interface of a request on the /api/drive endpoint
+ * Type of the data argument for the drive request, based on the request name
  */
-export interface IDriveRequest {
```
jtpio, line 46: Maybe we could document some of these refactoring changes in https://jupyterlite.readthedocs.io/en/latest/migration.html.

Not expecting them to be use extensively downstream, but still useful to have them documented somewhere.

martinRenou: Done!

jtpio (approved):
Thanks!

Let's get this in and released in the next alpha, so it can be tested downstream.

@martinRenou @bollwyvl please feel free to open a new issue or PR if something needs to be changed or fixed, thanks!

michaelweinold:
@martinRenou, could you help me make sense of this new feature? 

Following a quick search, I was unable to find documentation on `ContentsAPI` functionality.
From your changes, I see, that:

https://github.com/jupyterlite/jupyterlite/blob/a9dc3206019f40dcff913dcdcabe64b0cdcfa4f1/packages/contents/src/drivefs.ts#L375

1. Would this allow me to chose a different file system as default for my JupyterLite site?
2. Is this PR at all related to other efforts related to the file system, like:

 - https://github.com/jupyterlite/jupyterlite/issues/1041

For context: I am still looking for a good way to use `sqlite3` databases in the default directory - which is currently not possible, as per:

 - https://github.com/jupyterlite/pyodide-kernel/issues/35

PR #1395, opened 2024-05-24
DriveFS: Handle receiving a stream instead of a node

martinRenou:
This will allow removing the monkey patch in https://github.com/jupyterlite/xeus/blob/main/src/worker.ts#L19

This monkey patch was introduced in https://github.com/jupyterlite/xeus-python-kernel/pull/27

### DIFF
--- a/packages/contents/src/emscripten.ts
+++ b/packages/contents/src/emscripten.ts
@@ -64,21 +64,32 @@ export function instanceOfStream(
 }
 
 export interface IEmscriptenNodeOps {
-  getattr(node: IEmscriptenFSNode): IStats;
-  setattr(node: IEmscriptenFSNode, attr: IStats): void;
-  lookup(parent: IEmscriptenFSNode, name: string): IEmscriptenFSNode;
+  getattr(node: IEmscriptenFSNode | IEmscriptenStream): IStats;
+  setattr(node: IEmscriptenFSNode | IEmscriptenStream, attr: IStats): void;
+  lookup(
+    parent: IEmscriptenFSNode | IEmscriptenStream,
+    name: string,
+  ): IEmscriptenFSNode;
   mknod(
-    parent: IEmscriptenFSNode,
+    parent: IEmscriptenFSNode | IEmscriptenStream,
     name: string,
     mode: number,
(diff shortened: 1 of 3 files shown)


packages/contents/src/emscripten.ts
```diff
@@ -57,6 +57,12 @@ export interface IEmscriptenStream {
...
 }
 
+export function instanceOfStream(
+  nodeOrStream: IEmscriptenFSNode | IEmscriptenStream,
+): nodeOrStream is IEmscriptenStream {
+  return 'node' in nodeOrStream;
+}
+
 export interface IEmscriptenNodeOps {
   getattr(node: IEmscriptenFSNode): IStats;
```
jtpio, line 67: Should this be updated according to the changes above?

```suggestion
  getattr(node: IEmscriptenFSNode | IEmscriptenStream): IStats;
```

jtpio: If so, probably also for the other methods.

martinRenou: Ah right I missed it, thanks!

packages/contents/src/emscripten.ts
```diff
@@ -57,22 +57,39 @@ export interface IEmscriptenStream {
...
+  nodeOrStream: IEmscriptenFSNode | IEmscriptenStream,
+): nodeOrStream is IEmscriptenStream {
+  return 'node' in nodeOrStream;
+}
+
 export interface IEmscriptenNodeOps {
-  getattr(node: IEmscriptenFSNode): IStats;
-  setattr(node: IEmscriptenFSNode, attr: IStats): void;
-  lookup(parent: IEmscriptenFSNode, name: string): IEmscriptenFSNode;
+  getattr(node: IEmscriptenFSNode | IEmscriptenStream): IStats;
```
jtpio, line 67: Looks like this change of signature can be considered as a breaking change?

Although it's probably not directly used that much downstream, should we still document it in the migration guide? https://jupyterlite.readthedocs.io/en/latest/migration.html

martinRenou:
(adding back the enhancement label for the CI check)

PR #1401, opened 2024-06-03
Improve error message for when jupyter-server is not installed

martinRenou:


py/jupyterlite-core/jupyterlite_core/addons/contents.py
```diff
@@ -156,7 +156,14 @@ def one_contents_path(self, output_file_dir, api_path):
 
         self.maybe_timestamp(self.output_files_dir)
 
-        from jupyter_server.services.contents.filemanager import FileContentsManager
+        try:
+            from jupyter_server.services.contents.filemanager import FileContentsManager
+        except ImportError as e:
+            raise RuntimeError(
+                """jupyter-server is not installed. You cannot add custom content to jupyterlite.
```
jtpio, line 163: Maybe this could / should be raised from the check above directly?

https://github.com/jupyterlite/jupyterlite/blob/a9dc3206019f40dcff913dcdcabe64b0cdcfa4f1/py/jupyterlite-core/jupyterlite_core/addons/contents.py#L148-L152

jtpio:
Maybe we can add a quick note to the migration guide, since this might cause a change of behavior in some cases? https://jupyterlite.readthedocs.io/en/latest/migration.html#to-0-4-0

jtpio:
Trying this locally, it looks like the error is raised even when running a `jupyter lite build` without the `--conents` option.

jtpio:
Converting back to draft as it would be great to:

- only raise the error if `contents` is specified but `jupyter-server` is not installed
- add a test

jtpio:
Bumping to `0.5.0` to avoid a change of behavior, until we figure out a good way to raise this error smoothly.

jtpio (approved):
Thanks!

PR #1405, opened 2024-06-06
Add docs about the use of `SharedArrayBuffer`

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1294

## Code changes

Update documentation to mention the use of the `SharedArrayBuffer`.

- [x] List the two ways for accessing contents
- [x] Document how to set the headers on some hosts
- [x] ~Document the GitHub Pages trick~ -> moved to https://github.com/jupyterlite/jupyterlite/issues/1409
- [x] ~Wait for the Pyodide kernel to also support this: https://github.com/jupyterlite/pyodide-kernel/pull/114~ -> tracked in https://github.com/jupyterlite/jupyterlite/issues/1408

## User-facing changes

Docs update.

None

### DIFF
--- a/docs/howto/content/python.md
+++ b/docs/howto/content/python.md
@@ -50,7 +50,7 @@ TODO: mention GitHub Pages
 
 ### 2. Service Worker
 
-If the `SharedArrayBuffer` are not available in the browser, JuptyerLite will default
+If the `SharedArrayBuffer` are not available in the browser, JupyterLite will default
 back to using the Service Worker (which was used by default until JupyterLite 0.4.0).
 
 Synchronized content works by mounting a custom [Emscripten Filesystem][fs] (FS) which
--- a/docs/howto/content/python.md
+++ b/docs/howto/content/python.md
@@ -23,10 +23,13 @@ exposed to the kernels with two different ways:
    `SharedArrayBuffer`)
 2. Via a Service Worker
 
-### 1. `Atomics.wait`
+### 1. Synchronous communication with the kernel over `Atomics.wait` (via `SharedArrayBuffer`)
 
 By default, if the kernel supports it, synchronous communication via `SharedArrayBuffer`
 will be used for accessing files from the kernels.
+
+This approach should be more robust and avoid most of the issues related to the Service
+Worker and browser caching.
 
 However they require setting proper HTTP headers when serving the JupyterLite
 application:
--- a/docs/howto/content/python.md
+++ b/docs/howto/content/python.md
@@ -48,14 +48,14 @@ npx static-handler --cors --coop --coep --corp ./
 ```
 
 ```{tip}
-See the various deployment guides in the [How-to section of the documentation](../).
+See the various deployment guides in the [How-to section of the documentation](../index.md).
 ```
 
 ```{note}
 File system access via `SharedArrayBuffer` is available with the following Python kernels:
 
-- `jupyterlite-pyodide-kernel`
-- `jupyterlite-xeus` with the Xeus Python kernel
+- ❌ `jupyterlite-pyodide-kernel`
+- ✅ `jupyterlite-xeus` with the Xeus Python kernel
 ```
 
 ### 2. Service Worker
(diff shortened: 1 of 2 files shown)


docs/howto/content/python.md
```diff
@@ -14,6 +14,45 @@ data = pd.read_csv('file.csv')
...
+npx static-handler --cors --coop --coep --corp ./
+```
+
+```{note}
+TODO: mention GitHub Pages
+```
+
+### 2. Service Worker
+
+If the `SharedArrayBuffer` are not available in the browser, JuptyerLite will default
```
martinRenou, line 53:
```suggestion
If the `SharedArrayBuffer` are not available in the browser, JupyterLite will default
```

martinRenou (commented):
Thanks for starting this!

The wording is very general and it looks like it's something that all kernels support. 

We should probably be clear about which kernels have this behavior.

jtpio:
> The wording is very general and it looks like it's something that all kernels support.

Yes. For now it is in this section of the documentation, which uses Python snippets as example: https://jupyterlite.readthedocs.io/en/stable/howto/content/python.html

For now we only list xeus python and pyodide as the two kernels supporting file access. I guess we could also list them on this page too to make it more explicit.

jtpio:
For GitHub Pages, the idea would be use https://github.com/gzuidhof/coi-serviceworker, which basically registers another service worker to set the HTTP headers on the fly.

However there is currently no proper way to inject extra scripts and files during a `jupyter lite build` in the generated `index.html`. Folks could of course edit the generated `index.html` in the `output-dir` directory, but that sounds error prone and not very convenient.

Another idea would be integrate the logic from https://github.com/gzuidhof/coi-serviceworker in the existing service worker directly, and put it behind a flag (opt-in) to enable the COOP / COEP header trick at runtime via a `jupyter-lite.json` for instance.

Or we leave the case of GitHub Pages on the side for now for the 0.4.0, as it would still use the existing Service worker approach for accessing content.

jtpio:
Tracking some documentation follow-ups in:

- https://github.com/jupyterlite/jupyterlite/issues/1409
- https://github.com/jupyterlite/jupyterlite/issues/1408

jtpio:
Moving forward with this to cut a beta afterwards. We can always iterate on the docs as needed.

juntyr:
This is such an exciting feature @jtpio!

I wanted to try out the new SharedArrayBuffer-based content sharing for my deploy of JupyterLite. In my requirements.txt, I set

```
jupyterlite-core==0.4.0b1
jupyterlab~=4.2.3
jupyterlite-pyodide-kernel==0.4.0b0
```

My previous jupyter-lite.json config disabled the service worker and only used in-memory storage

```json
{
    "jupyter-lite-schema-version": 0,
    "jupyter-config-data": {
        "disabledExtensions": ["@jupyterlite/server-extension:service-worker"],
        "enableMemoryStorage": true,
        "contentsStorageDrivers": ["memoryStorageDriver"]
    }
}
```

Since I'm also using the SharedArrayBuffer-based trick for synchronous communication, my site is already deployed with the required HTTP CORS headers.

Still, when using the site, I get the following warnings:

```
JupyterLite ServiceWorker not available [index.js:185:24]
Kernel filesystem and JupyterLite contents will NOT be synced [index.js:188:24]
```

and the `/drive` mount is missing:

```python
from pathlib import Path
list(Path("/").iterdir())

# [PosixPath('/tmp'),
#  PosixPath('/home'),
#  PosixPath('/dev'),
#  PosixPath('/proc'),
#  PosixPath('/lib'),
#  PosixPath('/usr')]
```

Is there anything else I need to change in my settings to enable the new SharedArrayBuffer-based content sharing while also continuing to not enable the service worker?

Thanks for your help!

jtpio:
Thanks @juntyr for the details.

At first glance the dependencies and configuration look fine. Would you able to check in the browser dev tools console the `SharedArrayBuffer` are enabled?

image

juntyr:
I get the same console output as you - perhaps the link to my deployment might help: https://esiwace3-compression-lab.onrender.com/lab/index.html

jtpio:
@juntyr let's open a new issue or discussion to keep track of this?

Issue #1412, opened 2024-06-15
Contents Manager integration broken in pyodide-kernel on main


bollwyvl:
## Reference

- observed on review of e.g. https://github.com/jupyterlite/pyodide-kernel/pull/115 (and related)

## Description

It appears something changed substantially at the contents message-passing layer in #1383 (vs #1380) that has thoroughly broken `pyodide-kernel`'s ability to work with files in the virtual drive. 

While this is appearing in `pyodide-kernel`, I think it is lower level, and haven't yet done the deeper look into what might have changed to be emitting bad JSON which causes pyodide to crash.

The most thorough comparison file is probably `pyodide/virtual_drive.ipynb` which has decent coverage of what is known to be (im)possible, but the simple case below should be enough to get back to a breakpoint.

## Reproduce

1. Go to https://jupyterlite--1383.org.readthedocs.build/en/1383/_static/lab/index.html
2. Start a new pyodide notebook (might have to go to another folder because jupyterlab-git?)
3. Create a cell

```py
from pathlib import Path
Path("foo").mkdir()
```

5. See permanent `[*]` and irrecoverable hang 
6. Open browser console, see:

```
pyodide.asm.js:9 
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at De.request (worker.js:4:3262)
    at De.mknod (worker.js:4:3478)
    at Ie.mknod (worker.js:4:2374)
    at Object.mknod (pyodide.asm.js:9:207624)
    at Object.mkdir (pyodide.asm.js:9:207839)
    at ___syscall_mkdirat (pyodide.asm.js:9:259719)
    at pyodide.asm.wasm:0x440f0e
    at pyodide.asm.wasm:0x386ee5
    at pyodide.asm.wasm:0x269427
pyodide.asm.js:9 Stack (most recent call first):
pyodide.asm.js:9   File "/lib/python311.zip/pathlib.py", line 1116 in mkdir
pyodide.asm.js:9   File "<ipython-input-2-a1f4a7c00aea>", line 2 in <module>
pyodide.asm.js:9   File "/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3577 in run_code
pyodide.asm.js:9   File "/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3517 in run_ast_nodes
[...]

PR #1418, opened 2024-06-18
Fix some drive methods return types

martinRenou:
## References

Looking into fixing https://github.com/jupyterlite/jupyterlite/issues/1412

This bug seems to have been introduced in #1383


bollwyvl (approved):
Running through `virtual_drive.ipynb` looks fine on RTD :tada:

PR #1443, opened 2024-07-24
docs: update config file example with correct output dir parameter name

zonca:
The correct naming of the output directory configuration option is `output_dir` and not `outputDir`: https://github.com/jupyterlite/jupyterlite/blob/0b0d70620827d02b3e0e9462dc75230b6c7c7732/py/jupyterlite-core/jupyterlite_core/config.py#L139


jtpio (approved):
Thanks!

PR #1445, opened 2024-07-25
Update `yarn.lock`

jtpio:


bollwyvl:
Some things we could do for the future:
- add the `dedupe` check to a lint stage
- when `$CI`, add the output of the dist size to the markdown summary
  - or make an actual comment/issue
- add high-water marks for size (or compare against previous tarballs on pypi)

PR #1450, opened 2024-08-01
Update Vercel Setup Instructions for Node.js v20

michaelweinold:
After I found that current Vercel builds of JupyterLite pages fail, @diegofcornejo quickly found that the issue is with additional requirements of Node.js v20. 

## References

- https://github.com/diegofcornejo/vercel-jupyter-lite/issues/3#issuecomment-2263443845

## Code changes

`-y` flag added to `yum install wget -y`


jtpio (approved):
Thanks!

PR #1355, opened 2024-03-25
Update to Python 3.11 for the dev environment

jtpio:
## References

Update from Python 3.10 to Python 3.11 for the dev setup.

## Code changes

Dev dependency update.

## User-facing changes

None

## Backwards-incompatible changes

None

PR #1016, opened 2023-03-20
Require Python 3.8, test on Python 3.11

jtpio:
## References

Follow-up to https://github.com/jupyterlite/jupyterlite/pull/1001

## Code changes

- [x] Require Python 3.8
- [x] Test on Python 3.11 on CI

## User-facing changes

Site deployers will need Python 3.8+ to use the `jupyter-lite` CLI.

## Backwards-incompatible changes

None

PR #1462, opened 2024-08-20
Update to Python 3.12 for the dev environment and CI

EwoutH:
Python 3.12 is now almost a year available, so it looks like a good time to update the dev environment and the CI to it. Python 3.12 is faster and offers better error messages, among other features.

This PR does two things:

- Enable Python 3.12 testing on the CI
- Update from Python 3.11 to Python 3.12 for the dev setup.

If preferred, the minimum Python version could also be increased.

See also:
- https://github.com/jupyterlite/jupyterlite/pull/1016
- https://github.com/jupyterlite/jupyterlite/pull/1355

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -49,7 +49,7 @@ dependencies:
   ### DOCS ENV ###
   # runtimes
   - nodejs >=20,<21
-  - python >=3.11,<3.12
+  - python >=3.12,<3.13
   # build
   - doit >=0.34,<1
   - hatch >=1.6.3,<2
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -165,7 +165,7 @@ jlpm serve
 
 To serve with Python's built-in
 [`http.server`](https://docs.python.org/3/library/http.server.html) module (requires
-Python 3.8+):
+Python 3.7+):
 
 ```bash
 jlpm serve:py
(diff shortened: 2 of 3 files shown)


Carreau (approved):

jtpio:
Thanks @EwoutH.

Looks like the windows and ReadTheDocs failures may be relevant:

```
Encountered problems while solving:
  - nothing provides _python_rc needed by python-3.13.0rc1-h3b4618f_0_cp313t
```

Carreau:
I think we should also unpin sphinx (which forces an older version of docutils).

jtpio:
> I think we should also unpin sphinx (which forces an older version of docutils).

It looks like we have the following pin at the moment:

https://github.com/jupyterlite/jupyterlite/blob/6b3ae93ff1d8875e9b18acffd964d29a45b97a2d/docs/environment.yml#L33

But the latest version of sphinx is now 8. Hopefully we can indeed bump the upper bound without too much trouble.

EwoutH:
Feel free to directly modify this PR, if that helps.

jtpio:
Thanks!

I opened a separate PR to look into this, but it looks like the `sphinx` update breaks the build on ReadTheDocs: https://github.com/jupyterlite/jupyterlite/pull/1493

jtpio:
I tried to make some progress with the sphinx update in https://github.com/jupyterlite/jupyterlite/pull/1493 but could not make it work yet. So stopping for now.

@EwoutH @Carreau if one of you would like to help update to the latest sphinx, so we can then merge this PR, that would be great :pray: Thanks!

jtpio (approved):
Thanks!

Issue #1475, opened 2024-09-06
Add a noscript tag to the page with "js must be enabled" ?


jtpio:
Add a noscript tag to the page with "js must be enabled" ?

_Originally posted by @Carreau in https://github.com/jupyterlite/jupyterlite/issues/1374#issuecomment-2063448971_

PR #1476, opened 2024-09-09
Add noscript tag

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1475

## User-facing changes

If for some reasons JavaScript is not enabled, show the following message:

image

None


bollwyvl:
No script is good, but maybe doesn't have to be as aggressively red.

Additionally, as it's too early to be translated, it could be configured with `LiteBuildConfig.no_script_html` and injected into the various built `index.html` files.

app/index.template.html
```diff
@@ -34,6 +34,12 @@
...
   </head>
   {{#unless (ispage name 'tree')}}
-  <body data-notebook="{{ name }}"></body>
+  <body data-notebook="{{ name }}">
+    <noscript>
+      <div style="text-align: center; padding: 20px;">
+          JupyterLite requires JavaScript to be enabled in your browser.
+      </div>
+  </noscript>
+  </body>
```
Carreau, line 43: Dumb question, but should this be outside of the `{{#unless (ispage name 'tree')}}` ? I'm guessing tree also does not work ?

Carreau (approved):

app/index.template.html
jtpio: Yeah maybe, I don't recall why we have this check (but could likely be found in the github history).

jtpio:
For reference, the UI tests does not seem to work at the moment, even though the video and the Playwright trace show the text when JavaScript is disabled.

This seems to be reproducible with a plain Playwright self-contained example, so I opened https://github.com/microsoft/playwright/issues/32542.

jtpio:
> This seems to be reproducible with a plain Playwright self-contained example, so I opened [microsoft/playwright#32542](https://github.com/microsoft/playwright/issues/32542).

This doesn't look like it's going to be fixed upstream any time soon, so disabling the UI test for now and we can track adding back separately.

PR #1477, opened 2024-09-09
Troubleshoot UI tests

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1479

## Code changes

- [x] Update snapshots
- [x] Install UI tests with `--no-deps` for now. Otherwise, installing `jupyterlite-pyodide-kernel==0.3.0a0` would downgrade `jupyterlite-core`:

<details>
<pre>
Attempting uninstall: jupyterlite-core
    Found existing installation: jupyterlite-core 0.4.1
    Uninstalling jupyterlite-core-0.4.1:
      Successfully uninstalled jupyterlite-core-0.4.1
</pre>
</details> 

## User-facing changes

None

## Backwards-incompatible changes

None

PR #1478, opened 2024-09-09
Use custom `artifact_name` and `report_name` in `playwright-update.yml`

jtpio:
## References

The workflow would otherwise give the following error:

```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

image

Likely because we test on both `firefox` and `chromium` and the artifact names were the same.

## Code changes

Use custom `artifact_name` and `report_name` in `playwright-update.yml` to include the name of the browser.

## User-facing changes

None

None


jtpio:
Merging to check if it helps with https://github.com/jupyterlite/jupyterlite/pull/1477

PR #1481, opened 2024-09-17
Add a partially functional local event manager

afshin:
This replaces the service manager's `MockEventManager` with a minimally functional `LocalEventManager`

## References
N/A

## Code changes
Adds a `LocalEventManager` class.

## User-facing changes
Extensions that rely on the event system may start working in JupyterLite.

N/A

## Example

This is a small plugin I wrote to test the event manager in a JupyterLite instance. I added this plugin to the build.

This plugin connects to the signal from the command registry that emits every time a command is executed. When a command execution comes through, this plugin emits a `test-event` in the event system.

**Note:** Because there is no back-end server where event schemas are registered, *all* event emissions are accepted and streamed by the `LocalEventManager`. We could decide in a future iteration to have client-side schema registration, etc.

```ts
  {
    id: 'event-manager-test-plugin',
    autoStart: true,
    activate: ({ commands, serviceManager: { events }}) => {
      const VERSION = '1';
      const SCHEMA = `https://example.com/schema/test-event/v${VERSION}`;
      commands.commandExecuted.connect((_, { args, id }) => {
        events.emit({
          schema_id: SCHEMA,
          data: { command: id, args: args as unknown as any },
          version: VERSION,
        });
      });
      void (async () => {
        for await (const event of events.stream) {
          if (event.schema_id === SCHEMA) {
            // eslint-disable-next-line no-console
            console.log('test-event emission', event);
          }
        }
      })();
    },
  }
```
Screenshot 2024-10-02 at 12 26 10

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -5,7 +5,7 @@ import { Event, ServerConnection, ServiceManager } from '@jupyterlab/services';
 
 import { Application, IPlugin } from '@lumino/application';
 
-import { Stream } from '@lumino/signaling';
+import { Signal, Stream } from '@lumino/signaling';
 
 import { WebSocket } from 'mock-socket';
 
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @afshin for looking into this!

Is there a minimal example out there making use of the event system, that we could use for testing?

afshin:
> Is there a minimal example out there making use of the event system, that we could use for testing?

Cheers! I'm currently working on a minimal plugin that will work both for JupyterLab and JupyterLite and when I have this, I'll update here as well.

jtpio (approved):
Thanks!

PR #1489, opened 2024-09-25
Ping service worker every 20s

trungleduc:
## References

Most of web browsers have timeout for service worker (30 seconds for both Chrome and Firefox), if there is no activity for the service worker, it will be stopped by the browser. This PR pings the SW every 20 seconds to keep it alive. This might fix the issue with file accessing in the kernel which fails randomly in voici.

_Firefox stops the SW after 30 seconds_

litesw

### DIFF
--- a/packages/server/src/service-manager.ts
+++ b/packages/server/src/service-manager.ts
@@ -6,7 +6,7 @@ import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 import { IServiceWorkerManager, WORKER_NAME } from './tokens';
 
 const VERSION = PageConfig.getOption('appVersion');
-const SW_PING_ENDPOINT = '/api/keep-sw-alive';
+const SW_PING_ENDPOINT = '/api/service-worker-heartbeat';
 export class ServiceWorkerManager implements IServiceWorkerManager {
   constructor(options?: IServiceWorkerManager.IOptions) {
     const workerUrl =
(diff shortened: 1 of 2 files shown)


trungleduc:
@martinRenou this PR fixes the timeout issue we discussed about

martinRenou (approved):
Neat thanks!

trungleduc:
in Firefox, the sw is stopped even in the active tab, maybe we changed something in JupyterLite? I don't see the `/api/contents` ping in the network activity tab.

https://github.com/user-attachments/assets/2ff610c9-18bd-42ae-8e7d-c6f528eb0256

It's not really an issue for jupyterlite because when you open the notebook, the worker is started again, the worker misses only the first request used to awake the sw, but it's a request not handled by the sw.  In the case of voici, this missed request might be the `api/drive` call. which leads to the file missing issue.

packages/server/src/service-manager.ts
```diff
@@ -6,7 +6,7 @@ import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 import { IServiceWorkerManager, WORKER_NAME } from './tokens';
 
 const VERSION = PageConfig.getOption('appVersion');
-
+const SW_PING_ENDPOINT = '/api/keep-sw-alive';
```
martinRenou, line 9: Just a nitpicky comment

```suggestion
const SW_PING_ENDPOINT = '/api/service-worker-heartbeat';
```

packages/server/src/service-worker.ts
```diff
@@ -50,9 +50,12 @@ async function onFetch(event: FetchEvent): Promise<void> {
   const { request } = event;
 
   const url = new URL(event.request.url);
+  if (url.pathname === '/api/keep-sw-alive') {
```
martinRenou, line 53:
```suggestion
  if (url.pathname === '/api/service-worker-heartbeat') {
```

jtpio (approved):
Thanks!

PR #1495, opened 2024-10-11
Service worker unregistering should be completed before registering new service worker

andeplane:
## References
In https://github.com/jupyterlite/jupyterlite/pull/1336 we introduced detection of a new service worker to avoid mismatch between old version when a new version is in place. However, I noticed when we updated our JupyterLite from 0.4.1 to 0.4.2 today that this happened in console:
```
New version, unregistering existing service workers.
JupyterLite ServiceWorker was already registered
All existing service workers have been unregistered.
```
aka it did not have time to unregister before it tried to load again. This is because the promise is not finished when it tries to register again. 

## Code changes
In this PR I made `unregisterOldServiceWorkers` async and run it with `await` for it to complete before it continues.

## User-facing changes
When a new version of JupyterLite is in place, the first session will no longer crash.

None

### DIFF
--- a/packages/server/src/service-manager.ts
+++ b/packages/server/src/service-manager.ts
@@ -49,9 +49,7 @@ export class ServiceWorkerManager implements IServiceWorkerManager {
       console.info('New version, unregistering existing service workers.');
       const registrations = await navigator.serviceWorker.getRegistrations();
 
-      for (const registration of registrations) {
-        await registration.unregister();
-      }
+      await Promise.all(registrations.map((registration) => registration.unregister()));
 
       // eslint-disable-next-line no-console
       console.info('All existing service workers have been unregistered.');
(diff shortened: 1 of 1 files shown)


packages/server/src/service-manager.ts
```diff
@@ -47,17 +47,14 @@ export class ServiceWorkerManager implements IServiceWorkerManager {
...
-          }
-        })
-        .then(() => {
-          // eslint-disable-next-line no-console
-          console.info('All existing service workers have been unregistered.');
-        });
+      const registrations = await navigator.serviceWorker.getRegistrations();
+
+      for (const registration of registrations) {
+        await registration.unregister();
```
jtpio, line 53: Maybe there is a way to unregister them all at once and wait with `Promise.all`?

andeplane: That's better, I agree. Fixed in 987b0d52c31e36c005d964addb65815dae371e4b

jtpio (approved):
Thanks!

PR #1509, opened 2024-10-24
Align kernel URL for server with client kernel URL

fcollonval:
## References

The mock-socket library check the requested ws URL matches one of the created server. Slight discrepancy may arise due to the server URL definition being built slightly differently than the one in the `KernelConnection`.

## Code changes

Align the construction of the kernel URL with the construction done in the client

https://github.com/jupyterlab/jupyterlab/blob/090fe7383422b59588f0337fe0b5ee0d74899204/packages/services/src/kernel/default.ts#L1248

## User-facing changes

None

None


fcollonval:
@MeeseeksDev tag bug

jtpio (approved):
Thanks!

Issue #1508, opened 2024-10-23
File size in the DriveFS is broken


juntyr:
## Description

In jupyterlite-core v0.4.3, file sizes in the DriveFS (e.g. via `os.stat`) are broken. Downgrading to v0.4.2 changes the breakage (less cases are affected) but doesn't fix it. Since #1497 seems to have touched this code, @katotetsuro might be able to find the issue.

## Reproduce

1. Go to https://jupyterlite.readthedocs.io/en/stable/_static/lab/index.html in a browser that supports the service worker so that DriveFS will be used (check the console)
2. Open a new notebook
3. Run the following code:

```python
import os

with open("a.txt", "wb") as f:
    f.write(b"abc" * 1000)

print(os.path.getsize("a.txt"))
# prints 0 (wrong)

with open("a.txt", "rb") as f:
    print(len(f.read()))
# prints 3000 (correct)
```

This issue is also related to #1472: opening the file in write mode again (which should reset the file to zero size) just overrides existing content but never truncates the length.

## Expected behavior

File sizes should be correct and truncating writes should work.

## Context
- JupyterLite version: 0.4.3

katotetsuro:
Hello,

I am currently reading about the implementation of DriveFS, so I am not very familiar with it yet.

However, I confirmed that by adding the size attribute set in this way and saving the size when creating the file, print(os.path.getsize("a.txt")) outputs 3000 at that point.

https://github.com/jupyterlite/jupyterlite/compare/main...katotetsuro:jupyterlite:feature/drivefs-size-attr

(Note: pyodide kernel also needs to be built using the modified jupyterlite-core)

The changes in contents.ts are still tentative, but I hope this might be of some help to you.

juntyr:
> Hello,
> 
> I am currently reading about the implementation of DriveFS, so I am not very familiar with it yet.
> 
> However, I confirmed that by adding the size attribute set in this way and saving the size when creating the file, print(os.path.getsize("a.txt")) outputs 3000 at that point.
> 
> [main...katotetsuro:jupyterlite:feature/drivefs-size-attr](https://github.com/jupyterlite/jupyterlite/compare/main...katotetsuro:jupyterlite:feature/drivefs-size-attr)
> 
> (Note: pyodide kernel also needs to be built using the modified jupyterlite-core)
> 
> The changes in contents.ts are still tentative, but I hope this might be of some help to you.

The code looks mostly good - thanks for having a look at it!

[...]

PR #1511, opened 2024-10-25
fix file size handling in drivefs

katotetsuro:
## References

add `size` property to drivefs
details in #1508

- comlink.worker.ts in jupyterlite-pyodide-kernel also needs to refer this build
- truncate logic for scenarios such as opening existing files in write mode is not included

## Code changes

- add `size` property to `IEmscriptenFSNode`
- set size property when creating new file
- ~~handling size property when creating/modifying content (WIP)~~

None

### DIFF
--- a/packages/contents/src/contents.ts
+++ b/packages/contents/src/contents.ts
@@ -421,6 +421,7 @@ export class Contents implements IContents {
     item = {
       ...item,
       ...options,
+      size: options.content?.length || 0,
       last_modified: modified,
     };
 
@@ -472,8 +473,6 @@ export class Contents implements IContents {
         };
       }
     }
-
-    item = { ...item, size: item.content.length };
 
     await (await this.storage).setItem(path, item);
     return item;
--- a/packages/contents/src/contents.ts
+++ b/packages/contents/src/contents.ts
@@ -473,6 +473,8 @@ export class Contents implements IContents {
       }
     }
 
+    item = { ...item, size: item.content.length };
+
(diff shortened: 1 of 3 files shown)


martinRenou:
Thanks a lot!

> handling size property when creating/modifying content (WIP)

Would you like us to wait for you to look into it?

We can also get this PR in first, and update the size properly in a separate PR.

juntyr:
I'm still slightly suspicious that the size is stored as a separate property, which can get out of sync. Would it be possible to remove it entirely and to always compute the size based on the content?

katotetsuro:
> always compute the size based on the content?

Is it something like setting size to content.length here ? It seems possible to me.
Since it looks like the content is always being read anyway just by calling getattr (sorry if I’m wrong), there doesn’t seem to be any downside in terms of processing time.

https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/drivecontents.ts#L199

katotetsuro:
> https://github.com/jupyterlite/jupyterlite/blob/main/packages/contents/src/drivecontents.ts#L199

since `{ content: false }` is used in the GET request here, it seems that returning the size based on the actual content.length is not immediately feasible.

@martinRenou 

> We can also get this PR in first, and update the size properly in a separate PR.

As you suggested, it would be helpful for me if the size adjustment could be handled in a separate PR. I also took a quick look at the truncate logic, but it didn’t seem straightforward to me now. Thanks!

PR #1521, opened 2024-11-06
Update download logic so that it relies on mimetype instead of solely content.type

martinRenou:
## References

Looking into fixing https://github.com/jupytercad/JupyterCAD/issues/544

## Code changes

The downloading logic was relying on the `content.type` (only `notebook`, `json` and default `file` was downloadable). 

This means that custom file types (jcad, jgis etc) would not be downloadable.

I change the logic to rely on the mimetype instead, so that other file types can be downloaded if they have a supported mimetype.  

## User-facing changes

None

### DIFF
--- a/ui-tests/test/contents.test.ts
+++ b/ui-tests/test/contents.test.ts
@@ -172,6 +172,7 @@ test.describe('Contents Tests', () => {
   });
 
   test('Download a custom file type', async ({ page }) => {
+    await refreshFilebrowser({ page });
     const path = await download({ page, path: 'test.customfile' });
     expect(path).toBeTruthy();
 
--- a/ui-tests/test/contents.test.ts
+++ b/ui-tests/test/contents.test.ts
@@ -173,6 +173,7 @@ test.describe('Contents Tests', () => {
 
   test('Download a custom file type', async ({ page }) => {
     await refreshFilebrowser({ page });
+    await page.filebrowser.open('test.customfile');
     const path = await download({ page, path: 'test.customfile' });
     expect(path).toBeTruthy();
 
--- a/ui-tests/test/contents.test.ts
+++ b/ui-tests/test/contents.test.ts
@@ -165,21 +165,21 @@ test.describe('Contents Tests', () => {
 
     expect(parsed.cells[0].source).toEqual(source);
   });
-});
 
-test('Download a custom file type', async ({ page }) => {
-  const path = await download({ page, path: 'test.customfile' });
-  expect(path).toBeTruthy();
+  test('Download a custom file type', async ({ page }) => {
+    const path = await download({ page, path: 'test.customfile' });
+    expect(path).toBeTruthy();
 
-  const content = await fs.readFile(path, { encoding: 'utf-8' });
-  const lines = content.split('\n');
+    const content = await fs.readFile(path, { encoding: 'utf-8' });
+    const lines = content.split('\n');
 
-  // check the file is correctly formatted
-  expect(lines.length).toBeGreaterThan(1);
+    // check the file is correctly formatted
+    expect(lines.length).toBeGreaterThan(1);
 
-  const parsed = JSON.parse(content);
+    const parsed = JSON.parse(content);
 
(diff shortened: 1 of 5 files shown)


jtpio:
Thanks @martinRenou!

There is a UI test for downloading notebooks here:

https://github.com/jupyterlite/jupyterlite/blob/ac544cc2a2836aa8cb74fb280b2488bf43157e99/ui-tests/test/contents.test.ts#L146

Wondering if it could be extended (or a new one be added), to test downloading other kind of files? Just to make sure we don't break that again later.

martinRenou:
That would mean installing an lab extension that provides a custom file type. Are you fine with this?

jtpio:
Ah, I though it would be enough to add the file type to the `jupyterlite_config.json` used for the UI tests, with an `extra_file_tyes`: https://jupyterlite.readthedocs.io/en/stable/howto/content/files.html#customizing-mime-types

And then add a test file in https://github.com/jupyterlite/jupyterlite/tree/main/ui-tests/contents

martinRenou:
Ah it's probably enough! I thought it was about using using `addFileType` on the document registry. I'll have a try.

martinRenou:
@jtpio do you have a clue why the visual regression test is never trying to download my file? 
The log from here https://github.com/jupyterlite/jupyterlite/pull/1521/files#diff-88062eb06c9ca4fb9080028ccdb5d0169b6053e050452fc40c33e767ad5ca0c6R192 is never showing up

jtpio:
On the RTD preview, the log seems to be displayed but the download is not triggered:

[jupyterlite-download-issue-2.webm](https://github.com/user-attachments/assets/62a9954c-9bc6-4917-a546-8df374273d27)

It also seems to be displayed in the UI test trace:

image

So maybe it is related to a change in the download logic? The original UI tests is also failing now:

image

martinRenou:
Thanks for trying @jtpio

> On the RTD preview, the log seems to be displayed but the download is not triggered:

Indeed I had an issue in the implementation for default "file" type.

It turns out the issue is due to the download test itself which does not trigger the download command properly. I added a comment in the test.

ui-tests/test/utils.ts
```diff
@@ -22,6 +22,7 @@ export async function download({
   path: string;
 }): Promise<string> {
   await page.evaluate(async (path: string) => {
+    // TODO Fix this. There is no such thing as passing a `path` here, the filebrowser will download any selected file
```
jtpio, line 25: :+1: 

Not sure exactly sure why we had to go this way to download, maybe it was a limitation with Galata.

ui-tests/test/contents.test.ts
```diff
@@ -27,6 +27,11 @@ test.use({
 
 test.describe('Contents Tests', () => {
   test.beforeEach(async ({ page }) => {
+    page.on('console', (message) => {
+      // eslint-disable-next-line no-console
+      console.log('CONSOLE MSG', message.text());
+    });
```
jtpio, line 33: Maybe we don't need this anymore if it was for debugging?

Otherwise such console logging should probably apply to all UI tests, not just for the contents?

jtpio: OK we can add it again globally later if needed.

ui-tests/test/contents.test.ts
```diff
@@ -27,6 +27,11 @@ test.use({
...
+
```
jtpio, line 34:
```suggestion
```

jtpio (approved):
Thanks!

jtpio:
Released in https://github.com/jupyterlite/jupyterlite/releases/tag/v0.4.4.

ui-tests/test/contents.test.ts
martinRenou: > apply to all UI tests

Being able to see `console.log`s in the test outputs is useful IMO, I always end up adding these three lines for debugging purpose.

I would be +1 for keeping that around.

jtpio: Let's track that in an issue?

PR #1524, opened 2024-11-10
Fix contents copy stripping part of path

juntyr:
## Code changes

Before, copying to a directory would strip the first character of its path, putting the copied file at an invalid path (so it cannot be discovered in the UI filebrowser).

## User-facing changes

None


juntyr:
ping @jtpio

Do you know what the original strip was for? If it had to do with leading `/`, perhaps we should use [`PathExt.removeSlash`](https://jupyterlab.readthedocs.io/en/3.2.x/api/modules/coreutils.pathext.html#removeslash) instead?

jtpio:
> Do you know what the original strip was for? If it had to do with leading `/`, perhaps we should use [`PathExt.removeSlash`](https://jupyterlab.readthedocs.io/en/3.2.x/api/modules/coreutils.pathext.html#removeslash) instead?

I don't recall the exact reason, it could be because of leading `/` showing up on copy while they shouldn't. But using `PathExt.removeSlash` could indeed be a better approach already.

juntyr:
I'm not quite sure what to make of these test failures

jtpio:
Looks like an issue with `pypy` maybe? But it's probably unrelated to this change.

Of not, since Python 3.8 is now EOL, we could check bumping to 3.9 in the test matrix in a separate PR, to see if that fixes the issue.

jtpio:
> Of not, since Python 3.8 is now EOL, we could check bumping to 3.9 in the test matrix in a separate PR, to see if that fixes the issue.

Made some CI maintenance in https://github.com/jupyterlite/jupyterlite/pull/1531, let's see if it helps.

jtpio (approved):
Thanks!

PR #1514, opened 2024-10-30
Update to JupyterLab 4.3 and Notebook 7.3

jtpio:
## References

Looking into https://github.com/jupyterlite/jupyterlite/issues/1482 to check if there is something major that may be breaking.

## Code changes

- [x] Bump `@jupyterlab`, `@jupyter-notebook` packages
- [x] Bump `jupyterlab` and `notebook` dependencies
- [x] Bump `jupyterlab` and `notebook` in conda environments
- [x] Add to the migration guide for 0.5.0
- [x] Fix UI tests

## User-facing changes

- JupyterLab changelog: https://jupyterlab.readthedocs.io/en/latest/getting_started/changelog.html
- Notebook changelog: https://jupyter-notebook.readthedocs.io/en/latest/changelog.html

## Backwards-incompatible changes

TBD

Issue #1027, opened 2023-03-29
Enable UI tests for uploading files


jtpio:
This is an issue to track testing the upload of files, which is currently commented here:

https://github.com/jupyterlite/jupyterlite/blob/b88c5d81b6bbb6a8925bd64799194cf79ddbf5d8/ui-tests/test/contents.test.ts#L21-L32

Which should check as part of #1019 (or after) whether the new Galata allows for more flexibility here.

This will also be a follow-up to https://github.com/jupyterlite/jupyterlite/pull/1024, so we can test uploading large files.

bollwyvl:
> new Galata

In the meantime: as the _Open from Path_ command exercises most of the same code paths, perhaps a test could:

- generate some big, trash data (e.g. a large CSV, a large notebook with outputs, etc.), calculate the SHA
  - use `PYTHONHASHSEED`, or `Faker.seed`, or something to get a stable file without checking it in
- serve it from the same server hosting the static assets
- use the command palette to download/open the file
  - visually verify that it has the expected number of rows
- open it in a kernel with synced contents
  - use simple operations to verify the hash, after it's gone through _several_ layers of indirection

PR #1530, opened 2024-12-03
Fix file size handling for the contents drive

### LINKED ISSUE #1472: DriveFS cannot overwrite-truncate existing file
DriveFS currently doesn't handle setting the `size` attribute here:

https://github.com/jupyterlite/jupyterlite/blob/82007cea732ac122f3fa2b12f3f694d16af5c575/packages/contents/src/drivefs.ts#L295

This makes `open("file", "w").close()` fail silently, i.e. an existing file is not truncated.

### LINKED ISSUE #1508: File size in the DriveFS is broken

### LINKED ISSUE #1529: .csv file size showing as 0B
Using current https://jupyterlite.github.io/demo/lab/index.html I generate lots of images correctly listed in file tab, however the .csv I generate which is not empty (numbers show when opening it in JupyterLite tab, or in Excel after download) is listed with a file size of 0B

juntyr:
## References

Fixes #1472
Fixes #1508 
Fixes #1529

Supersedes #1511

## Code changes

- Implements `setattr size` by getting, truncating or extending, and writing back the content file
- Fixes file save file size handling by being explicit when the stored content is a partial binary string and when it is decoded to its format

## User-facing changes

Bug fixes

None


juntyr:
Interestingly, at least #1508 doesn't reproduce on the latest main branch deployment but does on my own one, which uses jupyterlite v0.4.4 as well but uses Pyodide v27 (dev). So perhaps this is an issue that will only materialise with the next Pyodide release?

juntyr:
I did some quick testing with the ReadTheDocs version and uploading large PDFs and notebooks seems to still work

jtpio:
Thanks @juntyr for this! I'll try to have a look soon.

cc @martinRenou who may also be interested

jtpio:
@juntyr maybe it could be a good time to start looking into adding UI tests for file uploads?

We have been tracking that in https://github.com/jupyterlite/jupyterlite/issues/1027. Maybe this could have also helped catch such regressions, and help make sure it does not break again later.

juntyr:
If you want the test to be added in this PR, I'll need your help as I've never written a UI test before

jtpio:
> If you want the test to be added in this PR, I'll need your help as I've never written a UI test before

I guess it would be fine to do it separately too, since it may need some time. And also since this PR fixes a few issues it would be great to make it available to users sooner.

jtpio (approved):
Thanks!

juntyr:
> > If you want the test to be added in this PR, I'll need your help as I've never written a UI test before
> 
> I guess it would be fine to do it separately too, since it may need some time. And also since this PR fixes a few issues it would be great to make it available to users sooner.

What’s your current timeline for the next path release? If the fix could be released sometime next week it would be fantastic (I’m trying to publish a release of my project before the holidays)

jtpio:
I think we can make a new release now from `main`, before a `0.5.0` with https://github.com/jupyterlite/jupyterlite/pull/1514.

jtpio:
Released: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.4.5

Thanks @juntyr for your fixes! Feel free to report more issues if you find any during your testing.

juntyr:
> Released: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.4.5
> 
> Thanks @juntyr for your fixes! Feel free to report more issues if you find any during your testing.

Thanks @jtpio for the quick release! I think the pyodide-kernel needs a release as well (with the jupyterlite deps bumped) since its workers bundle the drive contents.

Since I can only test the combination of new jupyterlite + old pyodide-kernel right now, the file size issues appear fixed (this one seems to be in the "inner" layer that the kernel workers call but don't inline), but the file truncation logic doesn't run yet (this one seems to be in the "outer" layer that is inlined in the kernel workers)

juntyr:
I've opened https://github.com/jupyterlite/pyodide-kernel/pull/146

Issue #1052, opened 2023-04-13
Real Time Collaboration does not work consistently and reliably


gitskim:
## Description

I enabled the Real Time Collaboration (RTC) feature by following the documentation here (https://jupyterlite.readthedocs.io/en/latest/howto/configure/rtc.html) and implemented it here (https://github.com/gitskim/gitskim.github.io). My website (https://www.dspractice.com) uses the jupyterlite web pages hosted on my github mentioned above. I got the RTC feature to work on my github directly as well as my website. However, the feature doesn't work reliably. 

The cases where it didn't work:
1. Creating multiple rooms didn't work after one room worked. After we created our first room, we created a second one. In the second room, My collaborator and I  couldn't see what each other were typing and on the top right, it said "sharing with 0 peers"
2. Initially, we could use RTC in a newly created room, both from my github and from dspractice.com. But then, we tried to use the same room ten minutes later (after creating the second room in point 1. and it didn't work at all. We could not see what each other were typing and on the top right it said "sharing with 0 peers".

Here's how to reproduce the behavior:

## Reproduce

1. Enable the RTC feature following the instructions here (https://jupyterlite.readthedocs.io/en/latest/howto/configure/rtc.html)
2. Create a room number 12345 and add it at the end of the URL like here: https://www.dspractice.com/questions/binaryclass?content=problem&room=12345
3. The real time collaboration feature works.
4. Create another room like this: https://www.dspractice.com/questions/metrics?content=problem&room=97dd421f-3cf8-4ce5-a9a3-6de3a25e46ba
5. The real time feature doesn't work anymore in the second room created in Step 4. 
6. Go back to the first room that worked, which was created in Step 2.
7. The originally working first room doesn't work again either.  

## Expected behavior
The second room `97dd421f-3cf8-4ce5-a9a3-6de3a25e46ba` should work like the first room `12345` worked. And the first room `12345` should work again when we went back to it.
![Screen Shot 2023-04-12 at 7 11 34 PM](https://user-images.githubusercontent.com/8352119/231654117-e841512f-9821-49bc-b46c-cea213a1fc83.png)

## Context

- JupyterLite version: 0.1.0-beta.15
- Operating System and version: macOS Big Sur Version 11.7 
- Browser and version: Chrome Version 112.0.5615.49 (Official Build) (x86_64)

<details><summary>Browser Output</summary>
<pre>
recognized feature: 'web-share'.
[...]

PR #1536, opened 2024-12-13
Fix outdated RTC docs, clean up handling of the `collaborative` flag

jtpio:
## References

The current RTC docs are quite misleading and suggest RTC should (partially) work with the latest versions of JupyterLite, which has not been the case since 0.1.x.

This PR removes some outdated documentation about RTC since it was applying to the 0.1.x versions, which is not maintained anymore.

Fixes https://github.com/jupyterlite/jupyterlite/issues/1052

## Code changes

- [x] Remove outdated docs
- [x] Point to https://github.com/davidbrochart/jupyter-shared-drive as a new experiment for enabling RTC in lite
- [x] Remove `collaborative` and `fullWebRtcSignalingUrls` from the schema for now
- [x] Clean up handling of the `room` when sharing links to files

## User-facing changes

Clarify the state of RTC in lite.

None


jtpio:
cc @bollwyvl if you have some input on the removal of `fullWebRtcSignalingUrls` and `collaborative` from the schema.

Since https://github.com/jupyterlite/jupyterlab-webrtc-docprovider is currently not working with the latest versions of JupyterLite, it would be less confusing to users if such RTC related options are removed from the docs. We can consider adding them again later, or let the extensions (https://github.com/davidbrochart/jupyter-shared-drive or something else) handle that independently.

bollwyvl:
Yep, it's fine and indeed, we should have done so a _while_ back. As it isn't validated at runtime, it's not going to cause any breakage, so wouldn't bother updating the version.

Another slightly less invasive path would be to mark it `deprecated`, but the docs tool we use doesn't support doing anything special for that.

jtpio:
Cool, thanks!

PR #1537, opened 2024-12-17
Drop Python 3.8

jtpio:
## References

Python 3.8 is EOL.

## Code changes

- [x] Require Python 3.9 or above
- [x] Bump to `ruff==0.8.3`
- [x] Run `ruff`
- [x] Add to migration guide


bollwyvl (approved):
:tada: thanks for patching this up. Many notebooks still have the old metadata... might just want to do that before 0.5.0, but getting this will help keep moving things along.

Issue #1549, opened 2025-01-02
Jupyter lite is not saving cell ids


martenrichter:
## Description

If you save a notebook with jupyter lite (e.g. use current try jupyter), the exported cells are missing the id field.

## Reproduce

Create a notebook on try jupyter, add one cell, save. Download the file, and view it in a text editor and you see the missing id file.
E.g.
```
{
  "metadata": {
    "kernelspec": {
      "name": "python",
      "display_name": "Python (Pyodide)",
      "language": "python"
    },
    "language_info": {
      "codemirror_mode": {
        "name": "python",
        "version": 3
      },
      "file_extension": ".py",
      "mimetype": "text/x-python",
      "name": "python",
      "nbconvert_exporter": "python",
      "pygments_lexer": "ipython3",
      "version": "3.8"
    }
  },
  "nbformat_minor": 4,
  "nbformat": 4,
  "cells": [
    {
      "cell_type": "code",
      "source": "%matplotlib inline\n%pip install -q ipywidgets\nfrom ipywidgets import interactive\nimport matplotlib.pyplot as plt\nimport numpy as np\n\ndef f(m, b):\n    plt.figure(2)\n    x = np.linspace(-10, 10, num=1000)\n    plt.plot(x, m * x + b)\n    plt.ylim(-5, 5)\n    plt.show()\n\ninteractive_plot = interactive(f, m=(-2.0, 2.0), b=(-3, 3, 0.5))\noutput = interactive_plot.children[-1]\ninteractive_plot",
      "metadata": {
        "trusted": true
      },
      "outputs": [
        {
          "execution_count": 1,
          "output_type": "execute_result",
          "data": {
            "text/plain": "interactive(children=(FloatSlider(value=0.0, description='m', max=2.0, min=-2.0), FloatSlider(value=0.0, descr…",
            "application/vnd.jupyter.widget-view+json": {
              "version_major": 2,
              "version_minor": 0,
              "model_id": "54a206f57a914537a6db3bda5a8a4cba"
            }
          },
          "metadata": {}
        }
      ],
      "execution_count": 1
    },
    {
      "cell_type": "code",
      "source": "",
      "metadata": {
        "trusted": true
      },
      "outputs": [],
      "execution_count": null
    },
    {
      "cell_type": "markdown",
      "source": "* Hallo",
      "metadata": {}
    },
    {
      "cell_type": "code",
      "source": "",
      "metadata": {
        "trusted": true
      },
      "outputs": [],
[...]

PR #1550, opened 2025-01-02
Increase nbformat_minor to 5 in contents package

martenrichter:
The default notebook contents were set in the Jupyter Lite contents package with nbformat_minor 4. The current Jupyter lab starts with nbformat_minor 5. This patch also changes the default in Jupyter Lite to 5. ng_format_minor 5 enables features such as cell preservation in Jupyter Lite.

## References

https://github.com/jupyterlite/jupyterlite/issues/1549

## Code changes

Just increase the version number of the default template.

## User-facing changes

User will get notebooks in the newer format.

None.

### DIFF
--- a/packages/contents/src/contents.ts
+++ b/packages/contents/src/contents.ts
@@ -872,7 +872,7 @@ namespace Private {
     metadata: {
       orig_nbformat: 4,
     },
-    nbformat_minor: 4,
+    nbformat_minor: 5,
     nbformat: 4,
     cells: [],
   };


martenrichter:
I hope that the old version was not kept intentionally.

martenrichter:
Interesting the "try lite now" link does not work.

jtpio:
Thanks @martenrichter!

> Interesting the "try lite now" link does not work. (404)

Yes the link is posted when the PR is opened, but it takes a few minutes for the build to complete.

> I hope that the old version was not kept intentionally.

I think this hasn't changed since the initial implementation, so it's probably because it's what Jupyter Server was returning by default at the time.

Looks like it should indeed be safe to bump the number. Maybe we can include that in the upcoming 0.5.0 and not backport to 0.4.x just in case.

martenrichter:
> > I hope that the old version was not kept intentionally.
> 
> I think this hasn't changed since the initial implementation, so it's probably because it's what Jupyter Server was returning by default at the time.
> 
> Looks like it should indeed be safe to bump the number. Maybe we can include that in the upcoming 0.5.0 and not backport to 0.4.x just in case.

Well, I can wait, as I have copied it in my ´code anyway. But I think it is also good if jupyter lab and lite are kept in sync.
When searching through the repos, I have only found it at code that dropped the id and in various examples and tests.
(Was the id the only change between the versions?)

Btw. is there a way for a user to upgrade the version via the UI?

packages/contents/src/contents.ts
```diff
@@ -872,7 +872,7 @@ namespace Private {
     metadata: {
       orig_nbformat: 4,
     },
-    nbformat_minor: 4,
+    nbformat_minor: 5,
```
bollwyvl, line 875: Hmmm... we almost certainly depend on the `@jupyterlab/nbformat` package (if indirectly), so we could replace the  magic numbers with, at least, _someone else's_ magic numbers, a la https://github.com/jupyterlab/jupyterlab/blob/v4.4.0a1/packages/notebook/src/model.ts#L113

```ts
import * as nbformat from '@jupyterlab/nbformat';

// ...
          nbformat: nbformat.MAJOR_VERSION,
          nbformat_minor: nbformat.MINOR_VERSION,
```

bollwyvl: gah, never mind: that's _still_ going to be `4`:

https://github.com/jupyterlab/jupyterlab/blob/main/packages/nbformat/src/index.ts#L22

bollwyvl: As we're thinking this goes into `0.5.0`, I think we should probably either:
- hoist the magic numbers up to e.g. `tokens.ts`
- find some other way to verify it against what a compatible install of `jupyter_server` provides

martenrichter:
May be it would be a good idea to just add a comment to:
https://github.com/jupyter/nbformat/blob/ba2c6f53afbeb18f6fad0c9b9f67def34e757cb4/nbformat/v4/nbbase.py#L23
that also jupyter lite contents object should be updated?

jtpio:
> May be it would be a good idea to just add a comment to: https://github.com/jupyter/nbformat/blob/ba2c6f53afbeb18f6fad0c9b9f67def34e757cb4/nbformat/v4/nbbase.py#L23 that also jupyter lite contents object should be updated?

Not sure about this. Maybe we can open a new issue here on the JupyterLite repo instead if we want to improve how we manage these numbers.

jtpio (approved):
Thanks!

PR #1558, opened 2025-01-14
Fix typo in CLI addons doc

lesteve:
Note: this does not really matter since `manager` variable is not used in the code.


jtpio (approved):
Thanks!

PR #1560, opened 2025-01-15
Troubleshoot UI tests

jtpio:
## References

Looking into recent failures with the chromium UI tests.

## User-facing changes

None

None


jtpio:
Looks like the minor snapshot updates help. Not sure what change led to the mismatch (unless it comes from a dependency). Will keep an eye on the next CI runs.

Issue #1540, opened 2024-12-18
Update to JupyterLab 4.4 and Notebook 7.4


jtpio:
It will be great to update early and more often to the new upstream versions this time.

At the moment there is a JupyterLab alpha:
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.4.0a0
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.4.0a1

No Notebook pre-release for now.

PR #1561, opened 2025-01-16
Update to JupyterLab 4.4 and Notebook 7.4

jtpio:
## References

First step towards https://github.com/jupyterlite/jupyterlite/issues/1540

## Code changes

Regular dependency updates.

## User-facing changes

Some from upstream releases.

Currently:

- Settings import / export from the settings editor

None


jtpio:
CI is passing, and it seems to be running fine on RTD.

Even though 0.5.0 was released recently, maybe we can just go ahead and release a first `0.6.0a0`. This would allow keeping up the upstream releases, and not fall behind too quickly.

jtpio:
OK let's get it in a pre-release.

jtpio:
Ah looks like there may be an issue with the "Notebook" toolbar button to open in Jupyter Notebook. Will have a look.

jtpio:
> Ah looks like there may be an issue with the "Notebook" toolbar button to open in Jupyter Notebook. Will have a look.

Looking fine now:


https://github.com/user-attachments/assets/b55cb12b-38bd-4e0d-9071-93aa39a3fbdd

Issue #1572, opened 2025-02-04
Expose and document more REPL options


jtpio:
### Problem

As a follow-up to #1570, it would be nice to expose more of the new options so they can be enabled via query string parameters too:

- `clearCellsOnExecute`
- `clearCodeContentOnExecute`
- `hideCodeInput`
- `promptCellPosition`
- `showBanner`

### Proposed Solution

This could be done by rebooting https://github.com/jupyterlite/jupyterlite/pull/525.

And the new options should be added to the following page for visibility (with screenshots?): https://jupyterlite.readthedocs.io/en/stable/quickstart/embed-repl.html

### Additional context

- #148 
- #1570 
- https://github.com/jupyterlab/jupyterlab/pull/13837

Issue #503, opened 2022-02-15
Provide `IURLParamManager`


jtpio:
### Problem

We should consider `provide`ing this as a  `IURLParamManager` which can have additional params added that modify how things work at different lifecycle hooks.

### Proposed Solution

For example:

```ts
urlParamManager.registerParam('code', {
  onAppStarted: async () => console.log('whatever')
  onWidgetAdded: (widget) => await ....
});
```

### Additional context

Originally posted by @bollwyvl in https://github.com/jupyterlite/jupyterlite/pull/498#discussion_r805349143

bollwyvl:
A further idea is capturing the params as a JSON schema, from which we can generate documentation and/or a link builder app, which is particularly important for escaping the `code`, etc. so `register` might be like:

```
urlParamManager.registerParam('code', {
  schema: {
    title: 'Code',
    description: 'Code to pre-populate',
    type: 'string'
  },
  onAppStarted: async () => console.log('whatever')
  onWidgetAdded: (widget) => await ....
});
```

PR #1573, opened 2025-02-04
Expose more REPL options via query string parameters

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1572.

Leaving https://github.com/jupyterlite/jupyterlite/issues/503 for another PR for now, as this could be implemented separately.

## Code changes

- [x] Expose more REPL options so they can be controlled via query string parameters.
- [x] Add documentation

## User-facing changes

Users can tweak more options of their embedded REPL via a single URL directly, without having to do a full deploy with a custom `overrides.json`.

Example with https://jupyterlite--1573.org.readthedocs.build/en/1573/_static/repl/index.html?toolbar=1&kernel=python&promptCellPosition=left&clearCellsOnExecute=1&hideCodeInput=1&clearCodeContentOnExecute=0

image

Handling of the `showBanner` option may not yet work as expected due to the current handling of the kernel banner: https://github.com/jupyterlab/jupyterlab/pull/17322

None

### DIFF
--- a/docs/quickstart/embed-repl.md
+++ b/docs/quickstart/embed-repl.md
@@ -98,3 +98,68 @@ pip install jupyterlab-gt-coar-theme
 
 See the [how-to guides](../howto/index.md) for more details on how to customize the
 environment and add more themes and extensions.
+
+### Clear cells on execute
+
+To automatically clear the previously executed cells on execute, set the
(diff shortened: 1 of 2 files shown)


jtpio:
Merging in its current state so we can iterate on it more easily.

PR #1577, opened 2025-02-16
[DOC] Update Python version in vercel/netlify docs

yogabonito:
Before this commit, Python 3.7 was recommended, but it is no longer supported.


jtpio (approved):
Thanks!

PR #1583, opened 2025-02-20
Fix --user installed labextensions not found

2xB:
## References

Fixes https://github.com/jupyterlite/pyodide-kernel/issues/125 and https://github.com/jupyterlite/demo/issues/131 .

## Code changes

Previously, when - without using a venv - one installed jupyterlite and lab extensions like the Pyodide kernel as user without admin rights, the extensions were not found by jupyterlite. This is fixed by respecting paths in which labextensions can be installed when installing into e.g. the local user folder.

## User-facing changes

User-installed labextensions outside venvs are now used by jupyterlite.

### DIFF
--- a/py/jupyterlite-core/jupyterlite_core/addons/federated_extensions.py
+++ b/py/jupyterlite-core/jupyterlite_core/addons/federated_extensions.py
@@ -2,7 +2,7 @@
 
 import json
 import re
-import sys
+import site
 import urllib.parse
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


2xB:
Of course one can also argue for the approach that not being in a virtual environment should cause an error instead. Then I believe this should also be documented, but an error providing information to users would also be a good start.

RMattB1:
> Of course one can also argue for the approach that not being in a virtual environment should cause an error instead. 

This does does not say anything about being in a virtual environment that I can tell: 
https://jupyterlite.readthedocs.io/en/latest/quickstart/standalone.html
Is there another doc somewhere which describes the way it _should_ be done, in a virtual environment?
Thanks!

RMattB1:
Other symptom descriptions, to help people find this fix:
kernel status idle
"select kernel" "no kernel"

Issue #279, opened 2021-07-21
Improve the tooling around the drawio overview diagram


jtpio:
In https://github.com/jupyterlite/jupyterlite/pull/278 we add an overview diagram to the documentation, for reference and so it can more easily be found and iterated on.

---

Perhaps as follow-on, but... 
- put the image in `examples` so it's something people can play with... 
  - assuming `jupyter-drawio` can handle SVG... 
  - otherwise the XML there, and `.svg` in `docs`
    - and a check that the svg is always newer than the xml, gah
- can we put under any sort of test? 
  - from the codebase, we pull out number/names of:
    - apps
    - kernels
    - storage drivers
    - CDN providers
  - with e.g. `lxml` (or even standard lib) we pull out shapes/strings that should match the above
  - test fail if they do not agree
    - or even have them as notebooks that would show what's out-of-date (but not strictly _fail_)

i did a bit about this on [ipydrawio](https://ipydrawio--52.org.readthedocs.build/en/53/tutorials/designing-jupyter-extensions/index.html#testing)... that's specifically with the `.dio.svg` format, but would probably work with the straight-up `mxModel` dialect.

_Originally posted by @bollwyvl in https://github.com/jupyterlite/jupyterlite/issues/278#issuecomment-883448265_

bollwyvl:
The extension serving this seems to be bitrotting vs newer labs... the launcher icon broke a while ago, and it leaves random artifacts on the screen... i'm kinda tempted to remove it and bring in ipydrawio...

jtpio:
> i'm kinda tempted to remove it and bring in ipydrawio..

I'm not sure there has been any work and maintenance going into `jupyterlab-drawio` recently. So we could switch of `ipydrawio` :+1:

PR #1590, opened 2025-02-26
Register server plugins as `ServiceManagerPlugin`, drop `JupyterLiteServer`

jtpio:
## References

Closes https://github.com/jupyterlite/jupyterlite/pull/1284
Closes https://github.com/jupyterlite/jupyterlite/pull/1285
Closes https://github.com/jupyterlite/jupyterlite/issues/279

Leverage https://github.com/jupyterlab/jupyterlab/pull/16794 in JupyterLite to register what used to be "server" plugins as JupyterLab `ServiceManagerPlugin`.

## Code changes

- [x] Register all plugins on the same app
- [x] Provide JupyterLite specific plugins as `ServiceManagerPlugin`s in a new `@jupyterlite/services-extension`
- [x] Remove `JupyterLiteServer`, `JupyterLiteServerPlugin`, `Router`
- [x] Provide custom `Kernel.IManager`, `KernelConnection`, Session.IManager`, `SessionManager` -> check upstream if `@jupyterlab/services` could allow for more flexbility
- [x] Provide `BrowserStorageDrive` as the `IDefaultDrive`
- [x] Custom `fetch` is still provided to log messages to the dev tools console, so it's easier to detect core plugins and extensions that may need to be adapted
- [x] Add to migration guide

## User-facing changes

None

Some to be expected for downstream applications that were reusing some `@jupyterlite` packages.

Kernels should normally still work as before without modifications, although they can still update to use the new `ServiceManagerPlugin`.

These backwards-incompatible changes should be listed in the migration guide.


jtpio:
We should be able to remove the custom kernel manager and kernel connection when https://github.com/jupyterlab/jupyterlab/pull/17395 is merged and released.

In the meantime, maybe it could be worth stopping here for now so we can make a new release and try these changes more thoroughly.

martinRenou (approved):
It's awesome to see this happening! Thanks for the PR!

Concerning the jupyterlite-xeus package, it means removing the split between labextension/liteextension which will reduce the codebase complexity quite a bit!

jtpio:
Thanks for the reviews!

OK getting this one in and following up with a new alpha release right after.

There may still be some things to will need fixing, but we can address them in separate PRs.

martinRenou:
@jtpio Thanks a lot for the release! Trying it out now.

It looks like `@jupyterlite/services-extension` is not published on NPM yet, would you mind publishing it?

martinRenou:
Actually it is, I wasn't finding it from the UI but it is installable.

jtpio:
It shows up on npmjs.com: https://www.npmjs.com/package/@jupyterlite/services-extension

image

juntyr:
@jtpio I have two server extensions where I'm currently unsure for how to migrate them, even after reading the migration guide:

- https://github.com/juntyr/a-jupyterlite-query-config updates the PageConfig for other extensions (e.g. the pyodide kernel), before they load and have a chance to look at the config. The ordering was achieved by having no dependencies and a lexicographically small name
- https://github.com/juntyr/a-jupyterlite-session (the cousin of https://github.com/juntyr/a-jupyterlab-session though the latter is already a frontend extension) depends on the content manager (internals) to create a new (session) folder in the contents (based on a URL parameter) before any other extension can look at it. Its cousin frontend extension then cd's the frontend to this directory, ensuring that e.g. the litegitpuller creates its files in the session folder

Would you perhaps be able to give me some hints? Thanks for your help!

jtpio:
@juntyr I guess the easiest would be to try and make sure they still behave as expected.

For the first one I wouldn't expect any difference. When it comes to plugin loading order, the `IServiceManager` is resolved first, and all the plugins it depends on:

https://github.com/jupyterlite/jupyterlite/blob/37c93371964c85156556e4196a8745bbd39fd0cd/app/index.template.js#L163

Then all the other plugins, including kernels, should be activated in the second phase:

https://github.com/jupyterlite/jupyterlite/blob/37c93371964c85156556e4196a8745bbd39fd0cd/app/index.template.js#L182-L183

jtpio:
@juntyr if you find any issue or have other questions, please don't hesitate to open a new issue, so we can have a look before the final 0.6.0 release (expected soon after JupyterLab 4.4.0). Thanks!

juntyr:
@jtpio Thanks! I managed to port https://github.com/juntyr/a-jupyterlite-query-config without issues, but I'm a bit stuck with https://github.com/juntyr/a-jupyterlite-session.

Previously, I used

```js
// Share the session folder location with the `a-jupyterlab-session`
//  frontend extension
app.router.get(
  API_ENDPOINT,
  async (_req: Router.IRequest) => new Response(session)
);
```

to register an API endpoint using the Router to share a string with the cousin extension. Now that the Router has been removed, what would be a good way to share this information with the other plugin?

Here's my current upgrade diff https://github.com/juntyr/a-jupyterlite-session/compare/main...upgrade

jtpio:
@juntyr to share information between plugins, you could follow the provider / consumer approach used by many plugins in JupyterLab and extensions. With one plugin providing a service and exposing a token, the other one requiring the token to get access to the service. And use the provided service to share information.

There is some information about this in the docs: https://jupyterlab.readthedocs.io/en/latest/extension/extension_dev.html#plugins-interacting-with-each-other

packages/contents/src/drive.ts
```diff
@@ -298,7 +372,7 @@ export class Contents implements IContents {
     const model = (item || serverItem) as IModel | null;
 
     if (!model) {
-      return null;
+      throw Error(`Could not find content with path ${path}`);
```
juntyr, line 375: Throwing an error here conflicts with the copy implementation in line 256:

```ts
// TODO: better handle naming collisions with existing files
while (await this.get(`${toDir}${name}`, { content: true })) {
  const ext = PathExt.extname(name);
  const base = name.replace(ext, '');
  name = `${base} (copy)${ext}`;
}
```
which expects the previous behavior of returning null.

juntyr: @jtpio Thanks for the token suggestion! I managed to port all extensions but testing them triggered this bug.

juntyr: Since this is in the contents implementation, any fix will require a JupyterLite and Pyodide kernel release since the latter inlines the implementation from them former

jtpio: oh, thanks for reporting, I'll have a look!

jtpio: > Pyodide kernel release since the latter inlines the implementation from them former

Not sure about the Pyodide update + release though, since this logic is now encapsulated in a drive?

juntyr: I hadn't thought of that - It would be fantastic if fixes now work without the double publish!

jtpio: Do you have an example to share to help reproduce the issue?

juntyr: My JupyterLite deployment is on the latest alpha, so you can open the dev console for https://lab.climet.eu/main

In the a-jupyterlite-session extension, I copy a file here:
https://github.com/juntyr/a-jupyterlite-session/blob/3798ca350d09c7dd14d547b9205add2dd449cf67/src/index.ts#L92-L98

In the dev console, you can see that this fails:

```
Failed to copy the requirements.txt file to the new session: Error: Could not find content with path .sessions/03-21-2025-15-56-57-1087a52a/requirements.txt
```

The error "Could not find content" is only emitted in the get method, which is called inside copy to check if the path to copy to already exists.

I'm also observing the issue with litegitpuller, though I'm not quire sure yet why since it doesn't use copy and wraps its get calls with catch. Maybe some other function in the drive now uses get and still expects the failure case to return null?

jtpio: OK thanks, I've opened https://github.com/jupyterlite/jupyterlite/issues/1599 to track this, please feel free to add more comments there if you would like to.

Issue #1047, opened 2023-04-10
Not possible to use non-standard custom storage drivers (doc)


larsmoa:
### Problem

Not sure if this is considered a documentation bug/improvement or an actual bug, but documentation on storage drivers is a bit fuzzy. According to the [Local Storage Drivers](https://jupyterlite.readthedocs.io/en/latest/howto/configure/storage.html#local-storage-drivers) its possible to use other storage drivers than the standard drivers that "may added by third-party extensions".

However, after trying it seems to me that this is not possible as storage is initialized prior to extensions and that custom storage would require something like #1041 (?).

### Suggested Improvement

Clarify limitations on custom storage drivers and/or provide example of how non-standard drivers can be used.

jtpio:
Thanks @larsmoa.

> its possible to use other storage drivers than the standard drivers that "may added by third-party extensions".

I think this was mentioned just to leave the door open to third-party extensions that may contribute additional storage drives. Not sure there is any at the moment that does this.

Agree the docs could be improved to make it less confusing :+1:

PR #1591, opened 2025-03-02
Fix memoryStorageDriver in jupyterlite.schema.v0.json

2xB:
## Code changes

Without knowing the details, while researching for https://github.com/jupyterlite/jupyterlite/issues/1047 (which I would still be interested in), I stumbled across this:

https://github.com/jupyterlite/jupyterlite/blob/db7eec412c417581b5155055c758f6838fa209ae/app/jupyterlite.schema.v0.json#L291-L302

Comparing to [documentation](https://jupyterlite.readthedocs.io/en/latest/howto/configure/storage.html#local-storage-drivers), the second `localStorageWrapper` string looks like it actually should be `memoryStorageDriver`. Therefore suggesting that in this PR just in case this is actually a bug.

### DIFF
--- a/app/jupyterlite.schema.v0.json
+++ b/app/jupyterlite.schema.v0.json
@@ -298,7 +298,7 @@
           "title": "memory",
           "description": "use the memory-based, volatile store: must be enabled with enableMemoryStorage",
           "type": "string",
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

PR #1596, opened 2025-03-19
Trigger signal for spec refresh

martinRenou:
## References

Fix regression seen in https://github.com/jupyterlite/xeus/pull/204


jtpio (approved):
Thanks for catching this!

Make sense, as otherwise the `specsChanged` was never emitted.

jtpio:
@martinRenou anything else that needs to be fixed? Or should we make a new 0.6.0 alpha release with this change?

martinRenou:
> should we make a new 0.6.0 alpha release with this change?

That'd be great! It's the only issue I caught with the alpha release

jtpio:
Released: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.6.0a4

PR #1597, opened 2025-03-20
Update to JupyterLab 4.4.0b2 and Notebook 7.4.0b2

jtpio:
## References

https://github.com/jupyterlite/jupyterlite/issues/1540

Update to:

- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.4.0b2
- https://github.com/jupyter/notebook/releases/tag/v7.4.0b2

## Code changes

- Bump lab and notebook
- Remove kernel manager and connection as a follow-up to https://github.com/jupyterlite/jupyterlite/pull/1590 and https://github.com/jupyterlab/jupyterlab/pull/17395
- Fix handling of translations, as a follow-up to https://github.com/jupyterlab/jupyterlab/pull/17414

PR #1598, opened 2025-03-21
Remove custom `IServerSettings` plugin

jtpio:
## References

Follow-up to:

- https://github.com/jupyterlite/jupyterlite/pull/1590
- https://github.com/jupyterlite/jupyterlite/pull/1597

Now that JupyterLite instantiates and provides its own kernel manager, replacing the default `IServerSettings` should not be needed anymore.

Instead we can just provide the `WebSocket` from `mock-socket` when instantiating the `KernelManager`.

Removing the custom `IServerSettings` should help avoid interfering with the built-in `WebSocket` and `fetch` too much.

## Code changes

- Remove `IServerSettings`, use the core plugin instead
- Provide `WebSocket` to the `serverSettings` of the `KernelManager`

## User-facing changes

None

None

### DIFF
--- a/app/consoles/package.json
+++ b/app/consoles/package.json
@@ -325,7 +325,6 @@
       "@jupyterlab/services-extension:kernel-manager",
       "@jupyterlab/services-extension:kernel-spec-manager",
       "@jupyterlab/services-extension:nbconvert-manager",
-      "@jupyterlab/services-extension:server-settings",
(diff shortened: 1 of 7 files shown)


jtpio:
Getting this one in and making a new pre-release, to help with https://github.com/jupyterlite/terminal/pull/49.

PR #1602, opened 2025-03-25
Refresh specs when new JupyterLite kernels are added

jtpio:
## References

This should help make the lite kernels visible in the UI immediately.

And remove the need for a manual spec refresh, as currently done in `jupyterlite-xeus`: https://github.com/jupyterlite/xeus/blob/1e9372e9ff6801ca1679cf4b259d9056f419c151/packages/xeus-extension/src/index.ts#L129

## Code changes

Automatically refresh the kernel manager specs when new lite specs are added.

## User-facing changes

Less likely to see missing kernels in the UI.

None


martinRenou (approved):
Thanks!!

PR #1606, opened 2025-03-27
Handle Drive.get exceptions for missing content

### LINKED ISSUE #1599: Check handling of missing content and copy in the default drive
My JupyterLite deployment is on the latest alpha, so you can open the dev console for https://lab.climet.eu/main

In the a-jupyterlite-session extension, I copy a file here:
https://github.com/juntyr/a-jupyterlite-session/blob/3798ca350d09c7dd14d547b9205add2dd449cf67/src/index.ts#L92-L98

In the dev console, you can see that this fails:

```
Failed to copy the requirements.txt file to the new session: Error: Could not find content with path .sessions/03-21-2025-15-56-57-1087a52a/requirements.txt
```

The error "Could not find content" is only emitted in the get method, which is called inside copy to check if the path to copy to already exists.

I'm also observing the issue with litegitpuller, though I'm not quire sure yet why since it doesn't use copy and wraps its get calls with catch. Maybe some other function in the drive now uses get and still expects the failure case to return null?

_Originally posted by @juntyr in https://github.com/jupyterlite/jupyterlite/pull/1590#discussion_r2007644056_

juntyr:
## References

Fixes #1599 

## Code changes

Transform exceptions into bools or nulls as required by the context

## User-facing changes

None

None

### DIFF
--- a/packages/contents/src/drive.ts
+++ b/packages/contents/src/drive.ts
@@ -204,7 +204,7 @@ export class BrowserStorageDrive implements Contents.IDrive {
     let dirname = PathExt.dirname(path);
     const basename = PathExt.basename(path);
     const extname = PathExt.extname(path);
-    const item = await this.get(dirname);
+    const item = await this.get(dirname).catch(() => null);
 
     // handle the case of "Save As", where the path points to the new file
     // to create, e.g. subfolder/example-copy.ipynb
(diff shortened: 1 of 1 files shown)


juntyr:
@jtpio Could you perhaps have a look at this PR today and if it looks good publish a fresh pre-release? The breaking of any kind of drive interactions (e.g. file uploads from the normal JupyterLab upload button) is making this more time sensitive for me

jtpio (approved):
Thanks @juntyr!

Right that makes sense and should limit the number of differences compared to the previous version of the contents manager.

jtpio:
Released: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.6.0a6

juntyr:
> Released: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.6.0a6

Thanks, everything works now <3

jtpio:
Thanks for the fix and testing!

Issue #1601, opened 2025-03-24
New repo shows file from a different repo


Guillaume-Garrigos:
## Description

In a nutshell, after cloning the jupyterlite demo repo, I obtain an instance of jupyterlite containing a file which shouldn't be there. Somehow my browser mixed files with another jupyterlite repo I created earlier.

## Reproduce

I am not sure this specific problem can be reproduced, but here is what I did:

1. I created a first repo to play with jupyterlite, trying to adapt a practical session for my students. Here is this first [repo](https://github.com/Guillaume-Garrigos/jlite-test) and the associated [website](https://guillaume-garrigos.com/jlite-test/lab/index.html). All that matter is that the `content` folder contains a certain notebook called `sujet.ipynb`
2. I encounter various problems with loading files and scripts, so I decide to investigate by creating a separate and new repo, which would serve as a minimal working example for my problems.
3. I create such MWE by using again the jupyterlite demo repo. Here is this [second repo](https://github.com/Guillaume-Garrigos/jlite_debug_local_import) and the corresponding [website](https://guillaume-garrigos.com/jlite_debug_local_import/lab/index.html). It is an exact copy with the very same files (except the README.md which I modified).
4. When opening the [launched website](https://guillaume-garrigos.com/jlite_debug_local_import/lab/index.html), I see on the left tab that there is the `sujet.ipynb` file here! I never put it there! I can open it and modify it.
5. Opening this link in a private tab, this spurious file is no longer there.

I note that only this file got duplicated across projects. For instance the date files in `Guillaume-Garrigos/jlite-test/data` did not get duplicated into `Guillaume-Garrigos/jlite_debug_local_import`.

Below is a screenshot of the second lauched lab, showing the spurious file.

![Image](https://github.com/user-attachments/assets/7182b938-1067-4ad2-9fa1-7db5b326c9cc)

## Expected behavior

Different instances of jupyterlite opened within the same browser should not communicate with each other.

## Context

- JupyterLite version: I do not know, I am just using the demo repo
- Operating System and version: Windows 10 22H2
- Browser and version: Firefox 136.0.1

Edit: add screenshot

jtpio:
Thanks @Guillaume-Garrigos for the report.

This is likely because the browser storage used in JupyterLite is shared for the same domains. So if the two JupyterLite instances correspond to GitHub Pages sites for the same user account (or organization), the host (username.github.io) will be the same.

[...]

PR #1612, opened 2025-04-02
Fix handling of contents when multiple lite sites are shared under the same host

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1601

So multiple deployments under the same domain, for example the same user or GitHub organization, do not share files.

## Code changes

- [x] Use `baseUrl` as part of the default `contentsStorageName` for the drive
- [x] Add to the migration guide

This is what it looks like in the dev tools for this PR:

image

## User-facing changes

Existing deployments may not find their previous content stored in the browser anymore, since the default storage name will have changed.

### DIFF
--- a/app/jupyterlite.schema.v0.json
+++ b/app/jupyterlite.schema.v0.json
@@ -101,7 +101,7 @@
           "$ref": "#/definitions/localforage-driver-set"
         },
         "contentsStorageName": {
-          "description": "name used to store Jupyter contents in the browser",
+          "description": "Name used to store Jupyter contents in the browser. The baseUrl is appended to the default value",
           "type": "string",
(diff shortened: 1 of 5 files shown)


jtpio:
cc @Guillaume-Garrigos, as discussed in #1601 this PR adds the `baseUrl` as part of the storage name by default.

In practice, this also means that updating from 0.5 to 0.6 will require users of a JupyterLite deployment to download the files they have created (if any) in their browser storage.

jtpio:
Moving forward with this as discussed in https://github.com/jupyterlite/jupyterlite/issues/1601, and so this can be better tested in the next pre-releases.

Issue #1582, opened 2025-02-20
Random `FileNotFoundError` when repeatedly calling `np.save` with the same jupyterlite site is opened in 2 concurrent browser tabs


ogrisel:
The following code seems to crash because of a bad interaction with the jupyterlite synchronized filesystem.

Steps to reproduce:

- open https://jupyter.org/try-jupyter/lab/
- open https://jupyter.org/try-jupyter/lab/ again in a different browser tabs while keeping the previous tab open;
- create a pyodide notebooks with the following cell in one of the 2 jupyterlite instance:

```python
import numpy as np
from pathlib import Path

folder = Path("generated_data")
folder.mkdir(exist_ok=True)

for i in range(1000):
    data = np.random.randn(int(1e1))
    np.save(folder / f"data_{i}.npy", data)
    if i % 10 == 0:
        print(i)
```

- run it and get:

```python
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In[1], line 9
      7 for i in range(1000):
      8     data = np.random.randn(int(1e1))
----> 9     np.save(folder / f"data_{i}.npy", data)

File /lib/python3.12/site-packages/numpy/lib/_npyio_impl.py:570, in save(file, arr, allow_pickle, fix_imports)
    568     if not file.endswith('.npy'):
    569         file = file + '.npy'
--> 570     file_ctx = open(file, "wb")
    572 with file_ctx as fid:
    573     arr = np.asanyarray(arr)

FileNotFoundError: [Errno 44] No such file or directory: 'generated_data/data_455.npy' 
```

The precise loop iteration index that triggers the failure is random.

The JS console displays:

<details>

```
Loading numpy [pyodide.asm.js:10:94840](https://cdn.jsdelivr.net/pyodide/v0.27.1/full/pyodide.asm.js)
Loaded numpy [pyodide.asm.js:10:94840](https://cdn.jsdelivr.net/pyodide/v0.27.1/full/pyodide.asm.js)
XHRGET
https://probabl-ai.github.io/calibration-cost-sensitive-learning/jupyterlite/api/contents/generated_data/all.json
[HTTP/2 404  138ms]

don't worry, about SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data... nothing's broken. If there had been a
[...]

PR #1608, opened 2025-03-28
Use a `MessageChannel` instead of `BroadcastChannel` to communicate with the Service Worker

jtpio:
## References

As noticed in https://github.com/jupyterlite/jupyterlite/issues/1582#issuecomment-2759081633, the use of a `BroadcastChannel` seems to be causing issues when multiple browser tabs are opened.

A `BroadcastChannel` is not strictly needed to communicate with the service worker in the same tab

## Code changes

- [x] Use a `MessageChannel` instead of `BroadcastChannel` to communicate with the Service Worker
- [x] Remove `IBroadcastChannelWrapper` entirely
- [x] Mention in the migration guide

## User-facing changes

Less bugs related to the file system access

## Backwards-incompatible changes

- Remove `IBroadcastChannelWrapper`

PR #1621, opened 2025-04-10
`BroadcastChannel` with identifier to properly send responses from the Service Worker

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1582

Alternative to https://github.com/jupyterlite/jupyterlite/pull/1610

## Code changes

- [x] Revert https://github.com/jupyterlite/jupyterlite/pull/1608
- [x] Keep `BroadcastChannel`
- [x] Add an identifier to properly respond from the Service Worker
- [x] Update migration guide
- [x] Update contents documentation
- [x] Add UI tests for checking multi-tab support, if possible

Kernels need to use the new identifier to have file system support.


packages/contents/src/drivefs.ts
```diff
@@ -80,6 +80,16 @@ export type TDriveRequest<T extends TDriveMethod> = {
...
 
+  /**
+   * The expected receiver of the request
+   */
+  receiver?: 'broadcast.ts';
+
+  /**
+   * A unique ID to identify the origin of this request
+   */
+  originId?: string;
```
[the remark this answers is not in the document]
jtpio: `browsingContextId` could also work, based on the terms used in:

- https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API
- https://developer.mozilla.org/en-US/docs/Glossary/Browsing_context

jtpio:
UI tests are passing with the changes in the Pyodide kernel: https://github.com/jupyterlite/pyodide-kernel/pull/182

This also appears to be working fine on RTD when writing contents from multiple tabs:

https://github.com/user-attachments/assets/2e81b1eb-d22d-4aae-a726-afcb64051391

Maybe we can move forward with that fix and make a new alpha release, then update https://github.com/jupyterlite/pyodide-kernel/pull/182.

jtpio:
cc @trungleduc who has also been looking into service worker issues, in case you would like to take a look

packages/contents/src/drivefs.ts
```diff
@@ -550,24 +555,28 @@ export abstract class ContentsAPI {
...
   request<T extends TDriveMethod>(data: TDriveRequest<T>): TDriveResponse<T> {
     const xhr = new XMLHttpRequest();
     xhr.open('POST', encodeURI(this.endpoint), false);
 
+    // Add the origin browsing context ID to the request
+    const requestWithBrowsingContextId = {
+      ...data,
+      browsingContextId: this._browsingContextId,
+    };
+
```
trungleduc, line 577: Do you think embedding the context ID via the request header is slightly cleaner? It helps avoid polluting the payload data.

trungleduc: Indeed, so maybe we could keep data as an object key and not spread it.

jtpio: Wouldn't keeping `data` collide with the existing `data` field used for some of the methods?

https://github.com/jupyterlite/jupyterlite/blob/b01f12f02ccc18aa0ab988e8162d9b4fc57cf93d/packages/contents/src/drivefs.ts#L87

For example with `mknod`:

image

trungleduc: i meant 

```ts
    const requestWithBrowsingContextId = {
      data,
      browsingContextId: this._browsingContextId,
    };
```

trungleduc: But it's just a minor detail. The remaining logic looks good

jtpio: ok I just pushed https://github.com/jupyterlite/jupyterlite/pull/1621/commits/0d009b4a9d04698f403f4ea1a747ddf71f91e393 to do something along those lines.

jtpio:
Thanks @trungleduc for the review 👍 

I'll go ahead and make a new alpha release with this change, then make sure https://github.com/jupyterlite/pyodide-kernel/pull/182 works as expected with the new pre-release.

Issue #1364, opened 2024-04-02
Menu "Save and export notebook as" not showing up


Popcorn7020:
## Description

In the online version of jupiter lab (JupyterLite) I'm unable to use the "Save and export notebook as", no menu is showing up
![image](https://github.com/jupyterlite/jupyterlite/assets/155472414/2419eac4-ec4d-4ff8-8889-647fb7d49293)

## Reproduce

1. Go to https://jupyter.org/try-jupyter/lab/
2. Select File menu
3. Select "Save and export notebook as" menu

## Expected behavior

I expect that a menu shows up to allow exporting and downloading the notebooks

## Context

- Browser and version: Edge 123.0.2420.65 and Chrome 123.0.6312.86

jtpio:
Thanks @Popcorn7020.

This looks like a duplicate of https://github.com/jupyterlite/jupyterlite/issues/543. Should we close it and continue the discussion in https://github.com/jupyterlite/jupyterlite/issues/543?

WesErg:
This is not a duplicate of #543.  I am suffering the same problem.  Is there any solution to this?

horasjey:
i'm having this problem, 
anoyone have solusi..?

vmoreno14:
Did anyone find a solution

stientist:
Hi everyone, this worked for me: After making your notebook, just go to https://jupyter.org/try-jupyter/tree/, which is basically like "file explorer" for your online jup nb's. Navigate to "notebooks" and just download! Pls let me know if this worked for you!  

Assuming that my solution actually works for everyone else,  I think that jupyter's IT did not respond because it seems like it would be easy for us to troubleshoot if we put in some effort. However, I figured this out by accident, and I have no idea how I even got to the tree from my initial notebook. Although I agree it was a straightforward solution, clearly there are many confused but competent users.

nicae-dev:
it's not working for me... We've installed jupyter on local server. In chrome console there is a 500 error in JS:
`[jupyter-domain]/user/username/api/nbconvert?1731677498679 500 (Internal Server Error)`

and if i try to run nbconvert in Ubuntu console, it ends up with strange error:

` jupyter nbconvert --to pdf test_simple.ipynb`

```
Traceback (most recent call last):
  File "/usr/local/bin/jupyter-nbconvert", line 5, in <module>
    from nbconvert.nbconvertapp import main
[...]

Issue #543, opened 2022-03-09
jupyterlite doesn't have print option


NAbdolkhani:
In this new interface, the print option is gone and there is no way to save the notebook as pdf. Is there an alternative? or can't you just bring the print option back?

bollwyvl:
> just bring the print option back?

The print feature upstream relies on server-side `nbconvert` to generate the "dead pixels" HTML.

To get to something better we have a few options: 
- run `nbconvert` in python in the browser
  - this would probably be the most faithful to the existing experience, and might support plugins, etc.
  - at present `nbconvert` cannot be installed directly, due to a missing `tornado` and `pyzmq`
    - we could _potentially_ shim these... or provide a patched `nbconvert`
  - we've been trying to avoid _requiring_ any heavy-duty wasm inside the server component 
- re-implement some subset of `nbconvert` in JS
  - this would probably be the most performant
  - theoretically, we could pack all of the jinja templates from `nbconvert` and use `nunjucks`, but...
- work with more directly with in-app HTML
  - this would potentially be interesting, as it might handle a broader swathe of outputs (canvas, etc.)

pierrepo:
Dear @bollwyvl I had a similar issue while trying to export a notebook as html. If I understand correctly your previsous answer, this is also due to the fact that `nbconvert` (or any substitute) is not (yet) available in jupyterlite. Am I correct?

bollwyvl:
nbconvert (with mad hacks) is now (somewhat) available in (a soon-to-be-released) jupyterlite.

https://github.com/jupyterlite/jupyterlite/pull/756#issuecomment-1214554290

sean-morris:
@bollwyvl I am hung up on the "somewhat" available part. From the link above, it appears that v0.1.0b12 includes the ability to print a notebook to a pdf. I working with this version. 

The example from this [gist](https://gist.github.com/bollwyvl/6b3cb4c46b1764c6d9ae1e5831f86d7a#file-nbconvert-in-jupyterlite-ipynb) is pretty intense. Is the implication that you need to run through the commands on the page referenced to create a context for converting and downloading a notebook to a PDF? 

I have moved through the gist but ran into issues right at the end of the process. The HTMLExporter in nbconvert has been moved for one, that was not a big deal -- nbconvert.exporters.HTMLExporter -- but the second involves a problem with bleach not having the attribute ALLOWED_STYLES defined. The error suggests upgrading bleach to 5 or greater. But this did not work. 

[...]

PR #1625, opened 2025-04-16
Disable the notebook export plugin

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1364

## Code changes

Disable the `@jupyterlab/notebook-extension:export` plugin in all relevant apps.

## User-facing changes

For end users, this has the effect of not showing the "Save and Export Notebook As..." menu item anymore.

This menu item didn't have any submenu anyway, so until there is a way to properly export a notebook.

image

None

### DIFF
--- a/app/consoles/package.json
+++ b/app/consoles/package.json
@@ -316,6 +316,7 @@
       "@jupyterlab/help-extension:about",
       "@jupyterlab/help-extension:open",
       "@jupyterlab/lsp-extension:plugin",
+      "@jupyterlab/notebook-extension:export",
       "@jupyterlab/notebook-extension:execution-indicator",
       "@jupyterlab/notebook-extension:kernel-status",
       "@jupyterlab/services-extension:config-section-manager",
--- a/app/edit/package.json
+++ b/app/edit/package.json
@@ -332,6 +332,7 @@
       "@jupyterlab/help-extension:about",
       "@jupyterlab/help-extension:open",
       "@jupyterlab/lsp-extension:plugin",
+      "@jupyterlab/notebook-extension:export",
       "@jupyterlab/notebook-extension:execution-indicator",
       "@jupyterlab/notebook-extension:kernel-status",
       "@jupyterlab/services-extension:config-section-manager",
--- a/app/lab/package.json
+++ b/app/lab/package.json
@@ -322,6 +322,7 @@
       "@jupyterlab/filebrowser-extension:share-file",
       "@jupyterlab/help-extension:about",
       "@jupyterlab/lsp-extension:plugin",
+      "@jupyterlab/notebook-extension:export",
       "@jupyterlab/services-extension:config-section-manager",
       "@jupyterlab/services-extension:connection-status",
       "@jupyterlab/services-extension:default-drive",
--- a/app/notebooks/package.json
+++ b/app/notebooks/package.json
@@ -342,6 +342,7 @@
       "@jupyterlab/help-extension:about",
       "@jupyterlab/help-extension:open",
       "@jupyterlab/lsp-extension:plugin",
+      "@jupyterlab/notebook-extension:export",
       "@jupyterlab/notebook-extension:execution-indicator",
       "@jupyterlab/notebook-extension:kernel-status",
       "@jupyterlab/services-extension:config-section-manager",
--- a/app/tree/package.json
+++ b/app/tree/package.json
@@ -298,6 +298,7 @@
       "@jupyterlab/help-extension:about",
       "@jupyterlab/help-extension:open",
       "@jupyterlab/lsp-extension:plugin",
+      "@jupyterlab/notebook-extension:export",
       "@jupyterlab/notebook-extension:execution-indicator",
       "@jupyterlab/notebook-extension:kernel-status",
       "@jupyterlab/notebook-extension:language-server",


bollwyvl:
Maybe providing at least the .ipynb (whatever is it's called in notebook)
would be an alternative, as there is no dedicated "Download" option?

jtpio:
Yeah, I've also been thinking after opening the PR that disabling the plugin entirely may be a bit too much, in case some extensions we don't know about provide some exporters.

There is already a way to download the notebook via a command. So maybe there could be at least one entry to do that, good point 👍 Also maybe JupyterLite could technically support more plain text format too, like Markdown.

jtpio:
Played a bit more with this, to check what could be done with libraries like `jsPDF`: https://github.com/parallax/jsPDF

https://github.com/user-attachments/assets/587c235e-d543-4032-b93a-2d10477ab7b7

bollwyvl:
Ha, yeah, jspdf and friends are great, until they aren't. Still want
nunjucks or minijinja to allow for the slicing and dicing people crave, and
it's a lot easier to make accessible html then pdf, beyond just being a
precursor. Back on the jupyter-desktop thing from last week, though: that
is actually _one_ compelling argument for electron/qt: direct, no-hack
access to pdf output, straight off the rendering engine. It's really quite
something to get razor sharp bqplot, etc.

bollwyvl:
Right, I love some pdf, epub, etc, especially if it these could optionally embed the original notebook+supporting assets inside, and lite could expose those all the way down to kernels. But getting to "can get the editable notebook i just made in one-to-two clicks" is the first, biggest thing.

```mermaid
graph TB
subgraph build
    tmpl-sys["sys.prefix/share/jupyter/nbconvert/templates"]
    tmpl-lite["{lite_dir}/nbconvert/templates"]
    tmpl-sys & tmpl-lite --> tmpl-out["{out_dir}/nbconvert/templates"]
end
subgraph run
    subgraph contents
        Untitled.ipynb
        Untitled.html
        Untitled.pdf
        Untitled.pdf -.-> Untitled.pdf.d/
        subgraph Untitled.pdf.d/
            Untitled.ipynb.in.pdf.d[Untitled.ipynb]
        end
    end
    Untitled.ipynb & tmpl-out --> nbconvert-lite
    to-html --> .html
    subgraph nbconvert-lite
        to-html([".convert(to='html')"])

        subgraph nbconvert-lite-pdf
            to-pdf([".convert(to='pdf', embed=['.ipynb'])"]) 
            --> Untitled.pdf.html 
            --> jspdf([jspdf]) --> Untitled.pdf
        end
    end
    contents --> drive --> kernel
end
```

krassowski:
One thing which does not work very well is the LaTeX formula rendering:

```tex
\int_2^3 x^2 \, dx=\frac{3^3}{3}-\frac{2^3}{3}=\frac{19}{3}
```

| Notebook | PDF |
|--|--|
| image | image |

Edit: some links:
- https://stackoverflow.com/questions/71750159/jspdf-unable-to-render-minus-sign-from-html-and-mathjax
- https://github.com/parallax/jsPDF/issues/953 suggests to reconfigure mathax to output SVG; we could do that for export

krassowski:
Another issue is around wrapping of rendered markdown cells:

| Notebook | PDF |
|--|--|
|image | image |

krassowski:
Another enhancement idea: should we remove active cell class for export?

jtpio:
> Another enhancement idea: should we remove active cell class for export?

Yes that make sense. In general anything that can make the notebook export look better (and closer to the nbconvert output) would likely be useful at this point.

jtpio:
@agriyakhetarpal @krassowski just checking if you are still interested in upstreaming the work you did in https://github.com/JupyterEverywhere/jupyterlite-extension, in a separate plugin in JupyterLite?

agriyakhetarpal:
Yes, definitely, @jtpio! We are also tracking the same and have created an issue here: https://github.com/JupyterEverywhere/jupyterlite-extension/issues/98. I am unsure what our prioritisation is/will be, but I think it can be something that I can take up outside of work time.

To clarify, will the following be the intended approach?
- revive the export plugin and add a PDF option
- improve the styling on a best-effort basis*
- wire it up to the "Export and Save Notebook As" option in the "File" navbar menu options

*We have discussed the ramifications of using something like `jsPDF` above, and I think this is a much harder task than expected. The output is hardly usable for real-world even after much of the effort I performed when attempting to tame `jsPDF`; it would need to be styled pretty extensively to resemble JupyterLab/nbconvert-style PDFs. At one point, I think it would also be useful to see if we can get a serverless `nbconvert` (with SwiftLaTeX for a WASM LaTeX engine) or WASM-compiled Pandoc working (as discussed in jupyter/nbconvert#1766).

jtpio:
> To clarify, will the following be the intended approach?

I think this sounds like a fine approach. If we are not sure yet about the export quality we would get from `jsPDF`, this could also be developed as an extension in a separate repo, that could then be installed in a JupyterLite deployment.

jimwhite:
I asked Perplexity and there are a multiple ports of LaTeX to WASM.  One of these could fix the formatting quality issue.

https://www.swiftlatex.com/
https://github.com/busytex/busytex

There are also Pandoc ports to WASM, although the top two both appear to have lost some maintainer support.

https://github.com/georgestagg/pandoc-wasm
https://github.com/haskell-wasm/pandoc-wasm

They're probably bigger than jsPDF so being able to choose one or the other might still be useful.
My thought is making a lite version of nbconvert which is probably a bigger issue, so this comment is probably not exactly on topic.

jtpio:
Thinking more about this, maybe we could first limit the scope of this PR to exporting to basic text formats.

And then see if there could be a way for extensions to add new exporters, so we could experiment more with PDF exports outside of the base JupyterLite repo.

jtpio:
Just pushed a couple of changes to:

- drop support for PDF exports for now
- integrate changes from https://github.com/jupyterlab/jupyterlab/pull/17482
- add some basic UI tests

image

jtpio:
Maybe we should look into adding a way to provide custom exports separately.

For example tackle https://github.com/jupyterlite/jupyterlite/issues/1629 first to have `@jupyterlite/services` package, before exposing a new `Token`. The `LiteNbConvertManager` would then use that new token to get the list of available exporters.

jtpio:
If folks would like to try it out: https://jupyterlite--1625.org.readthedocs.build/en/1625/_static/lab/index.html

jimwhite:
> If folks would like to try it out: https://jupyterlite--1625.org.readthedocs.build/en/1625/_static/lab/index.html

I gave it a whirl and kinda works but the menu appears too low and I gotta move mouse quickly but carefully to pick one of the formats before it disappears.   On current MacOS Chrome.
image

jtpio:
> I gave it a whirl and kinda works but the menu appears too low and I gotta move mouse quickly but carefully to pick one of the formats before it disappears. On current MacOS Chrome.

Ah right, this issue seems to be specific to lite sites deployed ReadTheDocs (not sure why though).

As an alternative, it's also possible to grab the build artifact from https://github.com/jupyterlite/jupyterlite/actions/runs/19314489744?pr=1625 and serve it locally with `python -m http.server`.

jimwhite:
> > I gave it a whirl and kinda works but the menu appears too low and I gotta move mouse quickly but carefully to pick one of the formats before it disappears. On current MacOS Chrome.
> ... it's also possible to grab the build artifact from https://github.com/jupyterlite/jupyterlite/actions/runs/19314489744?pr=1625 and serve it locally with `python -m http.server`.

Tried that out and the menu looks correctly placed.  I checked that I got both scripts and notebooks from "Saved and Export As...".

jtpio:
Thanks for testing! I'll go ahead with this PR then so we can continue iterating on it.

agriyakhetarpal:
Cross-posting the discussion happening today in the original issue here: https://github.com/jupyterlite/jupyterlite/issues/543#issuecomment-3971213627 (and the comments that followed), which has some details about how work in the community is proceeding regarding PDF exports.

Issue #1626, opened 2025-04-17
Cell execution stuck while one is writing to the file system in another browser tab


jtpio:
## Description

Follow-up to https://github.com/jupyterlite/jupyterlite/issues/1582.

A bug likely uncovered by #1621.

## Reproduce

In the first tab, create a notebook with the following content and execute:

```py
from pathlib import Path

for i in range(1000):
    Path('test1').write_text('1')

    if i % 10 == 0:
        print(i)

print('done')
```

In a second tab:

```py
from pathlib import Path
Path('test2').write_text('2')
```

![Image](https://github.com/user-attachments/assets/7a557fb6-f049-4691-ad01-7350609756d6)

## Expected behavior

The cell in the second tab should execute normally.

## Context

As discussed in https://github.com/jupyterlite/jupyterlite/issues/1582#issuecomment-2812007721 and spotted by @lesteve.

Reproduced with `0.6.0a8` and on `main` : 2f04dd7ef14876693ff5af30d9fa856c0f82a2ed

lesteve:
From https://github.com/jupyterlite/jupyterlite/issues/1582#issuecomment-2812007721 there was also the `Uncaught (in promise) TypeError: e.data is null` console log error in one browser tab where the other tab cel is interacting with the filesystem, but that may actually be the same underlying bug as this issue not sure.

jtpio:
> From [#1582 (comment)](https://github.com/jupyterlite/jupyterlite/issues/1582#issuecomment-2812007721) there was also the `Uncaught (in promise) TypeError: e.data is null` console log error in one browser tab where the other tab cel is interacting with the filesystem, but that may actually be the same underlying bug as this issue not sure.

Yeah, likely a hint there is still something not working quite as expected w.r.t the communication with the `ServiceWorker` via the `BroadcastChannel`.

jtpio:
https://github.com/jupyterlite/jupyterlite/pull/1627 should fix the dev tools error, but not the actual issue yet.

It seems related to the asynchronous passing of messages via the `BroadcastChannel` in the Service Worker here:

https://github.com/jupyterlite/jupyterlite/blob/2f04dd7ef14876693ff5af30d9fa856c0f82a2ed/packages/server/src/service-worker.ts#L137-L148

jtpio:
@lesteve if you would like to try the PR on the ReadTheDocs preview: https://jupyterlite--1627.org.readthedocs.build/en/1627/_static/lab/index.html

[...]

PR #1627, opened 2025-04-17
Fix concurrent file system access from multiple tabs

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1626

## Code changes

- [x] Better handle requests in the Service Worker
- [x] UI test

## User-facing changes

Hopefully less issues.

It should be possible to work on two notebooks in two different browser tabs at the same time:

https://github.com/user-attachments/assets/6cb61b02-0be2-40ad-861b-10b7e5052886

None

### DIFF
--- a/packages/server/src/service-worker.ts
+++ b/packages/server/src/service-worker.ts
@@ -139,8 +139,7 @@ async function broadcastOne(request: Request): Promise<Response> {
   const promise = new Promise<Response>((resolve) => {
     const messageHandler = (event: MessageEvent) => {
       const data = event.data;
-      const browsingContextId = data.browsingContextId;
-      if (browsingContextId !== message.browsingContextId) {
+      if (data.browsingContextId !== message.browsingContextId) {
         // bail if the message is not for us
         return;
       }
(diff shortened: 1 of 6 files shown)


packages/server/src/service-worker.ts
```diff
@@ -135,13 +135,23 @@ function shouldDrop(request: Request, url: URL): boolean {
...
+      if (browsingContextId !== message.browsingContextId) {
+        // bail if the message is not for us
+        return;
+      }
+      const response = data.response;
+      resolve(new Response(JSON.stringify(response)));
+      broadcast.removeEventListener('message', messageHandler);
     };
+
+    broadcast.addEventListener('message', messageHandler);
```
jtpio, line 152: It looks like the previous logic was setting `broadcast.onmessage` for each request, which would override the function previously defined by an in-flight request.

`addEventListener` / `removeEventListener` should help handle messages from multiple closures, when several (concurrent) file system requests are in-flight.

trungleduc (approved):

jtpio:
Thanks @trungleduc and @lesteve for the review and testing.

Merging so the fix can be available in the next pre-release.

andeplane:
Ahh thanks for fixing this! I never figured out exactly how to do it well, but a context id is a good solution

Issue #619, opened 2022-04-25
Improved loading indicator


tuanchoang:
### Problem

Right now, when JupyterLite is loading/launching in the browser, it could take a while to load and the unsuspecting user could be waiting somewhere ~10 seconds wondering if anything is happening before loading completes.

### Proposed Solution

JupyterLite could provide a better user experience by visualizing a progress bar and/or loading indicator while it is launching. This gives the user a sense of time and provides context that JupyterLite is loading and the expectation is to wait for it to finish loading.

jtpio:
Thanks @tuanchoang for the suggestion :+1: 

> Right now, when JupyterLite is loading/launching in the browser, it could take a while to load and the unsuspecting user could be waiting somewhere ~10 seconds wondering if anything is happening before loading completes.

Just to clarify, does that happen during the initial page load? Or are you referring to the slow loading times of the Python kernel when a notebook is opened?

tuanchoang:
@jtpio Sorry I could have been clearer. :)

Yeah, I'm referring to the initial page load. To the end-user, it could look like nothing is happening when in actuality JupyterLite hasn't finished loading.

jtpio:
Alright thanks!

Maybe there could then be some kind of static spinner / loaded that could be loaded on the page on startup: https://github.com/jupyterlite/jupyterlite/blob/main/app/lab/index.template.html

And removed when the app is ready.

tuanchoang:
Exactly, that'd be perfect! Just some sort of indicator to let the user know to wait and/or that it's loading.

bollwyvl:
Sounds great: lots of other stuff going on, so I'm not sure who would do this, and therefore, when this would land. Some thoughts:

This is one of the few things that _can't_ be an extension, and probably _does_ need to be replaceable, if not particularly _configurable_. My gut feel: 
- make a new file in the root called e.g. `loading.html`
  - copy it from the `lite_dir` or use a CLI option `--loading-html`
- load that as an iframe in the DOM of each `index.html`, no JS, no stacking
- in `config-utils.js`
  - start a `message` listener which can clear the iframe
    - also have a timeout, so if something _does_ go wrong
      - this _can_ be configurable
- in each of our base applications, `window.postMessage` right _before_ the moon animation starts
  - themes should still control those

`postMessage` is going to be A Thing, anyway (I've already hacked a couple of them up) so maybe we need to think about that some.

jtpio:
[...]

PR #1636, opened 2025-04-27
Add a loading indicator

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/619

## Code changes

- [x] Show a loading indicator before the actual app starts
- [x] Check if a theme is already saved in IndexedDB to theme the indicator properly and avoid bright flashes on page refresh
- [x] New `showLoadingIndicator` option which defaults to `false`, but is enabled for the `lab` app by default
- [x] Add UI test

## User-facing changes

Using a CPU slowdown in the dev tools to have the time to look at the loading indicator:

image

https://github.com/user-attachments/assets/a46f7434-b4d0-4a68-b71c-8227f1a74f4c

None

### DIFF
--- a/app/bootstrap.js
+++ b/app/bootstrap.js
@@ -38,6 +38,58 @@ function getOption(name) {
   return _CONFIG_DATA[name] || '';
 }
 
+/**
+ * Apply theme to loading indicator based on saved settings in IndexedDB
+ */
+async function applyThemeToAppLoadingIndicator() {
+  try {
+    const baseUrl = getOption('baseUrl');
+    const defaultStorageName = `JupyterLite Storage - ${baseUrl}`;
+    const storageName = getOption('settingsStorageName') || defaultStorageName;
+
+    const localforageModule = await import('localforage');
+    const localforage = localforageModule.default;
+
+    const settingsDB = localforage.createInstance({
+      name: storageName,
+      storeName: 'settings'
+    });
+
+    const key = '@jupyterlab/apputils-extension:themes';
+    const settings = await settingsDB.getItem(key);
+
+    let isDarkTheme = false;
+
+    if (settings) {
+      // Use regex to find "theme": "name of theme" pattern, since the
+      // settings are stored as a raw string
+      const themeRegex = /"theme"\s*:\s*"([^"]+)"/i;
+      const matches = settings.match(themeRegex);
+
+      if (matches && matches[1]) {
+        const themeName = matches[1].toLowerCase();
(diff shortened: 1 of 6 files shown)


jtpio:
cc @agriyakhetarpal in case you would like to review, as it may indeed be relevant for https://github.com/jupyterlite/jupyterlite-sphinx/issues/287

Do you think it could already be worth adding a hook to show / hide the loading indicator for all apps, and not just JupyterLab?

app/bootstrap.js
```diff
@@ -38,6 +38,78 @@ function getOption(name) {
...
+    } else {
+      document.body.classList.add('jp-mod-light');
+    }
+  } catch (e) {
+    console.warn('Could not apply theme to loading indicator:', e);
+    // Fallback to light theme on error
+    document.body.classList.remove('jp-mod-dark');
+    document.body.classList.add('jp-mod-light');
+  }
+}
```
agriyakhetarpal, line 111: Perhaps we could sniff the browser settings by checking for the `data-theme` and `data-mode` HTML attributes if they exist, and use them to determine the theme instead of falling back to a light theme by default?

agriyakhetarpal (approved):
Thanks for asking, @jtpio! I've added a response/suggestion at jupyterlite/jupyterlite-sphinx#287, which I think ties in with the loading indicators nicely. It would certainly be useful to add such a hook for all apps beyond just Lab.

app/bootstrap.js
jtpio: You mean `data-jp-theme` on `body`?

image

I think these attributes are set after the application has started, so they are likely added a bit too late in the bootstrap process.

jtpio:
> It would certainly be useful to add such a hook for all apps beyond just Lab.

Added the `showLoadingIndiator` option in https://github.com/jupyterlite/jupyterlite/pull/1636/commits/42ae35899d7bb020a8e4a32bc74e04e7fc15d27b

app/bootstrap.js
jtpio: Ah right, but those attributes will be in the parent window, and will need to be "sent" to the IFrame. Maybe https://github.com/jupyterlite/jupyterlite/issues/1637 could help define custom logic to get these attributes.

jtpio:
Probably good enough for now. We can get this in so it's easier to test in the pre-releases.

If anyone would like to make improvements to the loading indicator please feel free to open a new issue or PR, thanks!

Issue #275, opened 2021-07-19
await input()


davidbrochart:
Just wanted to share this issue in case someone has ideas.

### Problem

With pyolite we can now get a user input with:
```python
i = await input()
```
which is different than what we are used to in ipython:
```python
i = input()
```
The fact that we need to `await` is problematic because it means that wherever this is used, the whole call stack must now be async (you cannot use it in a regular Python function).

### Proposed Solution

There is already a running event loop in Jupyterlite, so we can't just run `input` until complete.
There are two tricks that I know of to work around that, unfortunately they don't work:
- running `input` until complete in another thread with its own event loop (pyolite doesn't support threads).
- using `nest-asyncio` and run `input` until complete in the current event loop (nest-asyncio `Can't patch loop of type <class 'pyodide.webloop.WebLoop'>`).

dsblank:
I am also interested in how to solve such a problem, as it is needed for the %pip magic discussion #237 (and other places).

pdeitel:
Hi, Is there any way to avoid having to use `await` with `input`? I am looking at JupyterLite for a novice-programmer-oriented product I am working on. I'd like to avoid having to introduce `await` to a novice.

joemarshall:
It is possible to do this if you use a service worker. I haven't done it in my jupyter because I didn't need it, but I do use a service worker for other things that need to be async in javascript whilst sync in the python webworker.

How it works is the service worker handles fetch requests, which you can trigger from a webworker (like pyolite is) by calling synchronous xmlhttprequest for a special URL that you define. The XMLHttpRequest waits until the fetch responds. So you can call the request, then in the serviceworker fire off whatever async things you want, and only respond to the fetch once that is done.

This can also be done to implement time.sleep without having to use anything clever which is nice.

pdeitel:
@joemarshall Thanks. But would this code also need to be in the notebook? I am trying to avoid showing my students anything they don't need to see. Also, any chance you have a link to some sample code I can look at?

joemarshall:
No it is a javascript that runs on the site, plus a python module that needs importing. I haven't got my service worker stuff anywhere accessible yet, will try and remember to post it up.

martinRenou:
@joemarshall I've just spent quite some time trying to have a service worker implementing a sleep function like in this blogpost: https://jasonformat.com/javascript-sleep/

My service worker seems to install and activate properly, but the "fetch" event is never triggered.

[...]

PR #1640, opened 2025-05-01
Support kernel `stdin` request via ServiceWorker and SharedArrayBuffer

ianthomas23:
This PR adds support in JupyterLite for kernel synchronous `stdin` requests via SharedArrayBuffer and Service Worker, fixing #275. It works with PRs in other repos, most of which have been merged but not released yet. I have tested it locally with the following combinations of kernel and functions:

Kernels | Functions
--- | ---
pyodide, xeus-python | `input`, `getpass.getpass`
xeus-r | `readline`
xeus-cpp | `std::cin`

Obligatory screencast (using xeus-python and Service Worker):

https://github.com/user-attachments/assets/d11abfd0-ff67-4cf7-8fb7-d4ee6819fa65

The changes for SharedArrayBuffer are minimal as we send and receive kernel messages via the usual mechanism to the kernel in the UI thread. Most of the changes are in the kernel repos.

The changes for Service Worker are greater, and they are intended to also support `stdin` requests for the JupyterLite `terminal` which I will work on soon. The high-level view is that a kernel/terminal makes a synchronous `XMLHttpRequest` to `baseUrl + '/stdin/kernel'` (or `'/stdin/terminal'`) and blocks the kernel/terminal webworker until the reply is received. The Service Worker handles this request, passing it on to JupyterLite running in the UI thread, which does what it need to get input from the user in an `async` manner and return the reply.

Changes to the Service Worker itself are minimal, reusing the same `BroadcastChannel` method used for drive requests. The Service Worker and `ServiceWorkerManager` don't know anything about the `data` coming in, they just pass it on unchanged. For the remainder of the code I have made some design decisions which I think are reasonable, but I am happy to change any of this in line with whatever JupyterLite core devs think is best.

To handle `'/stdin/<suffix>'` requests the consumer (kernel or terminal) registers a `stdinHandler` for a particular `suffix`, and the `ServiceWorkerManager` calls the appropriate one based on the `suffix` of an incoming message. For `terminal` the handler will be located in the `terminal-extension` code so won't require any extra support in the JupyterLite code base. For `kernel` the handler is in `LiteKernelClient` and it has to extract the `clientId` from the `input_request` message, find the corresponding `IKernel` and send the `input_request` via that kernel. Some of this code is a little awkward, to get the `kernelId` for a `clientId` I am walking the opposite mapping rather than store a new mapping. And to send the `input_request` to the frontend I need the `sendMessage` which is not exposed by an `IKernel`, so I am storing them in a new `_kernelSends` map similar to the existing `_kernels`, `_clients` and `_kernelClients` maps. Perhaps instead of this separate map it could be combined with the `kernelId` to `IKernel` map instead?

There are no new tests yet. I see that `ui-tests` includes some pyodide kernel tests so when jupyterlite/pyodide-kernel#183 is merged and in a pre-release, I will be able to add tests using that.


packages/application-extension/src/index.tsx
```diff
@@ -487,9 +489,17 @@ const serviceWorkerManagerPlugin: JupyterFrontEndPlugin<IServiceWorkerManager> =
...
   autoStart: true,
   provides: IServiceWorkerManager,
-  activate: (app: JupyterFrontEnd) => {
+  optional: [IKernelClient],
+  activate: (app: JupyterFrontEnd, kernelClient?: IKernelClient) => {
     const { contents } = app.serviceManager;
-    return new ServiceWorkerManager({ contents });
+    const serviceWorkerManager = new ServiceWorkerManager({ contents });
+    if (kernelClient !== undefined) {
+      serviceWorkerManager.registerStdinHandler(
```
ianthomas23, line 497: Here registering the stdin handler in the same plugin that creates `ServiceWorkerManager`. Maybe it should be in a separate plugin of its own?

packages/kernel/src/client.ts
```diff
@@ -278,12 +289,45 @@ export class LiteKernelClient implements Kernel.IKernelAPIClient {
...
+  }
+
   private _kernels = new ObservableMap<IKernel>();
   private _clients = new ObservableMap<WebSocketClient>();
   private _kernelClients = new ObservableMap<Set<string>>();
   private _kernelspecs: IKernelSpecs;
   private _serverSettings: ServerConnection.ISettings;
   private _changed = new Signal<this, IObservableMap.IChangedArgs<IKernel>>(this);
+  private _stdinPromise?: PromiseDelegate<KernelMessage.IInputReplyMsg>;
+  private _kernelSends = new ObservableMap<(msg: KernelMessage.IMessage) => void>();
```
ianthomas23, line 330: Here is the mapping for kernel `sendMessage` functions, could be combined with `_kernels` instead?

packages/server/src/service-worker.ts
```diff
@@ -151,6 +154,8 @@ async function broadcastOne(request: Request): Promise<Response> {
     broadcast.addEventListener('message', messageHandler);
   });
 
+  // Add URL pathname to message
+  message.pathname = url.pathname;
```
ianthomas23, line 158: Note adding the `url.pathname` to messages sent from ServiceWorker to `ServiceWorkerManager`. This is so `/stdin/kernel` and be distinguished from `/stdin/terminal` as otherwise this information does not get to the `ServiceWorkerManager`. The alternative is to not pass it and instead change the `registerStdinHandler` to take another function which is a callback to check if a particular stdin message should be handled by that handler by checking some specifics of the `data` passed. For example, the kernel stdin handler would check that `header.msg_type === 'input_request'`.

packages/server/src/service-worker.ts
```diff
@@ -117,7 +117,10 @@ async function refetch(request: Request): Promise<Response> {
  * Whether a given URL should be broadcast
  */
 function shouldBroadcast(url: URL): boolean {
-  return url.origin === location.origin && url.pathname.includes('/api/drive');
+  return (
+    url.origin === location.origin &&
+    (url.pathname.includes('/api/drive') || url.pathname.includes('/stdin/'))
```
martinRenou, line 122: I'd use the same naming convention as for the drive access. A better name for this would probably have been `/sw-api` or `/service-worker-api`, but it's out of scope for this PR.

```suggestion
    (url.pathname.includes('/api/drive') || url.pathname.includes('/api/stdin/'))
```

jtpio:
> I see that `ui-tests` includes some pyodide kernel tests so when [jupyterlite/pyodide-kernel#183](https://github.com/jupyterlite/pyodide-kernel/pull/183) is merged and in a pre-release, I will be able to add tests using that.

Similar to previous PRs that had changes both in JupyterLite and `jupyterlite-pyodide-kernel`, we can use a dev build of `jupyterlite-pyodide-kernel` uploaded to ReadTheDocs to already have a UI tests here. This will avoid having to make many pre-releases when things change and get adjusted on one side or another.

For reference:

- https://github.com/jupyterlite/jupyterlite/pull/1621 used this technique
- Also mentioned in https://github.com/jupyterlite/pyodide-kernel/pull/183#issuecomment-2836402137

packages/server/src/service-worker-manager.ts
```diff
@@ -174,19 +186,30 @@ export class ServiceWorkerManager implements IServiceWorkerManager {
   /**
    * Handle a message received on the BroadcastChannel
    */
-  private _onBroadcastMessage = async <T extends TDriveMethod>(
+  private _onBroadcastMessage = async (
     event: MessageEvent<{
-      data: TDriveRequest<T>;
+      data: any;
```
jtpio, line 191: Is there a way to keep a proper type for `data` instead of `any`?

packages/server/src/service-worker.ts
ianthomas23: I can change this, but it will need corresponding changes in `jupyterlite/xeus`and `pyodide-kernel`.

packages/server/src/service-worker-manager.ts
ianthomas23: No, it has to be `any`. The `stdin` requests intentionally do not need to be understood by the service worker or service worker manager, they just pass them through. This is so that clients of this functionality such as the Terminal can pass through whatever they need without JupyterLite understanding it. The alternative is that we define a type for `ITerminalStdinRequest` in either the Terminal or JupyterLite and have to import it into the other project, which is not good.

We could be more explicit here and use something like
```ts
data: TDriveRequest<T> | IStdinRequest
```
and have `IStdinRequest` as `any`.

packages/server/src/tokens.ts
```diff
@@ -59,6 +69,13 @@ export namespace IServiceWorkerManager {
      */
     workerUrl?: string;
   }
+
+  /**
+   * Interface for handler of stdin requests received via Service Worker.
+   */
+  export interface IStdinHandler {
+    (message: any): Promise<any>;
```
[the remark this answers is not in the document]
ianthomas23: No! Stdin requests are not just for kernels, they are for the terminal too!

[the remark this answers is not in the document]
jtpio: Maybe worth leaving a comment about that then.

packages/server/src/service-worker.ts
ianthomas23: Done in 17a6c0a.

ianthomas23:
> Similar to previous PRs that had changes both in JupyterLite and `jupyterlite-pyodide-kernel`, we can use a dev build of `jupyterlite-pyodide-kernel` uploaded to ReadTheDocs to already have a UI test here. This will avoid having to make many pre-releases when things change and get adjusted on one side or another.

OK, I will do this. But I note that there have already been 6 pre-releases of `pyodide-kernel` so from the outside it does not look like making pre-releases is something to be avoided.

packages/server/src/service-worker.ts
martinRenou: Thanks! Sorry for missing this in the other PRs. I just release jupyterlite-xeus 4.0.0a7 with the changes

jtpio:
> OK, I will do this. But I note that there have already been 6 pre-releases of `pyodide-kernel` so from the outside it does not look like making pre-releases is something to be avoided.

Right. It's not about the number of pre-releases, but to make it simpler to test the changes from the two repos while reducing maintenance load, when there is a dependency between PRs like it's the case here.

ianthomas23:
I've disabled the new `stdin` ui-tests on firefox because IIUC firefox >= 138 is required for the ServiceWorker to support this functionality.

packages/server/src/tokens.ts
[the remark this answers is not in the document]
ianthomas23: Extra documentation on `any` added in 3268f48.

packages/server/src/service-worker-manager.ts
ianthomas23: Extra documentation on any added in 3268f48.

jtpio: Thanks!

jtpio:
> I've disabled the new `stdin` ui-tests on firefox because IIUC firefox >= 138 is required for the ServiceWorker to support this functionality.

hmm there are other tests that use the Service Worker and should be passing with Firefox. Would be interesting to check the Playwright report, but the previous runs do not seem to be accessible anymore after the rebase. How about re-enabling the test for Firefox to get a CI run and investigate a bit?

ui-tests/requirements.txt
```diff
@@ -1,3 +1,4 @@
+https://jupyterlite-pyodide-kernel--183.org.readthedocs.build/en/183/_static/jupyterlite_pyodide_kernel-0.6.0a5-py3-none-any.whl
```
jtpio, line 1: Maybe we can also update the version of `jupyterlite-pyodide-kernel` used in the demo site here with this link too:

https://github.com/jupyterlite/jupyterlite/blob/e17054905a26304d5b8f643af3e34c81d91e590c/examples/jupyter_lite_config.json#L5

So it's easier on the RTD preview for this PR.

packages/server/src/tokens.ts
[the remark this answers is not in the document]
jtpio: Thanks!

ui-tests/test/kernels.spec.ts
```diff
@@ -142,4 +142,60 @@ test.describe('Kernels', () => {
     expect(output).toBeTruthy();
     expect(output![0].trim()).toBe('Final check');
   });
+
+  test('Stdin using pyodide kernel', async ({ page }) => {
```
jtpio, line 146: Looks like the Firefox failure may be because the Pyodide kernel takes some time to start:

image

[jupyterlite-firefox-ui-tests.webm](https://github.com/user-attachments/assets/5f620ed3-2674-4028-81c0-216b639005f9)

So maybe it could be worth trying with a higher timeout like above?

```ts
test.setTimeout(120000);
```

ianthomas23: I'll try it.

ianthomas23: That seems to have worked.

jtpio:
Seems to be working nicely when testing with the ReadTheDocs preview on Firefox 🎉 

https://jupyterlite--1640.org.readthedocs.build/en/1640/_static/lab/index.html


https://github.com/user-attachments/assets/e21d6dac-1585-464c-a3cb-2ab0599a2a8b

packages/application-extension/src/index.tsx
martinRenou: I believe that's fine 👍🏽 This logic of `registerStdinHandler` is not used in the case of Shared Array Buffers, right?

ianthomas23: Correct. The changes for SharedArrayBuffer are in the kernel repo code (xeus and pyodide), in the communciation between the UI thread and web worker. JupyterLite doesn't need to know about that, it just deals with the `input_request` and `input_reply` messages as though they are normal kernel messages.

jtpio (approved):
Thanks for working on this!

Let's get this in and released so it can be tested more easily.

PR #1642, opened 2025-05-06
Toolbar item to show the kernel logs

### LINKED ISSUE #1498: Surface console messages from kernels to the UI
### Problem

Currently it can be difficult for some users to check if there is an error in the console, especially for those not familiar with the dev tools.

### Proposed Solution

Maybe it could be interesting to leverage the `jupyterlab-js-logs` extension for this, which should already do the job for console messages logged in the main window: https://github.com/jupyterlab-contrib/jupyterlab-js-logs

However it looks like it currently does not pick up messages from kernels running in a web worker:

image

Probably the web workers would first have to forward console messages to the main thread via `postMessage`.

### Additional context

Opening this issue in the main JupyterLite repo as this is something that could apply to all kernels in general.

jtpio:
## References

Fixes #1498 

Add a new plugin to make the state of the kernel more visible so users have a better idea if something is broken, based on the comment in https://github.com/jupyterlite/jupyterlite/issues/1498#issuecomment-2759122690.

Showing the ❌ relies on kernels reporting `critical` logs to the log console associated with the notebook, using the JupyterLab `ILoggerRegistry` directly. For the Pyodide kernel: https://github.com/jupyterlite/pyodide-kernel/pull/177

That way JupyterLite does not need to expose new APIs. Instead it gets the information for the `ILoggerRegistry` directly.

## Code changes

- [x] Add a new `@jupyterlite/apputils-extension:kernel-status` to make the kernel status 
- [x] Check for `critical` messages in the notebook logger to report a broken kernel
- [x] Update docs to mention the log console and setting the log level
- [x] UI test

## User-facing changes

**Kernel Status Indicator**

By default, the toolbar item shows 3 states:

https://github.com/user-attachments/assets/4f2ca441-c6c0-4582-aa91-00e27eb9f1b1

**Reporting errors**

If a critical error is reported by the kernel, for example when using the following configuration for the Pyodide kernel used during the kernel initialization process:

```json
"litePluginSettings": {
  "@jupyterlite/pyodide-kernel-extension:kernel": {
    "loadPyodideOptions": {
      "packages": ["unknownpackagetoload"]
    }
  }
}
```

https://github.com/user-attachments/assets/6386c3de-0c33-4a70-8ef3-c1e2115cad24

**Log Console status bar item flashes**

The default status bar item for the log console will flash when new messages are added, but only if they match the current log level:

https://github.com/user-attachments/assets/ef320e7c-c0c7-4842-a4e1-a7dbff3a928f

None


jtpio:
The Pyodide kernel requiring more (optional) tokens actually unveiled a bug in JupyterLab: https://github.com/jupyterlab/jupyterlab/issues/17549

package.json
```diff
@@ -86,5 +86,8 @@
     "typedoc-plugin-markdown": "~4.6.2",
     "typescript": "~5.5.4"
   },
-  "packageManager": "yarn@3.5.0"
+  "packageManager": "yarn@3.5.0",
+  "resolutions": {
+    "@jupyterlab/fileeditor-extension@~4.4.2": "patch:@jupyterlab/fileeditor-extension@npm%3A4.4.2#./.yarn/patches/@jupyterlab-fileeditor-extension-npm-4.4.2-8e26958be6.patch"
```
jtpio, line 91: This will be removed when https://github.com/jupyterlab/jupyterlab/pull/17550 is merged and released upstream.

jtpio:
cc @martinRenou if you would like to have a look, to check how this will affect the current implementation in `jupyterlite-xeus`.

Here is the corresponding PR in `jupyterlite-pyodide-kernel`: https://github.com/jupyterlite/pyodide-kernel/pull/177

Currently JupyterLite does not try to modify the default log level of the log console, so the `info` messages will only appear when the user changes the log level explicitly (since the flashing may become annoying after a while otherwise). But https://github.com/jupyterlab/jupyterlab/issues/11225 should help configure deployments with different defaults.

martinRenou:
Thanks! I'll have a look, I like the new spinner in the toolbar!

jtpio:
Maybe at some point we could also hide the default kernel status indicator (called `executionProgress`) and only show the new one in the toolbar:

image


But for now we could likely leave it there since it displays more information and some users may like it. Otherwise it's possible to remove it from the toolbar via the settings system already.

martinRenou (approved):
Thanks! That looks way nicer than the jupyterlite-xeus approach

juntyr:
@jtpio Is there a way to configure the default level of the log console? I couldn't find a setting in the settings editor (if it was in there it would allow me to override the default for my JupyterLite deployment)

juntyr:
Btw, is it possible to log messages to the Jupyter logger from Python (inside the Pyodide kernel)?

jtpio:
> Btw, is it possible to log messages to the Jupyter logger from Python (inside the Pyodide kernel)?

This implementation does not provide a hook to kernels for doing that, as it may get complicate with other languages.

But I guess you could create a small Jupyter Widget that depends on `ILoggerRegistry`, and writes to it when it receives messages from the kernel?

jtpio:
OK let's get this in and released 🚀

Issue #1651, opened 2025-05-16
/api/stdin is using the same broadcast channel as /api/drive


martinRenou:
## Description

The new stdin feature is making use of the drive `BroadcastChannel` named `/api/drive.v1`.

We should consider either having a specific channel for stdin `/api/stdin.v1`, or change the name to something more general `/api/kernel.v1` (maybe even take the opportunity to go for `/sw-api/kernel.v1`)

martinRenou:
@ianthomas23 @jtpio @trungleduc thoughts about this? Adding it to the wishlist for 0.6.x. We should handle this before the final release IMO, otherwise it will be annoying to rename later.

trungleduc:
I vote for `/sw-api` prefix

trungleduc:
> a general /sw-api/kernel.v1 handling both

This sounds more reasonable.

ianthomas23:
Yes we should do it as soon as possible, it will need changes to `jupyterlite`, `jupyterlite/xeus` and `pyodide-kernel`.

Both the `drive` and `stdin` will also be used by `terminal`, so I don't particularly like having the word `kernel` in the name. And the `stdin` implementation in JupyterLite is generic enough to support arbitrary usage of it by other extensions. It is hard to imagine what those might be, but one day someone will write one.

So I would prefer `/sw-api/drive.v1` and `/sw-api/stdin.v1`.

ianthomas23:
Unless everyone is happier with `/sw-api/kernel.v1` on that basis that this functionality is for kernels but we allow terminals to use it also.

trungleduc:
i don't have strong opinion on path the name

martinRenou:
So how about simply `/sw-api.v1` for the broadcast channel name?

jtpio:
Sticking to a single `BroadcastChannel` to handle the routing of message sounds simpler 👍 Also in case more functionalities need to be added later.

jtpio:
Other options for the name that would still use `/api` as the prefix:

- `/api/sw.v1`
- `/api/serviceworker.v1` so it's more explicit

bollwyvl:
And yet _another_ options would be to have one, unversioned pipe, and push a namespace (which could include the message) into the message. This could be a top-level `$schema` key, rather than some new construct, similar to the  `jupyter_events` scheme.

martinRenou:
@jtpio I see you tagged it 0.6.x milestone, I think we should not, it's a breaking change.

[...]

PR #1654, opened 2025-05-20
Rename service-worker broadcast channel to something more general

martinRenou:
## References

Fix https://github.com/jupyterlite/jupyterlite/issues/1651

I went for `/sw-api.v1` which had the most thumbs up in the discussion

### DIFF
--- a/packages/contents/src/drivefs.ts
+++ b/packages/contents/src/drivefs.ts
@@ -24,7 +24,7 @@ import {
 } from './emscripten';
 
 export const DRIVE_SEPARATOR = ':';
-export const DRIVE_API_PATH = '/api/drive.v1';
+export const DRIVE_API_PATH = '/sw-api.v1';
 
 export const BLOCK_SIZE = 4096;
 
--- a/packages/contents/src/drivefs.ts
+++ b/packages/contents/src/drivefs.ts
@@ -24,7 +24,7 @@ import {
 } from './emscripten';
 
 export const DRIVE_SEPARATOR = ':';
-export const DRIVE_API_PATH = '/sw-api.v1';
+export const SERVICE_WORKER_API_PATH = '/sw-api.v1';
 
 export const BLOCK_SIZE = 4096;
 
--- a/packages/contents/src/drivefs.ts
+++ b/packages/contents/src/drivefs.ts
@@ -24,7 +24,6 @@ import {
 } from './emscripten';
 
 export const DRIVE_SEPARATOR = ':';
-export const SERVICE_WORKER_API_PATH = '/sw-api.v1';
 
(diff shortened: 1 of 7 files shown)


packages/server/src/service-worker.ts
```diff
@@ -6,7 +6,7 @@ const CACHE = 'precache';
 /**
  * Communication channel for drive access
  */
-const broadcast = new BroadcastChannel('/api/drive.v1');
+const broadcast = new BroadcastChannel('/sw-api.v1');
```
jtpio, line 9: Maybe worth also updating the dosctring above, and any other mentions of `drive` to something a bit more generic?

packages/contents/src/drivefs.ts
```diff
@@ -24,7 +24,7 @@ import {
 } from './emscripten';
 
 export const DRIVE_SEPARATOR = ':';
-export const DRIVE_API_PATH = '/api/drive.v1';
+export const SERVICE_WORKER_API_PATH = '/sw-api.v1';
```
jtpio, line 27: Actually, do we need to keep this in `drivefs.ts`? It looks like it's only exported from this module, and used as the name of the `BroadcastChannel`?

jtpio:
> So kernels still make requests to `/api/drive` and it still works, since this hasn't changed.

Unless we want kernels to make requests to `/sw-api.v1/drive` and `/sw-api.v1/stdin` now? And in this case that would indeed require a change in kernels too.

martinRenou:
> Unless we want kernels to make requests to /sw-api.v1/drive and /sw-api.v1/stdin now? And in this case that would indeed require a change in kernels too.

We'd also need to change the service-worker heartbeat url. I'd also be fine keeping things this way for now.

packages/server/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * The service-worker broadcast channel id
+ */
+export const SERVICE_WORKER_BROADCAST_CHANNEL_ID = '/sw-api.v1';
```
jtpio, line 4: Do we need a new file to export this constant? Looks like `tokens.ts` already exports `WORKER_NAME`, so maybe we could also put this const there?

martinRenou: I wanted a safe place for the service-worker to import only simple stuff without pulling the entire world. Turns out the service-worker cannot import from that file without breaking completely, so I'll just get rid of it.

jtpio (approved):
Thanks!

jtpio:
Maybe we can release `0.6.0b1` to have this change out while still in the pre-release cycle.

jtpio:
Released in https://github.com/jupyterlite/jupyterlite/releases/tag/v0.6.0b1

PR #1670, opened 2025-06-04
Introduce IndexedDB-backed StateDB

martinRenou:

### DIFF
--- a/packages/apputils-extension/src/index.tsx
+++ b/packages/apputils-extension/src/index.tsx
@@ -1,6 +1,8 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+import { PageConfig } from '@jupyterlab/coreutils';
+
 import {
   JupyterFrontEnd,
   JupyterFrontEndPlugin,
@@ -22,13 +24,18 @@ import {
   nullTranslator,
 } from '@jupyterlab/translation';
 
+import { IStateDB } from '@jupyterlab/statedb';
+
 import { extensionIcon } from '@jupyterlab/ui-components';
 
 import {
+  IndexedDBDataConnector,
   LiteLicensesClient,
   LitePluginListModel,
   LiteTranslatorConnector,
 } from '@jupyterlite/apputils';
+
+import { ILocalForage } from '@jupyterlite/localforage';
 
 import { kernelStatusPlugin } from './kernelstatus';
 
@@ -154,11 +161,40 @@ const translatorConnector: JupyterFrontEndPlugin<ITranslatorConnector> = {
   },
 };
 
+/**
+ * The state database for storing application state.
+ */
+const statePlugin: JupyterFrontEndPlugin<IndexedDBDataConnector> = {
+  id: '@jupyterlite/apputils-extension:state',
+  description: 'Provides the application state. It is stored per workspaces.',
+  autoStart: true,
+  provides: IStateDB,
+  requires: [ILocalForage],
+  optional: [],
+  activate: (_: JupyterFrontEnd, forage: ILocalForage) => {
+    const baseUrl = PageConfig.getOption('baseUrl');
+    const defaultStorageName = `JupyterLite Storage - ${baseUrl}`;
+    const storageName =
+      PageConfig.getOption('contentsStorageName') || defaultStorageName;
+    const storageDrivers = JSON.parse(
+      PageConfig.getOption('contentsStorageDrivers') || 'null',
+    );
+    const { localforage } = forage;
+
+    return new IndexedDBDataConnector({
+      storageName,
+      storageDrivers,
+      localforage,
+    });
+  },
+};
+
 const plugins: JupyterFrontEndPlugin<any>[] = [
   licensesClient,
   pluginManagerPlugin,
   translatorConnector,
   kernelStatusPlugin,
+  statePlugin,
 ];
 
 export default plugins;
--- a/packages/apputils/src/statedb.ts
+++ b/packages/apputils/src/statedb.ts
@@ -1,0 +1,132 @@
+import { PromiseDelegate } from '@lumino/coreutils';
+
+import { IDataConnector } from '@jupyterlab/statedb';
+
+import type localforage from 'localforage';
+
+/**
+ * The name of the local storage.
+ */
+const DEFAULT_STORAGE_NAME = 'JupyterLite StateDB Storage';
+
+/**
+ * A StateDB data connector backed by IndexedDB
+ */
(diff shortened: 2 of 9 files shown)


martinRenou:
~~I will revisit this. This is not the way to go actually.~~ Done

martinRenou:
It's working great on https://jupyterlite--1670.org.readthedocs.build/en/1670/

Although it seems the UI-tests and Windows 🤔 don't like this.

packages/apputils-extension/src/index.tsx
```diff
@@ -154,11 +166,54 @@ const translatorConnector: JupyterFrontEndPlugin<ITranslatorConnector> = {
...
+  id: '@jupyterlite/apputils-extension:workspace-manager',
+  description: 'The workspace manager plugin.',
+  autoStart: true,
+  provides: IWorkspaceManager,
+  requires: [ILocalForage],
+  optional: [],
+  activate: (_: null, forage: ILocalForage): Workspace.IManager => {
+    const defaultStorageName = 'JupyterLite Storage - Workspaces';
+    const storageName =
+      PageConfig.getOption('contentsStorageName') || defaultStorageName;
```
bollwyvl, line 182: Consider:
- naming this something new (e.g. `workspaceStorageName`) to separate it from `settings` and `contents`
- adding to the schema

packages/apputils-extension/src/index.tsx
```diff
@@ -25,11 +34,14 @@ import {
 import { extensionIcon } from '@jupyterlab/ui-components';
 
 import {
+  WorkspaceManager,
```
bollwyvl, line 37: perhaps `LiteWorkspaceManager` to be consistent with other divergences from upstream.

packages/apputils/src/statedb.ts
```diff
@@ -0,0 +1,146 @@
...
+    const storage = await this.storage;
+
+    // const items = (await storage.keys()).map(async key => await storage.getItem(key)) as any[];
+    const items: { [key: string]: T } = {};
+    for (const key of await storage.keys()) {
+      // Casting to string here, we know for sure the item is there
+      items[key] = (await storage.getItem(key)) as T;
+    }
+
+    const result = Object.keys(items).reduce(
```
bollwyvl, line 89: Perhaps consider reworking as a dumb `for` loop instead of an inline `reduce`.

packages/apputils/src/workspaces.ts
```diff
@@ -0,0 +1,16 @@
+import { ServerConnection, Workspace } from '@jupyterlab/services';
+
+import { IndexedDBDataConnector } from './statedb';
+
+export class WorkspaceManager extends IndexedDBDataConnector<Workspace.IWorkspace> {
+  constructor(options: IndexedDBDataConnector.IOptions) {
+    super({ ...options, storageName: 'JupyterLite Workspaces' });
+
+    this.serverSettings = ServerConnection.makeSettings();
```
jtpio, line 9: Maybe it should accept the `serverSettings?` in its options instead of creating new one? (for consistency, even though it's likely unused)

martinRenou:
@jtpio as you pointed to me, it's indeed due to tests that refresh the page that the CI failure is happening. 

What I see is that if you:
- create a new notebook
- DO NOT SAVE the created notebook
- refresh

The unsaved notebook is properly recovered when recovering the workspace and the layout, but since the notebook was not saved the first time, the page prompts you for choosing a kernel.

I feel like it's the correct behavior, and we should update the tests to save unsaved notebooks before refreshing.

jtpio:
The `win-dev` check does not seem related (also failing on https://github.com/jupyterlite/jupyterlite/pull/1674).

jtpio:
> the CI failure is happening.

This test closes the notebook before the page gets reloaded:

https://github.com/jupyterlite/jupyterlite/blob/53d740bbd2628bdfaadc0f63a05e0d8fb766852f/ui-tests/test/contents.spec.ts#L374

But on the video we see the notebook is restored on startup:

[jupyterlite-close-notebook-refresh.webm](https://github.com/user-attachments/assets/da85626f-151d-49cc-a296-9b6c5ae3e83c)

So maybe that's a real issue with how the state is saved to IndexedDB. It seems to be reproducible with a local build:


https://github.com/user-attachments/assets/271744cc-2c68-4577-8749-6b9a3d91ba20

But it's the same behavior with JupyterLab actually:

https://github.com/user-attachments/assets/3d75dafb-a5cb-4755-8515-49da883fe2a6

Waiting a couple of seconds before refreshing the page seems to be helping though, and the notebook is not restored after the page is reloaded:


https://github.com/user-attachments/assets/bc3de04c-e430-4dbc-b740-445416a3f982

jtpio:
> Waiting a couple of seconds before refreshing the page seems to be helping though, and the notebook is not restored after the page is reloaded:

It's probably because of the use of a `Debouncer` here in the JupyterLab plugin, without specifying a `limit`, so it defaults to 500 ms:

https://github.com/jupyterlab/jupyterlab/blob/4e39df7774cae64b4595ab841236d51b41bcfa2d/packages/apputils-extension/src/index.ts#L426-L431

jtpio:
Out of curiosity, I quickly tried adding the `@jupyterlab/workspace-extension` locally to see if it would be working.

Most of the UI components show up, although the main issue is with the handling of the workspace URL:

https://github.com/user-attachments/assets/21e28d98-f75e-4621-9527-782d39489c2c

So maybe we can leave the PR like this for now, it's already an improvement for restoring layouts! And then look into proper workspace support separately.

martinRenou:
@jtpio thanks a ton for the final push on this, sorry for letting this aside

martinRenou:
I renamed the PR so it reflects a bit more what it does

ui-tests/test/contents.spec.ts
```diff
@@ -100,6 +100,11 @@ test.describe('Contents Tests', () => {
     expect(output).toBeTruthy();
     expect(output![0]).toBe('4');
 
+    await page.notebook.close();
+    // wait for the state to be saved
+    // TODO: find a better way than waiting?
```
jtpio, line 105: Left a comment to check if there could be a way to avoid the `page.waitForTimeout` to get around the lab debouncer.

jtpio (approved):
Thanks @martinRenou!

Looks good to me, and as mentioned in the last comment it's already a nice improvement 👍 

We can look into proper workspace support in a follow-up PR, using `@jupyterlab/workspaces-extension`.

Feel free to have a look at the few commits I pushed, if you would like to review the changes.

martinRenou (commented):
I cannot approve my own PR, but I approve the commits you added on top of it, thanks @jtpio !

PR #1672, opened 2025-06-05
docs: Update deployment instructions for Vercel and Netlify

diegofcornejo:
## Code changes

- Replaced shell hook/activate logic with `micromamba run` to avoid issues in non-interactive shells.
- Removed `eval "$(micromamba shell hook)"` and `micromamba activate` for better portability.
- Updated Python version to 3.12 for better forward compatibility.
- Added `set -e` to fail fast on any error.
- Made `yum install wget` conditional to avoid breaking on systems where `yum` is not available (e.g., Netlify).
- All steps are now more robust and suitable for CI/CD pipelines and ephemeral environments.

### DIFF
--- a/docs/howto/deployment/vercel-netlify.md
+++ b/docs/howto/deployment/vercel-netlify.md
@@ -57,30 +57,30 @@ Then create a new `deploy.sh` file with the following content:
 
 ```bash
 #!/bin/bash
+set -e
 
-yum install wget -y
+# Install wget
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

At some point we could also consider using `uv` or `pixi` as examples for this deployment script, if that can help simplify things a bit.

jtpio:
@meeseeksdev please backport to 0.6.x

PR #1683, opened 2025-07-07
Add basic interrupt functionality

krassowski:
## References

While I ultimately I would love to solve #459, I wonder if we could just cancel the execution of following cells for now?

## Code changes

Cancel execution of following cells.

It would have been cleaner if we could specify cancellation reason in `async-mutex` (https://github.com/DirtyHairy/async-mutex/issues/89)

## User-facing changes

Interruption stops execution of the cells scheduled to execute after the cell currently executing. It does not interrupt execution of the currently executing cell.

| Before | After |
|--|--|
| interrupt does nothing, all cells run | interrupt stops execution of following cells |
| image |image |

None

### DIFF
--- a/packages/kernel/src/client.ts
+++ b/packages/kernel/src/client.ts
@@ -110,13 +110,15 @@ export class LiteKernelClient implements Kernel.IKernelAPIClient {
           ) {
             // expected to throw when mutex.cancel() is called below on cell execution error or on interrupt
             const cancelReason = this._cancelReason.get(mutex);
-            if (cancelReason === 'interrupt') {
+            if (
+              cancelReason === 'interrupt' &&
+              msg.header.msg_type === 'execute_request'
+            ) {
               // Clear cancel reason so that only one cell includes error.
               this._cancelReason.delete(mutex);
               await mutex.waitForUnlock();
               // Send interrupt error to all clients
               const clients = this._kernelClients.get(kernelId) ?? new Set();
-              const date = new Date().toISOString();
               const content: KernelMessage.IReplyErrorContent = {
                 status: 'error',
                 ename: 'Kernel Interrupt',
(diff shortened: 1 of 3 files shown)


jtpio:
Thanks @krassowski for looking into this!

> I wonder if we could just cancel the execution of following cells for now?

That would already be a nice improvement.

Also linking to https://github.com/jupyterlab/jupyterlab/issues/7825 as somewhat related.

krassowski:
@jtpio any chance we could get this one into 0.6.x and aim for a full support of interrupt (which would require new API additions) in 0.7.0?

packages/kernel/src/client.ts
```diff
@@ -107,7 +108,69 @@ export class LiteKernelClient implements Kernel.IKernelAPIClient {
...
+              }
+              sendMessage(
+                KernelMessage.createMessage<KernelMessage.IExecuteReplyMsg>({
+                  channel: 'shell',
+                  session: clientId,
+                  parentHeader: (msg as KernelMessage.IExecuteRequestMsg).header,
+                  msgType: 'execute_reply',
+                  content: {
+                    ...content,
+                    execution_count: 0,
```
krassowski, line 155: This `execution_count` is not used anywhere, but required for `error`; there is also `abort` (and `aborted`?!) but the former at least was deprecated and the latter is not in JupyterLab interfaces.


It is a bit of an open question as to what Jupyter Messaging protocol was meant to be here and what JupyterLab typing should be:
- https://github.com/jupyterlab/jupyterlab/issues/17685
- https://github.com/jupyterlab/jupyterlab/issues/17686
- https://github.com/jupyter/jupyter_client/issues/1063

I think we can keep it as-is here for now.

jtpio:
> any chance we could get this one into 0.6.x and aim for a full support of interrupt (which would require new API additions) in 0.7.0?

If we treat this as a bug I guess it can be backported to 0.6.x? But at the same time it also looks like a new feature.

Otherwise would it be ok to use a 0.7.0 pre-release? Happy to cut a new one once this is merged.

Also, would it be possible to add a new UI test to make sure it doesn't break at some point later? One that checks that clicking on the interrupt button correctly cancels the next cells would be enough (and we could check they don't have any output for instance)

krassowski:
> If we treat this as a bug I guess it can be backported to 0.6.x? But at the same time it also looks like a new feature.
>
> Otherwise would it be ok to use a 0.7.0 pre-release?

A pre-release works for me, I just thought it would be best for other users to benefit from this one before 0.7 (which may take quite some time). I guess we do not need to rush this in the sense that it could go into a pre-release and if everyone is happy with it then it can be backported later.

krassowski:
> would it be possible to add a new UI test to make sure it doesn't break at some point later? One that checks that clicking on the interrupt button correctly cancels the next cells would be enough (and we could check they don't have any output for instance)

Done!

packages/kernel/src/client.ts
```diff
@@ -107,7 +108,69 @@ export class LiteKernelClient implements Kernel.IKernelAPIClient {
...
+                  },
+                  metadata: {
+                    casuse: 'interrupt',
```
jtpio, line 158: Should this be named `cause`?

```suggestion
                    cause: 'interrupt',
```

packages/kernel/src/client.ts
```diff
@@ -195,7 +258,11 @@ export class LiteKernelClient implements Kernel.IKernelAPIClient {
       // to match the JupyterLab behavior
       if (msg.header.msg_type === 'execute_reply') {
         const executeReplyMsg = msg as KernelMessage.IExecuteReplyMsg;
-        if (executeReplyMsg.content.status === 'error') {
+        if (
+          executeReplyMsg.content.status === 'error' &&
+          executeReplyMsg.metadata.casuse !== 'interrupt'
```
krassowski, line 263:
```suggestion
          executeReplyMsg.metadata.cause !== 'interrupt'
```

jtpio:
Thanks @krassowski for working on this!

It seems that clicking on the interrupt button while the kernel is still starting does not cancel the execution of the next cells:

https://github.com/user-attachments/assets/f6a1fed6-0117-49cf-9276-e908bb6f2622

Not sure if that would be possible to address that with the current approach. Otherwise I guess it's still ok to have a look later at it, if it already improves the situation for other cases.

krassowski:
> Not sure if that would be possible to address that with the current approach

It looks like this is because the messages asking for cell execution are only sent after the `kernel_info_request` reply is received, so cancelling mutex earlier - although it already exists - does not cancel messages that were not yet scheduled. One way around it is to await for kernel being ready before proceeding with cancellation. I will give it a try.

krassowski:
It worked, and CI is all green again :)

PR #1688, opened 2025-07-23
use bound instance methods for DriveFSEmscriptenNodeOps

### LINKED ISSUE #1687: Use instance-bound methods for DriveFS classes
### References

- unblocks https://github.com/jupyterlite/pyodide-kernel/issues/199

### Problem

It appears `pyodide 0.28.0` (or `emscripten 4`, bit hard to tell with bundling) uses references to file system node methods, instead of direct calls, so `this` evaluates to `undefined`. So instead of:

```ts
drive.getattr(node)
```

...things like the below are failing:

```ts
const getattr = someFeatureDetection(node);
gettar(whatever);
//  Error: Cannot read properties of undefined (reading 'node')
```

### Proposed Solution

Change the `DriveFSEmscriptenNodeOps` implementation signatures from unbound methods to instance-bound ones:

```ts
// before
getattr(value: IEmscriptenFSNode | IEmscriptenStream): IStats { /* whatever */  }
// after
getattr = (value: IEmscriptenFSNode | IEmscriptenStream): IStats => { /* whatever */  }
```

... then check the _next_ level. Or, defensively make _all_ `DriveFS*` methods instance-bound.

### Additional context

This could _theoretically_ break downstream constructor customization of `DriveFSEmscriptenNodeOps`, but seems unlikely... but this might only be masking _more_ compatibility issues, so it's unclear whether this can target `0.6.x` (preferable) or `main`. Worst case scenario would be to have to overload more of DriveFS in `pyodide-kernel` 😿.

bollwyvl:
## References

- fixes #1687

## Code changes

- defined all methods with `method = (..): ... =>` instead of `method()` on DriveFSEmscriptenNodeOps

## User-facing changes

- n/a

- can potentially impact downstreams that overload this class and use `this.method` in the `constructor`

## Future Work

- [ ] hopefully backport this to 0.6.x #1689

### DIFF
--- a/packages/contents/src/drivefs.ts
+++ b/packages/contents/src/drivefs.ts
@@ -262,25 +262,25 @@ export class DriveFSEmscriptenNodeOps implements IEmscriptenNodeOps {
     this.fs = fs;
   }
 
-  protected node(
+  protected node = (
     nodeOrStream: IEmscriptenFSNode | IEmscriptenStream,
-  ): IEmscriptenFSNode {
+  ): IEmscriptenFSNode => {
     if (instanceOfStream(nodeOrStream)) {
       return nodeOrStream.node;
     }
     return nodeOrStream;
-  }
+  };
 
-  getattr(value: IEmscriptenFSNode | IEmscriptenStream): IStats {
+  getattr = (value: IEmscriptenFSNode | IEmscriptenStream): IStats => {
     const node = this.node(value);
     return {
       ...this.fs.API.getattr(this.fs.realPath(node)),
       mode: node.mode,
       ino: node.id,
     };
-  }
+  };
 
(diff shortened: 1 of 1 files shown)


bollwyvl:
Here's the offending code in the upstream pyodide:

```
stat(path, dontFollow) {
    var lookup = FS.lookupPath(path, {
        follow: !dontFollow
    });
    var node = lookup.node;
    var getattr = FS.checkOpExists(node.node_ops.getattr, 63);
    return getattr(node)
},
```

This looks like it is coming off `emscripten` directly, so would be pretty hard to fix on that side.

bollwyvl:
@jtpio ready for review. I don't think the network issues are related to this PR.

PR #1689, opened 2025-07-23
[0.6.x] use bound instance methods for DriveFSEmscriptenNodeOps

bollwyvl:
## References

- manual backport of #1688

## Code changes

> - defined all methods with `method = (..): ... =>` instead of `method()` on DriveFSEmscriptenNodeOps

## User-facing changes

- n/a

- n/a (maybe)

### DIFF
--- a/py/jupyterlite-core/jupyterlite_core/tests/test_serve.py
+++ b/py/jupyterlite-core/jupyterlite_core/tests/test_serve.py
@@ -43,7 +43,11 @@ def test_serve(an_empty_lite_dir, script_runner, base_url, an_unused_port):  # p
     url = f"http://127.0.0.1:{an_unused_port}{base_url}"
 
     server = subprocess.Popen(args, cwd=str(an_empty_lite_dir))  # noqa: S603
-    time.sleep(2)
+    time.sleep(5)
+
+    # Check if server process is still running
+    if server.poll() is not None:
+        raise RuntimeError(f"Server process exited early with code {server.returncode}")
 
     app_urls = [""]
     for app in ["lab", "tree", "repl"]:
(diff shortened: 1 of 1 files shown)


bollwyvl:
Testing locally (bit of a mess), this gets us _further_, but the two workers _also_ contain a bundled copy of the changed file. Looking to see how to hack in the right content (ha!)... it may also be worth publishing the `.tgz` on ReadTheDocs, as well.

bollwyvl:
Ah, with a hot patch of `node_modules` before building the workers, got all the way up to actually fixing issues against upstream breaking changes: in this case in `micropip` :tada:

bollwyvl:
@jtpio ready for review. I don't think the network issues are related to this PR.

Downstream (no PR yet) I've gotten up to where everything is working again.

jtpio:
Nice, thanks @bollwyvl for looking into this 👍 

I'll have a look, hopefully the failing CI is just transient (or easy to fix).

bollwyvl:
>
> I'll have a look, hopefully the failing CI is just transient (or easy to
> fix).
>

Indeed! But also, no hurry: looks like pyodide 0.28.1 _just_ dropped with
exciting new URL behavior changes, and micropip changes I haven't yet
reviewed. Might also look at filling out more of the pip CLI, I guess, if I
have to do much more there.

jtpio:
From https://github.com/jupyterlite/jupyterlite/issues/1687 

> This could _theoretically_ break downstream constructor customization of `DriveFSEmscriptenNodeOps`, but seems unlikely... but this might only be masking _more_ compatibility issues, so it's unclear whether this can target `0.6.x` (preferable)

If this is mostly for Pyodide 0.28.x and given the changes in that Pyodide release, maybe it would be safer / easier to just start making 0.7.x pre-releases for `jupyterlite-pyodide-kernel`?

bollwyvl:
Welp, _mostly_ for pyodide, because that's what i'm working on. Any emscripten 4 kernel (or other exotic user) would encounter the same issue, as the feature detection comes from the bolierplate generated by emscripten itself.

jtpio:
Closing since https://github.com/jupyterlite/jupyterlite/pull/1688 is in `0.7.0`, so we can focus on `0.7.x` and `0.8.0` pre-releases.

PR #1702, opened 2025-08-12
allow-bubbling-down

DerThorsten:
## References

this is needed for handling keyevents in widgets properly.
Without this, keyboard-event callbacks for widegets like  ipycanvas are not handled properly.
For instance when pressing 'a' a new cell is created, even when the canvas has focus

### DIFF
--- a/app/index.template.js
+++ b/app/index.template.js
@@ -180,6 +180,6 @@ export async function main() {
   }
 
   // 4. Start the application, which will activate the other plugins
-  await app.start();
+  await app.start({ bubblingKeydown: true });
   await app.restored;
 }


krassowski:
This is a heavy setting to change, likely would break accessibility experience. Could `ipycanvas` use `data-lm-suppress-shortcuts` instead (https://jupyterlab.readthedocs.io/en/latest/extension/notebook.html#keyboard-interaction-model)?

DerThorsten:
> This is a heavy setting to change, likely would break accessibility experience. Could `ipycanvas` use `data-lm-suppress-shortcuts` instead (https://jupyterlab.readthedocs.io/en/latest/extension/notebook.html#keyboard-interaction-model)?

but it seems to be the default for vanilla jupyter lab https://github.com/jupyterlab/jupyterlab/blob/0feb1d368a0e241b62118b19f4f38d58a77c7dc1/jupyterlab/staging/index.js#L284

This is also why things like ipycanvas work as expeceted in vanilla jupter.


I did not know about `data-lm-suppress-shortcuts` and will also try to use them in ipycanvas.

brichet:
> but it seems to be the default for vanilla jupyter lab

For reference, we updated it in Jupyterlab in https://github.com/jupyterlab/jupyterlab/pull/14115, to fix the tab trap issue.

krassowski:
> but it seems to be the default for vanilla jupyter lab

Ah, right - thank you please disregard.

jtpio (approved):
Thanks!

Should we expect a big change of behavior w.r.t accessibility? Or should it be considered a bug fix?

To know if this should be backported to 0.6.x, or if it's fine to only have it in the 0.7 pre-releases?

DerThorsten:
> Thanks!
> 
> Should we expect a big change of behavior w.r.t accessibility? Or should it be considered a bug fix?

I dont know tbh

brichet:
> Or should it be considered a bug fix?

I think that it should be considered as a bug fix.

You can try with this snippet in Jupyterlab and Jupyterlite:

```python
from IPython.display import HTML, Javascript
HTML("""
<script>
function onkeypressed(e) {
    if (e.keyCode >= 65 && e.keyCode <= 90) {
        console.log('KEYCODE: ', e.key);
        e.stopPropagation();
    }
}
</script>
<div onkeydown="onkeypressed(event)" tabindex=0>just a div</div>
""")
```

If you first click on the div in the output (to give it focus), typing letter on keyboard will console log with Jupyterlab but not with Jupyterlite.

jtpio:
ok thanks! Let's backport then

PR #1709, opened 2025-08-15
also ignore atime in drivefs

bollwyvl:
## References

- continues #1705

## Code changes

- [x] `DriveFSEmscriptenNodeOps.setattr`
  - [x] ignore `atime`
  - [x] treat `mtime` as `timestamp`

## User-facing changes

Viewing the console won't show hundreds of when programatically touching lots of files:

```
comlink.worker.js:10 setattr mtime of 1755222090987 on FSNode not yet implemented
```

- n/a

## Notes

We probably need to look more deeply at these issues.


bollwyvl:
Actually, even as my description shows, it is `mtime`, not `atime`, which we _also_ don't implement. 

I think we're generally seeing an emscripten 3 vs 4 difference, and probably need to fix it properly.

jtpio (approved):
Thanks!

Indeed we'll probably need to have a closer at it, but for now the change looks fine as is 👍

jtpio:
Closing as handled by https://github.com/jupyterlite/jupyterlite/pull/1781

Issue #1663, opened 2025-06-02
Update to JupyterLab 4.5 and Notebook 7.5


jtpio:
Pre-releases are already available:

- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.5.0a1
- https://github.com/jupyter/notebook/releases/tag/v7.5.0a0

jtpio:
From https://github.com/jupyterlite/jupyterlite/pull/1684#issuecomment-3178304215

> > [ ]  Fix handling of the browser tab title
> 
> Currently the tab title will show this currently:

<img width="3392" height="1942" alt="Image" src="https://github.com/user-attachments/assets/11b8d918-5cee-4a36-bdaa-8743055c990f" />


> This is caused by this logic in JupyterLab, which expects the name of the notebook to be part of the `path`, while JupyterLite does not specify it, or use a URL query string parameter instead: (`?path=example.ipynb`):
> 
> https://github.com/jupyterlab/jupyterlab/blob/0feb1d368a0e241b62118b19f4f38d58a77c7dc1/packages/apputils-extension/src/index.ts#L383-L386

This was fixed in https://github.com/jupyterlab/jupyterlab/pull/17772 and should be fixed here in JupyterLite once we update to JupyterLab 4.5.0a3.

PR #1710, opened 2025-08-15
Update to JupyterLab 4.5.0a3 and Notebook 7.5.0a2

krassowski:
## References

- #1663
- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.5.0a3

## Code changes

None

## User-facing changes

- Adds search in terminals 

None

### DIFF
--- a/ui-tests/requirements.txt
+++ b/ui-tests/requirements.txt
@@ -2,5 +2,5 @@ jupyterlab-language-pack-fr-FR
 jupyterlite-javascript-kernel==0.4.0a0
 jupyterlite-p5-kernel==0.1.1
 jupyterlite-pyodide-kernel==0.7.0a1
-notebook==7.5.0a1
+notebook==7.5.0a2
 theme-darcula==4.0.0


krassowski:
It looks like there is something wrong with the upload button:

| Before | After |
|--|--|
| dark-theme-firefox-linux | dark-theme-firefox-linux |
| tree-firefox-linux | tree-firefox-linux |

Both browsers exhibit this, and it shows up in both light and dark theme.

krassowski:
Is it some odd interaction that requires Notebook to be updated at the same time?

jtpio:
> Is it some odd interaction that requires Notebook to be updated at the same time?

Maybe? Usually we try to update both lab and notebook at once, and deduplicate the packages.

A new notebook was just released, so we can update to it here too now: https://github.com/jupyter/notebook/releases/tag/v7.5.0a2

krassowski:
It looks like it was that indeed. Thank you for the Notebook release! This one is ready now.

PR #1715, opened 2025-08-28
Fix: Drive issue when writing a previously created + deleted file

martinRenou:
## References

Fixing https://github.com/jupyterlite/xeus/issues/273

The issue was that the kernel worker thinks the file still exists (it's not aware that the main thread removed the file) and it still has a in-memory `node` mentioning the existence of the removed file, so when it tries to `open` the removed file, it does it on the outdated `node`. 

The fix is about manually re-creating the file upon `open`, in case the `open` is triggered for writing.

This PR fixes also the case for reading a deleted file. This was also making the kernel crash, instead of properly reporting a FileNotFound error

https://github.com/user-attachments/assets/0da1c079-6c1f-4a91-8c9d-cffeae57dee5

### DIFF
--- a/packages/contents/src/drivefs.ts
+++ b/packages/contents/src/drivefs.ts
@@ -171,14 +171,14 @@ export class DriveFSEmscriptenStreamOps implements IEmscriptenStreamOps {
 
         // if writing
         const flags = stream.flags ?? stream.shared.flags;
-        const accmode = flags & 3; // mask with O_ACCMODE
-        if (
-          accmode === 1 || // O_WRONLY (write only)
-          accmode === 2 || // O_RDWR (read/write)
-          flags & 0x200 || // O_TRUNC
-          flags & 0x400 || // O_APPEND
-          flags & 0x40 // O_CREAT
-        ) {
+        let parsedFlags = typeof flags === 'string' ? parseInt(flags, 10) : flags;
+        parsedFlags &= 0x1fff;
+
+        let needsWrite = true;
+        if (parsedFlags in flagNeedsWrite) {
+          needsWrite = flagNeedsWrite[parsedFlags];
+        }
+        if (needsWrite) {
           stream.node = this.fs.node_ops.mknod(
             stream.node.parent,
             stream.node.name,
--- a/packages/contents/src/drivefs.ts
+++ b/packages/contents/src/drivefs.ts
@@ -98,19 +98,17 @@ type TDriveResponses = {
   lookup: DriveFS.ILookup;
   mknod: null;
   getattr: IStats;
-  get:
-    | {
-        /**
-         * The returned file content
-         */
-        content: any;
+  get: {
+    /**
+     * The returned file content
+     */
+    content: any;
 
-        /**
-         * The content format
-         */
-        format: Contents.FileFormat;
-      }
-    | undefined;
+    /**
+     * The content format
+     */
+    format: Contents.FileFormat;
+  } | null;
   put: null;
 };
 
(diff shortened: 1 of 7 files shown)


packages/contents/src/drivefs.ts
```diff
@@ -98,19 +98,17 @@ type TDriveResponses = {
...
+    /**
+     * The returned file content
+     */
+    content: any;
+
+    /**
+     * The content format
+     */
+    format: Contents.FileFormat;
+  } | null;
```
martinRenou, line 111: That's more homogeneous with the other requests

packages/contents/src/drivefs.ts
```diff
@@ -294,7 +322,14 @@ export class DriveFSEmscriptenNodeOps implements IEmscriptenNodeOps {
           const size = value;
           const path = this.fs.realPath(node);
           if (this.fs.FS.isFile(node.mode) && size >= 0) {
-            const file = this.fs.API.get(path);
+            let file;
+            try {
+              file = this.fs.API.get(path);
+            } catch (e) {
+              // TODO: Should do anything here? Should we create the file?
```
martinRenou, line 329: If you guys don't mind, I'd like that we track this in an issue later. I'm not sure what we should do when `setattr` is called on a removed file. I'd be tempted to create the file but after briefly trying it seems to have some unwanted behaviors.

martinRenou:
@jtpio I don't think the patch approach is working (or I'm not using it correctly?), the latest build on RTD https://jupyterlite--1715.org.readthedocs.build/en/1715/_static/lab/index.html should make the kernel crash on any `open` call (because of the dummy patch https://github.com/jupyterlite/pyodide-kernel/pull/210), but `open` calls do work!

jtpio:
> I don't think the patch approach is working, the latest build on RTD https://jupyterlite--1715.org.readthedocs.build/en/1715/_static/lab/index.html should make the kernel crash on any `open` call (because of the dummy patch [jupyterlite/pyodide-kernel#210](https://github.com/jupyterlite/pyodide-kernel/pull/210)), but `open` calls do work!

Maybe the patch needs to include the lib files too, not just the patch for the TypeScript source.

I used this approach for several PRs across these two repos and it worked fine (although a bit tedious).

martinRenou:
Makes sense, I just pushed a patch for `lib` and trying it out soon

martinRenou:
Thanks @jtpio ! Confirming this works!

https://jupyterlite--1715.org.readthedocs.build/en/1715/try


https://github.com/user-attachments/assets/01391ab5-104f-489f-bc07-82eb83f2d442

ui-tests/test/service-worker.spec.ts
```diff
@@ -104,4 +104,20 @@ test.describe('Service Worker Tests', () => {
...
+
+    // Create the file once
+    await page.notebook.open(notebook);
+    await page.notebook.runCellByCell();
+
+    // Delete it from the UI
+    await page.filebrowser.contents.deleteFile('data.json');
+
+    // Recreate it
+    await page.notebook.runCellByCell();
```
jtpio, line 121: Does this detect any potential error in the cell output, or would it fail silently?

martinRenou: it seems there is no way of doing this with Galata. But now that jupyterlite stops execution on error, I guess `runCellByCell` should be enough to check the notebook runs without error?

martinRenou: Since the test is enough to catch the kernel crashing issue we were seeing, I don't think the test is a blocker to merge the PR.

We could track that in an issue in JupyterLab to add galata APIs for actually checking notebook runs.

jtpio:
> Make a new JupyterLite 0.7 pre-release with this change

Starting it now.

jtpio:
Released: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.7.0a5

Issue #1722, opened 2025-09-09
`DriveFSEmscriptenNodeOps.readlink` should throw `EINVAL` (28) error rather than `EPERM` (63)


ianthomas23:
The `DriveFSEmscriptenNodeOps.readlink` function currently throws an `EPERM` (63) error here:
https://github.com/jupyterlite/jupyterlite/blob/7f7ae499c6df20885936d38659e939a7d9738f45/packages/contents/src/drivefs.ts#L427-L429

I think it should throw an `EINVAL` (28) error instead. The equivalent `MEMFS.readlink(node)` code generated by emscripten is:
```typescript
readlink(node) {
    if (!FS.isLink(node.mode)) {
        throw new FS.ErrnoError(28);
    }
    return node.link;
},
```
which throws a 28 if the `node` is not a link. Also the `FS.readlink(path)` boilerplate code generated by emscripten that calls the various node `readlink` functions is:
```typescript
readlink(path) {
    var lookup = FS.lookupPath(path);
    var link = lookup.node;
    if (!link) {
        throw new FS.ErrnoError(44);
    }
    if (!link.node_ops.readlink) {
        throw new FS.ErrnoError(28);
    }
    return link.node_ops.readlink(link);
},
```
which throws a 28 error if the link node does not implement a `readlink`.

It is possible to demonstrate this using `python` (both pyodide and xeus kernels show the same):

<img width="993" height="512" alt="Image" src="https://github.com/user-attachments/assets/44d91929-fb16-4a16-819c-9f320d8ddb31" />

using the `os.readlink` function. The screenshot shows that for `os.readlink("/drive")` (which is in the DriveFS) returns error 63, and for `os.readlink("/")` (which is not in the DriveFS) returns 28. This code could be the basis for a test.

Although this looks just a cosmetic change, it is in fact an important. `EINVAL` (28) is an expected error in some system calls such as `realpath`:
https://github.com/kraj/musl/blob/e8844e5b0ad417c23e249d1e9def9413631e4352/src/misc/realpath.c#L102C1-L103C1
which allows the system call to continue, whereas an `EPERM` (63) is unexpected and will error out without further functionality running.

I found this via a convoluted route of building `git2cpp` for the JupyterLite `terminal` which allows `git status` to work in a non-DriveFS directory but it fails early in a DriveFS directory.

jtpio:
Closing as fixed by https://github.com/jupyterlite/jupyterlite/pull/1723

PR #1723, opened 2025-09-09
Throw EINVAL (28) in `DriveFSEmscriptenNodeOps.readlink`

ianthomas23:
Throw EINVAL (28) in `DriveFSEmscriptenNodeOps.readlink` rather than EPERM (63).

## References

Fixes issue #1722.

## Code changes

Simple one-line fix.

Added test.

## User-facing changes

No user-facing changes.

No backwards-incompatible changes.

### DIFF
--- a/packages/contents/src/drivefs.ts
+++ b/packages/contents/src/drivefs.ts
@@ -425,7 +425,7 @@ export class DriveFSEmscriptenNodeOps implements IEmscriptenNodeOps {
   };
 
   readlink = (node: IEmscriptenFSNode | IEmscriptenStream): string => {
-    throw new this.fs.FS.ErrnoError(this.fs.ERRNO_CODES['EPERM']);
+    throw new this.fs.FS.ErrnoError(this.fs.ERRNO_CODES['EINVAR']);
   };
 }
 
--- a/examples/jupyter_lite_config.json
+++ b/examples/jupyter_lite_config.json
@@ -2,7 +2,7 @@
   "LiteBuildConfig": {
     "contents": ["."],
     "federated_extensions": [
-      "https://conda.anaconda.org/conda-forge/label/jupyterlite_pyodide_kernel_alpha/noarch/jupyterlite-pyodide-kernel-0.7.0a2-pyh7ca27dd_0.conda",
+      "https://jupyterlite-pyodide-kernel--221.org.readthedocs.build/en/221/_static/jupyterlite_pyodide_kernel-0.7.0a2-py3-none-any.whl",
       "https://conda.anaconda.org/conda-forge/noarch/bqplot-0.12.43-pyhd8ed1ab_0.conda",
       "https://conda.anaconda.org/conda-forge/noarch/ipycanvas-0.13.2-pyhd8ed1ab_0.conda",
       "https://conda.anaconda.org/conda-forge/noarch/ipycytoscape-1.3.3-pyhd8ed1ab_1.tar.bz2",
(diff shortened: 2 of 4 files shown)


ianthomas23:
The new test fails as it is using the pyodide kernel, and for that to work we'd need to be testing against a pyodide-kernel-extension built against the code changed in this PR?

bollwyvl:
> testing against a pyodide-kernel-extension built against the code changed

yep. I haven't yet dug into the dark `webpack`magic to make module federation work correctly inside the workers.

jtpio:
> The new test fails as it is using the pyodide kernel, and for that to work we'd need to be testing against a pyodide-kernel-extension built against the code changed in this PR?

We likely need to do the same as https://github.com/jupyterlite/jupyterlite/pull/1640 or more recently https://github.com/jupyterlite/jupyterlite/pull/1715:

- opening a PR on https://github.com/jupyterlite/pyodide-kernel with the changes, using `jlpm patch`
- link to the dev build for that PR, which is uploaded to ReadTheDocs

That was we can make sure the tests do get run and avoid too many back and forth.

ianthomas23:
Using the wheel from jupyterlite/pyodide-kernel#221 the new UI test passes and the readthedocs site works as intended:

Screenshot 2025-09-10 at 11 21 52

jtpio:
NIce, thanks!

The failing tests seem to be related to the coverage threshold, not sure how this PR is supposed to be related though:

```
FAIL Required test coverage of 82% not reached. Total coverage: 46.25%
```

ianthomas23:
The same coverage failure occurs in another PR jupyterlite/jupyterlite#1725 so presumably there is a change in dependencies that has broken the coverage calculation in some way.

ianthomas23:
`pytest` plugins used in latest CI run that passes:
```
plugins: metadata-3.1.1, anyio-4.10.0, xdist-3.8.0, html-4.1.1, console-scripts-1.4.1, cov-6.3.0
```
and in this PR:
```
plugins: metadata-3.1.1, anyio-4.10.0, cov-7.0.0, xdist-3.8.0, html-4.1.1, console-scripts-1.4.1
```

`pytest-cov 7.0.0` released 2 days ago.

jtpio:
Arf, thanks for looking into this 👍 

Maybe we should just pin for now? Unless it's easy to fix the tests for the new version?

ianthomas23:
I have a fix almost ready...

ianthomas23:
Fix in #1727.

packages/contents/src/drivefs.ts
```diff
@@ -425,7 +425,7 @@ export class DriveFSEmscriptenNodeOps implements IEmscriptenNodeOps {
   };
 
   readlink = (node: IEmscriptenFSNode | IEmscriptenStream): string => {
-    throw new this.fs.FS.ErrnoError(this.fs.ERRNO_CODES['EPERM']);
+    throw new this.fs.FS.ErrnoError(this.fs.ERRNO_CODES['EINVAR']);
```
martinRenou, line 428: Sorry for the wrong error number, this makes much more sense!

Long term (if there is a need for it?) we should have JupyterLab drives support symlinks for real.

ianthomas23:
Rebased after #1730 merged to check that CI passes.

ianthomas23:
Test failure is timing related, must be here:
https://github.com/jupyterlite/jupyterlite/blob/c86eb5cf4f4dcad11def37a5ed7b21a574b82cb8/py/jupyterlite-core/jupyterlite_core/tests/test_cli.py#L136

On my local mac dev machine this tests passes (although various others fail).

jtpio:
> On my local mac dev machine this tests passes (although various others fail).

Yes it looks a bit flaky. Restarted and it passed ✅

jtpio:
There will likely be new alpha releases for the upstream JupyterLab and Jupyter Notebook, so maybe we can wait a bit for these releases to happen, then update JupyterLite to the new versions and finally cut a new JupyterLite pre-release.

Then we'll also be able to update https://github.com/jupyterlite/pyodide-kernel/pull/221 to use newer `@jupyterlite` packages.

PR #1727, opened 2025-09-11
Support pytest-cov 7.0.0

ianthomas23:
The recent release of `pytest-cov 7.0.0` needs an extra config setting of `patch = subprocess` as otherwise coverage of subprocesses is not gathered, see https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html.

I've added the new config setting to `pyproject.toml` (in approximately alphabetical order) rather than create a new `.coveragerc`.

This should fix recent PRs erroneously reporting low coverage percentages.

PR #1730, opened 2025-09-16
Pin pytest-cov <7

ianthomas23:
Pin `pytest-cov <7`. Alternative to #1727 as that is evidently non-trivial to complete.


jtpio:
Opened to track removing the pin: https://github.com/jupyterlite/jupyterlite/issues/1732

PR #1740, opened 2025-09-24
Fix links to local files in Markdown

krassowski:
## References

Follow-up to #1707 

## Code changes

Uses context provided by https://github.com/jupyterlab/jupyterlab/pull/17882

## User-facing changes

Now uploaded images work AND links to uploaded images work.

None


krassowski:
It looks like the `context` need to be threaded down via `AttachmentsResolver` upstream first:

image

jtpio:
There was a GitHub Actions incident yesterday which affected MacOS runners, normally it should be fixed now: https://www.githubstatus.com/incidents/071h21gptcp0

krassowski:
So the tests here highlight an issue with markdown rendering in Firefox. It looks as if the previous (cached?) content was reused instead of the updated cell being rendered. We start with a picture

image

change it to a link

image

run the cell

image

and we get a picture again :exploding_head: 

image

This does not happen in Chrome.

krassowski:
I cannot reproduce this manually. It does look very similar to the thing we saw in https://github.com/jupyterlite/jupyterlite/pull/1744.

jtpio:
Currently we don't have Firefox UI tests for JupyterLab: https://github.com/jupyterlab/jupyterlab/issues/15088

But maybe it could be worth adding a couple of UI tests running on Firefox upstream, for example related to these issues. Just so that in case it's an issue upstream in lab, we could more confidently move forward here in lite and investigate this separately.

krassowski:
Nothing useful in the console logs:

```
+   Object {
+     "location": Object {
+       "columnNumber": 1308201,
+       "lineNumber": 0,
+       "url": "http://localhost:8000/build/jlab_core.7b12f00.js",
+     },
+     "text": "Starting WebSocket: ws://localhost:8000/api/kernels/26a3a807-0222-476b-90bb-30fedb1374d3",
+     "type": "debug",
+   },
+   Object {
+     "location": Object {
+       "columnNumber": 1308201,
+       "lineNumber": 0,
+       "url": "http://localhost:8000/build/jlab_core.7b12f00.js",
+     },
+     "text": "Starting WebSocket: ws://localhost:8000/api/kernels/26a3a807-0222-476b-90bb-30fedb1374d3",
+     "type": "debug",
+   },
+   Object {
+     "location": Object {
+       "columnNumber": 1308201,
+       "lineNumber": 0,
+       "url": "http://localhost:8000/build/jlab_core.7b12f00.js",
+     },
+     "text": "Starting WebSocket: ws://localhost:8000/api/kernels/26a3a807-0222-476b-90bb-30fedb1374d3",
+     "type": "debug",
+   },
+ ]
```

krassowski:
Ok, I see that the `resolveUrl` logic gets hit only once which confirms that the cell does not get re-rendered.

I think this is a direct consequence of https://github.com/jupyterlab/jupyterlab/issues/17941 because the logic for re-rendering Markdown cells [short-circuits if the cell is a placholder](https://github.com/jupyterlab/jupyterlab/blob/8e70045584a7c517841ab37c2782c574c4999519/packages/cells/src/widget.ts#L2543-L2558):

```js
  private _updateRenderedInput(): Promise<void> {
    if (this.placeholder) {
      return Promise.resolve();
    }

    const model = this.model;
    const text =
      (model && model.sharedModel.getSource()) || this._emptyPlaceholder;
    // Do not re-render if the text has not changed.
    if (text !== this._prevText) {
      const mimeModel = new MimeModel({ data: { 'text/markdown': text } });
      this._prevText = text;
      return this._renderer.renderModel(mimeModel);
    }
    return Promise.resolve();
  }
```

jtpio:
@krassowski just checking if you would like to finish that PR before the next JupyterLite pre-release? Or if we should go ahead with a pre-release from the current `main` branch to test the `contentVisibility` mode more easily, and finish that PR later?

krassowski:
I am happy either way. Basically if you wish to cut a pre-release please go ahead. I am traveling tomorrow so may be unable to work on this before Monday. My focus right now is on adding a Firefox test upstream to make sure next lab prerelease fixes the underlying problem.

PR #1742, opened 2025-09-25
Correct use of EINVAL

ianthomas23:
Correct the fix introduced in #1723, it should be `EINVAL` not `EINVAR`.


ianthomas23:
I've confirmed that `jupyterlite 0.7.0a7` containing this fix does indeed work for `git2cpp` in the jupyterlite terminal.

PR #1744, opened 2025-09-29
Update to JupyterLab `4.5.0b0` and Notebook `7.5.0b0`

jtpio:
## References

- https://github.com/jupyterlab/jupyterlab/releases/tag/v4.5.0b0
- https://github.com/jupyter/notebook/releases/tag/v7.5.0b0

## User-facing changes

- `contentVisiblity` windowing mode by default

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -2,8 +2,8 @@ name: jupyterlite-dev
 # this is a conda/mamba environment for working on JupyterLite and upstreams
 
 channels:
-  - conda-forge/label/jupyterlab_alpha
-  - conda-forge/label/notebook_alpha
+  - conda-forge/label/jupyterlab_beta
+  - conda-forge/label/notebook_beta
   - conda-forge
 
 dependencies:
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -2,8 +2,8 @@ name: jupyterlite-docs
 # this is a conda/mamba environment for building the JupyterLite documentation
 
 channels:
-  - conda-forge/label/jupyterlab_alpha
-  - conda-forge/label/notebook_alpha
+  - conda-forge/label/jupyterlab_beta
+  - conda-forge/label/notebook_beta
   - conda-forge
   - nodefaults
 


jtpio:
The Firefox `MathJax` failures are reproducible when running the UI tests locally:

image

But it appears to be working fine with the latest Firefox Dev Edition (`144.0b6`):

image

cc @krassowski could this be related to https://github.com/jupyterlab/jupyterlab/commit/88ef373039a8cc09f27d3814382a512d9033675c ?

krassowski:
> could this be related to https://github.com/jupyterlab/jupyterlab/commit/88ef373039a8cc09f27d3814382a512d9033675c ?

That or `content-visibility` mode. I guess one way to check is to attempt to update `0.6.x` branch with `v4.4.9`?

krassowski:
I cannot repro with Firefox 142.0.1 (64-bit) either (note: CI is using Firefox 141.0)

jtpio:
It could indeed be `content-visibility`.

When running the UI tests locally and earching for `display equation` matches a hidden element:

image

krassowski:
^ That is the cell editor, this is expected. The question is what happened to rendered view.

jtpio:
That I don't know yet, I thought there were multiple matches but actually there is only one for the cell editor yes.

Maybe it was an issue with Firefox 141 that has been fixed in a later version.

krassowski:
Speaking of CodeMirror I run into this (it broke editing in editor):

image

I've never seen this before.

Do CodeMirror versions here require syncing or are those pulled from lab dependencies automatically?

krassowski:
I can reproduce this (broken editor) reliably:
- have something in clipboard
- create a new cell (<kbd>a</kbd>)
- change it to markdown (<kbd>m</kbd>)
- enter edit mode (<kbd>enter</kbd>)
- paste something (<kbd>ctrl</kbd> + <kbd>v</kbd>)

I cannot reproduce this upstream in JupyterLab `4.5.0b0`.

jtpio:
> Do CodeMirror versions here require syncing or are those pulled from lab dependencies automatically?

We don't depend on the `@codemirror` packages here directly, but they are listed in `resolutions` so they end up in the `shared` scope.

krassowski:
I can reproduce the broken editor in JupyterLab when Notebook is installed on Binder (only on Firefox).

jtpio:
Seeing the same issue with the latest `main` (https://jupyterlite.readthedocs.io/en/latest/_static/lab/index.html), but only when using `contentVisiblity` as the windowing mode:

https://github.com/user-attachments/assets/0627e200-9857-47f9-9630-bfc7639c4d68

krassowski:
I have a hunch is relates to cell placeholders handling, I opened https://github.com/jupyterlab/jupyterlab/issues/17941 with my reasoning.

krassowski:
Still no good idea why it would only happen on Firefox and only if Notebook packages are also installed.

jtpio:
> Still no good idea why it would only happen on Firefox and only if Notebook packages are also installed.

I just tried in a brand new environment with just `jupyterlab==4.5.0b0` installed, and while there doesn't seem to be an error logged in the dev tools, the behavior seems very similar:

https://github.com/user-attachments/assets/7387f7ed-342a-4b11-895f-a3504f8be6e3

jtpio:
@krassowski since the Latex rendering issue only seems to be happening with the UI tests, maybe we should skip the test here to move forward with the update and https://github.com/jupyterlite/jupyterlite/pull/1740?

We look into the issue as part of https://github.com/jupyterlab/jupyterlab/issues/17941

krassowski:
Yes it makes sense to move forward

jtpio:
Thanks @krassowski for the review!

PR #1759, opened 2025-10-20
Use Node.js 24 / npm 11 to publish releases

jtpio:
## References

Temporary use node 24 here until the npm 11 issue in the jupyter releaser is fixed and released: https://github.com/jupyter-server/jupyter_releaser/issues/617

This will allow using the npm trusted publishers. See the following issues for more details:
- https://github.com/jupyterlab/jupyterlab/issues/17993
- https://github.com/jupyter/notebook/issues/7745

## Code changes

Update to Node 24 in the publish-release workflow.

## User-facing changes

None

None


jtpio:
The trusted publishers have been configured on npmjs.com.

So we can try this with the next pre-release.

PR #1768, opened 2025-11-03
Update to JupyterLab 4.5.0rc0 and Notebook 7.5.0rc0

jtpio:
## References

Part of https://github.com/jupyterlite/jupyterlite/issues/1663

## Code changes

- [x] Update `@jupyterlab` and `@jupyter-notebook` packages
- [ ] Update conda-deps

## User-facing changes

Some from the upstream releases.

## Backwards-incompatible changes

None

PR #1769, opened 2025-11-04
Update to JupyterLab `4.5.0rc0`, Notebook `7.5.0rc0`, minimum python to 3.10, test on python 3.14

bollwyvl:
## References

- includes #1768 

## Code changes

- [x] remove EOL python 3.9
  - [x] remove references to `importlib_metadata` 
- [x] add python 3.14 to test matrix
- [x] add dependency on `pycparser` on `pypy`
  - > may not be strictly accurate, but easier to reason about than messing with the version of `cffi`
- [x] add `nh3` pin for pypy in build deps
- [ ] bump dev python to 3.13?

## User-facing changes

- users of python 3.9 will not be able to `pip install` the `jupyterlite*` packages without patches

- users of python 3.9 will not be able to `pip install` the `jupyterlite*` packages without patches

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,7 +167,8 @@ jobs:
           jupyter lite list || exit 1
           jupyter lite status --debug || exit 1
       - name: Setup pip (test)
-        run: ${{ matrix.python-command }} -m pip install -r requirements-test.txt --prefer-binary
+        run: |
+          ${{ matrix.python-command }} -m pip install --prefer-binary -r requirements-test.txt
       - name: Pin pytest for pypy
         if: ${{ matrix.python-version == 'pypy-3.10' }}
         run: |


jtpio:
Thanks Nick for helping with this.

Makes me wonder: do we really need to keep testing on `pypy`? Do people use PyPy to build JupyterLite sites?

bollwyvl:
> keep testing on pypy? 

Welp, it's still an order of magnitude faster than `cpython` for many workloads. I think in a production setting (e.g. binderhublite) it would be the right choice, but only if it's not on the end user to test it. Also testing on at least _an_ non-cpython python sometimes finds subtle issues.

It would be lovely to be more easily able to validate some of this locally, but for the pypy case, would have to take a long hard look at using `uv` given the maintenance burden of All The Conda Forge On All The Arches was indeed too much of a burden.

jtpio (approved):
Thanks!

PR #1773, opened 2025-11-12
Document usage of `pyodide-http` with xeus-python

jtpio:
## References

Fixes https://github.com/jupyterlite/xeus/issues/284

Since there is already documentation about this in the main JupyterLite documentation, you can directly consolidate the existing docs.

## Code changes

- [x] Document usage of `pyodide-http` with xeus-python

## User-facing changes

Useful to end users who need to access remote data.

image

None

### DIFF
--- a/docs/howto/content/python.md
+++ b/docs/howto/content/python.md
@@ -133,3 +133,15 @@ pyodide_http.patch_all()
 data = pd.read_csv("https://raw.githubusercontent.com/jupyterlite/jupyterlite/main/examples/data/iris.csv")
 data
 ```
+
+````{tip}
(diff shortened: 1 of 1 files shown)


martinRenou:
Thanks! I have mixed feelings having documentation for jupyterlite/xeus in here. We should make it more discoverable from the jupyterlite/xeus probably.

jtpio:
> Thanks! I have mixed feelings having documentation for jupyterlite/xeus in here. We should make it more discoverable from the jupyterlite/xeus probably.

Yeah, that's right, probably these docs were initially added when there were no proper docs for `jupyterlite-xeus`. Maybe we could move some of the docs from here to the Jupyterlite Xeus repo and just link to them from the main Jupyterlite docs (and same for the Pyodide kernel)

Issue #1629, opened 2025-04-18
Restructure packages


jtpio:
Now that all plugins are on the same app (https://github.com/jupyterlite/jupyterlite/pull/1590) and services plugin in the `packages/services-extension`, it could be useful to look into restructuring and simplifying the `packages`.

Some packages like `@jupyterlite/licences`, `@jupyterlite/translation` and `@jupyterlite/settings` mostly define only one class to override the upstream behavior.

Also since we now have `@jupyterlite/services-extension`, it could make sense to group "services" related components under `@jupyterlite/services`, similar to the upstream `@jupyterlab/services`. To include kernels, kernel specs, sessions, event manager, nbconvert manager...

Maybe `@jupyterlite/localforage` and `@jupyterlite/types` could be folded into an existing package (`type` currently export only 1 interface).

In the end we could end up with a more streamline list of packages:

- `_metapackage`
- `application`: same as now
- `application-extension`: general application plugins
- `apputils`: licenses, translation, service worker
- `apputils-extension`: plugins to expose apputils functionalities
- `notebook-application-extension`: plugins specific to the Notebook application
- `repl-extension`: plugins specific to the REPL application
- `services`: contents, kernels, kernel specs, sessions, events, nbconvert, settings
- `services-extension`: `ServiceManager` plugins
- `ui-components`: icons

Currently kernels depend on:

- `@jupyterlite/contents` for `DriveFS`
- `@jupyterlite/kernel` for registering kernels
- `@jupyterlite/server` for the accessing files via the service worker manager

So changing the structure of packages would require updating kernels to the new package, although this wouldn't be too difficult to do (mostly search / replace).

PR #1777, opened 2025-11-14
Create `@jupyterlite/services` package

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1629

## Code changes

- [x] Add a new `@jupyterlite/services` package that follows the JupyterLab services package structure
- [x] Deprecate the kernel, session and contents packages
- [x] Add to the migration guide
- [x] Update docs

## User-facing changes

None

None, as previous packages are still kept.

We should consider removing these packages in `0.8.0` though.

### DIFF
--- a/docs/reference/contents.md
+++ b/docs/reference/contents.md
@@ -14,15 +14,15 @@ may depend on how kernels are implemented.
 Kernels using Emscripten (like [pyodide](https://github.com/jupyterlite/pyodide-kernel/)
 or [xeus kernels](https://github.com/jupyterlite/xeus/)) relies on the
 [Emscripten filesystem](https://emscripten.org/docs/api_reference/Filesystem-API.html)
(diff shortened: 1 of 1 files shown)


jtpio:
CI is passing, and it appears to be working fine on RTD too.

cc @martinRenou if you would like to have a quick look. Downstreams like `jupyterlite-xeus` could update their imports when this is available, although the `@jupyterlite/kernel`, `@jupyterlite/contents` and `@jupyterlite/server` packages are still kept around for now for backwards compatibility (but deprecated)

jtpio:
Thanks for the review!

> I like that it's just one dependency now

Yes, and also that it matches the `services` / `services-extension` of upstream JupyterLab, so it's a bit easier to know what core plugins JupyterLite replaces.

Let's go ahead with this then. I'll look into making a new pre-release with this (hopefully tomorrow) so we can test it more easily.

PR #1780, opened 2025-11-17
Expose `INbConvertExporters` to register custom exporters

jtpio:
## References

Continuation of #1625 

So third-party extensions can provide custom exporters, such as PDF (https://github.com/jupyterlite/jupyterlite/issues/543).

## Code changes

- [x] Expose `INbConvertExporters`
- [x] Move `LiteNbConvertManager` to the `@jupyterlite/services` package
- [x] Add documentation

## User-facing changes

None

None

### DIFF
--- a/docs/howto/extensions/custom-exporters.md
+++ b/docs/howto/extensions/custom-exporters.md
@@ -124,4 +124,3 @@ export default exporterPlugin;
 
 - [Create a frontend extension](frontend.md)
 - [JupyterLab Extension Tutorial](https://jupyterlab.readthedocs.io/en/stable/extension/extension_tutorial.html)
-- [@jupyterlite/services API Documentation](https://jupyterlite.readthedocs.io/en/latest/api/index.html)


jtpio:
cc @agriyakhetarpal @krassowski since you worked on https://github.com/JupyterEverywhere/jupyterlite-extension, which provides a way to export to PDF.

The idea would be to get this change in the upcoming JupyterLite 0.7 so a basic PDF exporter could be added as a third-party extension in JupyterLite deployments.

Also if you feel like extracting what you have in Jupyter Everywhere to a separate extension, that would be great! The extension could then live under one of the `jupyterlite` / `jupyterlab-contrib` / `JupyterEverywhere` orgs maybe?

krassowski:
I like this direction as it de-risks shipping something as an opt-in which then can then be incrementally improved with a single place where we can contribute to.

jtpio:
Also added some documentation for writing a custom exporter plugin, which could be used to bootstrap such extension.

jtpio:
Moving forward with this so it can be included in the upcoming `0.7.0rc0` release.

We can still iterate more on the API if needed.

PR #1786, opened 2025-11-20
Linter: Enforce type imports when applicable

martinRenou:
## References

This can have the byproduct of reducing worker sizes if we are lucky (if only `import type` is required, we don't pull the JS code for the imported library)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,6 +38,13 @@ module.exports = {
     '@typescript-eslint/no-namespace': 'off',
     '@typescript-eslint/no-var-requires': 'off',
     '@typescript-eslint/no-use-before-define': 'off',
+    '@typescript-eslint/consistent-type-imports': [
+      'error',
+      {
+        prefer: 'type-imports',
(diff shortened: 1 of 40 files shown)


krassowski:
We might consider this in core jupyterlab too at some point.

martinRenou:
> We might consider this in core jupyterlab too at some point.

I was actually going to propose a PR there. It may change quite a lot of code so it may be disruptive with existing PRs, but it may be for the good.

jtpio:
Related: https://github.com/jupyterlab/jupyterlab/issues/17763

martinRenou:
I opened https://github.com/jupyterlab/jupyterlab/pull/18129

martinRenou:
Sounds good for the rc release, I could then share some numbers on how/if it impacts worker sizes

bollwyvl:
Oooh baby, linter rules with a fixes that directly correspond to less runtime code being shipped to the end user! :heart_eyes_cat:

PR #1788, opened 2025-11-20
Remove `SessionContext` workaround

jtpio:
## References

Checking if this workaround is still needed, since https://github.com/jupyterlab/jupyterlab/pull/15958 is now available in JupyterLab 4.5.0.

For reference the workaround was added in https://github.com/jupyterlite/jupyterlite/pull/1227

## Code changes

- [x] Remove `SessionContext` workaround 

## User-facing changes

None

None

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -4,12 +4,11 @@
 import type { JupyterFrontEndPlugin } from '@jupyterlab/application';
 import { ILabShell, IRouter, JupyterFrontEnd } from '@jupyterlab/application';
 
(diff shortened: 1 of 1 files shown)


jtpio:
Looking fine locally with `jupyterlab-github`:

image

jtpio:
This actually seems to be leading to some issues, but only with files at the root of a drive.

Here in the JupyterLab code `dirName` should be `FileSystemAccess:` for the drive to be correctly picked up by the JupyterLite session client:

image

For files in nested directories in a drive:

image

Maybe the easiest would be to keep this workaround for now in JupyterLite, and investigate more upstream what the correct behavior should be.

PR #1782, opened 2025-11-18
Add user-facing changelog for 0.7.0

jtpio:
## References

Part of https://github.com/jupyterlite/jupyterlite/issues/1664

## Code changes

- [x] Add user-facing changelog for the 0.7 release
- [x] Add screenshots / screencasts
- [x] Highlight some upstream improvements

## User-facing changes

Documentation.

## Backwards-incompatible changes

None

PR #1789, opened 2025-11-21
Add version compatibility table to the README

jtpio:
## References

Follow-up to https://github.com/jupyterlite/jupyterlite/pull/1782#issuecomment-3559071292

## Code changes

- [x] Add version compatibility table to the README
- [x] Clean up the README

## User-facing changes

README updates.

None


jtpio:
@bollwyvl probably we don't need to specify the Python version in this compatibility table? As it would otherwise be confusing to users since they may think that it corresponds to the Python version when using the notebook in a browser.

Also, the error will likely be pretty clear if they try to run a `jupyter lite build` on an unsupported Python version.

Issue #1696, opened 2025-08-08
Avoid 404 when fetching `all.json` files


jtpio:
Currently some of the frontend plugins rely on well-known files, for example to know if some contents and files have been indexed and are available as server files.

However this leaves 404 errors in the dev tools console when it's not the case, increasing noise:

<img width="2582" height="68" alt="Image" src="https://github.com/user-attachments/assets/83b7b586-a67d-4120-99d2-5d5b771e0f08" />

Maybe it could be worth looking into using the `PageConfig` to store the information?

The idea would be to dynamically add a new field to the page config, at build time, if some files have been indexed. The frontend plugin could use the information in the page config to know if there is an `all.json` file to fetch.

PR #1804, opened 2025-12-02
Avoid 404 when fetching `all.json` files

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1696

## Code changes

- [x] Populate the name of the `all.json` file (used for indexing contents) as a page config option.
- [x] UI test
- [x] Mention in the migration guide

This way the frontend does not need to hardcode `all.json` anymore, since the name will be generated by the addon at build time.

## User-facing changes

Less noise and 404s in the dev tools console.

**Before**

image

**After**

image

None

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -16,6 +16,29 @@ This change should be transparent for most users. However, if you have custom bu
 configurations that extend or modify the JupyterLite webpack configuration, you may need
 to update your setup to use rspack instead.
 
+### Contents and Workspaces Indexing
+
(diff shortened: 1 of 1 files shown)


jtpio:
cc @agriyakhetarpal since you left a 👍 on the top comment, in case you would like to have a look.

jtpio:
> there won't be 404 errors any longer going forward

Right that's the main reason for this change.

Thanks for the review!

Issue #1231, opened 2023-10-26
Add UI test to check accessing files from a custom drive


jtpio:
https://github.com/jupyterlite/jupyterlite/pull/1227 restored the ability for a lite kernel to access files from a custom drive.

Ideally we should add a test to catch potential regressions that might be happening upstream in JupyterLab, like it did when updating from JupyterLab 3 to JupyterLab 4 (see https://github.com/jupyterlite/jupyterlite/pull/1227 for more context).

Some ideas:

- Implement and use a mock drive that stores files in memory
- Use an existing drive extension
- Use the example from https://github.com/jupyterlab/extension-examples/issues/215 (to be implemented)

bollwyvl:
While a mock would be effective, perhaps there would be value in having an (optional, read-only) IDrive extension that exposed one or more of the following:
- all the actual files in the lite_dir (not just content)
  - at present, the optional `SHA256SUMS` file is the only thing that lists everything
- settings
  - the raw, as-stored settings
  - a derived `overrides.json`
- the file systems of a(ll) kernel(s)

Having these viewable from within the UI _might_ be useful for e.g. _Download_ (especially `overrides.json`) but would certainly tick the box for testing.

jtpio:
> optional, read-only) IDrive extension that exposed one or more of the following:

That would indeed be useful.

https://github.com/jupyterlite/jupyterlite/issues/1041 would also be an option since the drive would likely be included by default (or in the `jupyterlite` metapackage).

PR #1807, opened 2025-12-04
Show JupyterLab and Notebook versions in the about dialog

### LINKED ISSUE #1317: Should the JupyterLite `About` splash display the current JupyterLab version too? (I don't see an obvious way of displaying the JupyterLab version number?)
Should the JupyterLite `About` splash display the current JupyterLab version too? (I don't see an obvious way of displaying the JupyterLab version number?)

image

_Originally posted by @psychemedia in https://github.com/jupyterlite/jupyterlite/issues/1141#issuecomment-1784884272_

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1317
Fixes https://github.com/jupyter/try-jupyter/issues/62

## Code changes

- [x] Add new page config option to specify extra version info that to be displayed in the about dialog
- [x] Add a script to update these values more easily
- [x] UI test

## User-facing changes

More information about the underlying lab and notebook:

image

None


jtpio:
OK that's interesting. The Rspack build fails on windows (only) when using `aboutVersions` as the name: https://github.com/jupyterlite/jupyterlite/actions/runs/22725244110/job/65898367172?pr=1807


```
Run jlpm build:prod
  JupyterLab: 4.6.0-alpha.3 (@jupyterlab/application)
  Jupyter Notebook: 7.6.0-alpha.3 (@jupyter-notebook/application)
Updated aboutVersions in D:\a\jupyterlite\jupyterlite\app\jupyter-lite.json
[@jupyterlite/metapackage]: Process started
[@jupyterlite/metapackage]: Process exited (exit code 0), completed in 4s 455ms
Done in 4s 460ms
[@jupyterlite/app]: Process started
[@jupyterlite/app]: Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues: panicked at crates\rspack_plugin_real_content_hash\src\lib.rs:464:9:
[@jupyterlite/app]: RealContentHashPlugin: circular hash dependency
[@jupyterlite/app]: Process exited (exit code 4294967295), completed in 12s 249ms
The command failed for workspaces that are depended upon by other workspaces; can't satisfy the dependency graph
Failed with errors in 12s 253ms
Error: Process completed with exit code 1.
```

Which looks similar to https://github.com/jupyterlab/jupyterlab/issues/18245#issuecomment-3675392312.

Using something else like `aboutUpstreams` (before) or `versionInfo` (after) makes it pass.

jtpio:
Marking as ready so this could go in one of the next pre-releases.

Long term maybe it could be worth looking into having a proper drive / side panel with more information about the built site users are interacting with, something along the lines of https://github.com/jupyterlite/jupyterlite/issues/1231#issuecomment-1781036828

Or even also make the extension manager partially work with lite.

@bollwyvl what do you think about the current state now?

Also cc @krassowski as the author of https://github.com/jupyter/try-jupyter/issues/62

bollwyvl:
I guess I feel like, from a "what am i using" perspective, the license manager already has all of this data for JS stuff. If anything, we should revisit that UX upstream (it has has some style regressions, to be sure), and maybe consider adding an SBOM output format. This would make a whacking good MIME renderer for that widely-used data format, and is relevant in many ecosystems, not just `npm` (e.g. `uv export --format cyclonedx1.5`) 

This was kind of the motivation for not treating this as "some versions of things hard coded at build time" as magic strings, but rather "a runtime query for existing data that well-behaved apps and extensions should be advertising in `third-party-licenses.json`".

jtpio:
Right the license manager may be a ok place for such information.

Although currently it does not seem to be showing the extension versions consistently:

image

The information about the `@jupyterlab` and `@jupyter-notebook` packages is somewhat available, but not very easy to find.

Maybe this could indeed be improved upstream.

bollwyvl:
Yep, widgets uses an old `jupyterlab-builder` that doesn't build the licenses.

jtpio:
@bollwyvl maybe the current state of the PR could still be a good incremental improvement for the upcoming JupyterLite 0.8.0 release? 

And then we can look into improving the UX upstream in JupyterLab separately?

krassowski (approved):
I think this is worth it already!

PR #1808, opened 2025-12-04
Implement ContentProviderRegistry

martinRenou:
## References

Alternative to https://github.com/jupyterlite/jupyterlite/pull/1745 that will not require some nasty hacks to extract part of the drive implementation in some "default content provider".

Depends on https://github.com/jupyterlab/jupyterlab/pull/18169

Closes #1745


jtpio:
FYI @martinRenou https://github.com/jupyterlite/jupyterlite/pull/1819 updated to JupyterLab 4.5.1, which should help unblock this PR.

martinRenou:
Thanks for rebasing! If you agree with the change, I believe this is ready to go

Issue #1565, opened 2025-01-20
Supplying `--output-dir` can fail on repeated JupyterLite builds when it is different each time


agriyakhetarpal:
## Description

Hi! I came across this weird error: on repeated invocations of the `jupyter lite build` command, it is possible that building JupyterLite can fail at the last step (`build:lite:patch` where the build system patches the `jupyter-lite.json` file).

## Reproduce

```bash
python -m venv venv
source venv/bin/activate
python -m pip install jupyterlite

# initial build, passes
jupyter lite build --output-dir output1 --lite-dir .
# this fails
jupyter lite build --output-dir output2 --lite-dir .
```

<details><summary>Stack trace</summary>
<p>

After building a JupyterLite site in `output1/`, run

```bash
jupyter lite build --output-dir output2 --lite-dir .
```

which then prints out

```
static:jupyter-lite.json
.  pre_status:static:jupyter-lite.json
    tarball:         jupyterlite-app-0.5.0.tgz 14MB
    output:          /Users/agriyakhetarpal/jupyterlite-reproducer/output2
    lite dir:        /Users/agriyakhetarpal/jupyterlite-reproducer
    apps:            
    sourcemaps:      True
    unused packages: True
archive:archive
contents:contents
icons:icons
lite:jupyter-lite.json
mimetypes:jupyter-lite.json
serve:contents
settings:overrides
translation:translation
.  status:archive:archive
[LiteBuildApp] No archive (yet): jupyterlite-reproducer-jupyterlite.tgz
.  status:contents:contents
    contents: 0 files
.  status:icons:icons
    favicon files: 0 files
.  status:lite:jupyter-lite.json
[LiteBuildApp]     jupyter-lite.(json|ipynb): 14 files
.  status:mimetypes:jupyter-lite.json
    filetypes:         26 
.  status:serve:contents
    url: http://127.0.0.1:8000/
    server: stdlib
    headers:
.  status:settings:overrides
    overrides.json: 0
.  status:translation:translation
    translation files: 0 files
static:output_dir
-- pre_init:static:output_dir
static:unpack
.  init:static:unpack
/Users/agriyakhetarpal/jupyterlite-reproducer/venv/lib/python3.12/site-packages/jupyterlite_core/addons/base.py:295: UserWarning: install libarchive-c for better performance when working with archives: No module named 'libarchive'
  return has_optional_dependency(
[...]

Issue #1095, opened 2023-06-14
Build errors when jupyter-lite.json exists in a subdirectory of lite dir


tomjakubowski:
## Description

I found this one while wrapping a Jupyterlite build in Bazel.

If you run `jupyter lite build` in a lite dir with a subdirectory containing `jupyter-lite.json`, the build errors trying to write to the subdirectory in the output dir with `FileNotFoundError: [Errno 2] No such file or directory:`.

## Reproduce

```
mkdir -p lite/sub
touch lite/sub/jupyter-lite.json
cd lite
jupyter lite build
...

translation:copy
.  build:lite:patch:a/jupyter-lite.json
TaskError - taskid:build:lite:patch:a/jupyter-lite.json
PythonAction Error
FileNotFoundError: [Errno 2] No such file or directory: '/Users/tom/tmp/fff/dist/a/jupyter-lite.json'
```

## Expected behavior

I'm not sure, I don't really know much about how configs merge and the like. I imagine it could create the subdirectory as needed in the output directory so it can write the config there, or it could merge it to a root `jupyter-lite.json`. I don't know.

## Context

- JupyterLite version: 0.1.0
- Operating System and version: macOS 13.3.1
- Browser and version: N/A

tomjakubowski:
This is pretty easy to work around in Bazel with the `--lite-dir` flag

jtpio:
Linking to https://github.com/jupyterlite/jupyterlite/issues/1037 as likely related.

Issue #1037, opened 2023-04-04
error when running initial cli build with `jupyterlite-core`


stevejpurves:
## Description

I am initialising and building a new standalone site [as per the docs](https://jupyterlite.readthedocs.io/en/latest/quickstart/standalone.html) but keep hitting an error

```
.  build:lite:patch:_output/jupyter-lite.json
TaskError - taskid:build:lite:patch:_output/jupyter-lite.json
PythonAction Error
Traceback (most recent call last):
  File "/Users/stevejpurves/dev/jlite/env/lib/python3.9/site-packages/doit/action.py", line 461, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/Users/stevejpurves/dev/jlite/env/lib/python3.9/site-packages/jupyterlite_core/addons/base.py", line 240, in merge_one_jupyterlite
    out_path.write_text(json.dumps(config, **JSON_FMT), **UTF8)
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/pathlib.py", line 1285, in write_text
    with self.open(mode='w', encoding=encoding, errors=errors) as f:
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/pathlib.py", line 1252, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/pathlib.py", line 1120, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/stevejpurves/dev/jlite/dist/_output/jupyter-lite.json'
```

## Reproduce

1. Create a clear `venv` `python -m venv install env`
2. activate it
3. `python -m pip install jupyter jupyterlab`
4. `python -m pip install jupyterlite-core`
5. in an empty folder
6. `jupyter lite init`
7. `jupyter lite build --output-dir dist`

```
static:jupyter-lite.json
.  pre_status:static:jupyter-lite.json
    tarball:         jupyterlite-app-0.1.0-rc.0.tgz 7MB
    output:          /Users/stevejpurves/dev/jlite/dist
    lite dir:        /Users/stevejpurves/dev/jlite
    apps:            
    sourcemaps:      True
    unused packages: True
archive:archive
contents:contents
lite:jupyter-lite.json
mathjax:status
mimetypes:jupyter-lite.json
serve:contents
settings:overrides
translation:translation
[...]

Issue #718, opened 2022-07-07
Issue with "Deploy JupyterLite on a standalone server or locally" docs


rickwierenga:
## Description

I am following [this guide](https://jupyterlite.readthedocs.io/en/latest/quickstart/standalone.html#get-an-empty-jupyterlite-site).

`jupyter lite init` gives the following error:

```
...
static:unpack
.  init:static:unpack
lite:patch:_output/jupyter-lite.json
lite:patch:_output/lab/jupyter-lite.json
lite:patch:_output/repl/jupyter-lite.json
lite:patch:_output/retro/jupyter-lite.json
lite:patch:_output/retro/tree/jupyter-lite.json
lite:patch:_output/retro/edit/jupyter-lite.json
lite:patch:_output/retro/consoles/jupyter-lite.json
lite:patch:_output/retro/notebooks/jupyter-lite.json
lite:patch:_output/jupyter-lite.ipynb
lite:patch:_output/lab/jupyter-lite.ipynb
lite:patch:_output/repl/jupyter-lite.ipynb
lite:patch:_output/retro/jupyter-lite.ipynb
lite:patch:_output/retro/tree/jupyter-lite.ipynb
lite:patch:_output/retro/edit/jupyter-lite.ipynb
lite:patch:_output/retro/consoles/jupyter-lite.ipynb
lite:patch:_output/retro/notebooks/jupyter-lite.ipynb
[LiteBuildApp] WARNING | [lite] [translation] `jupyterlab_server` was not importable, cannot create translation data No module named 'jupyterlab_server'
translation:copy
.  build:lite:patch:_output/jupyter-lite.json
TaskError - taskid:build:lite:patch:_output/jupyter-lite.json
PythonAction Error
Traceback (most recent call last):
  File "/Users/rickwierenga/workspace/test/env/lib/python3.9/site-packages/doit/action.py", line 461, in execute
    returned_value = self.py_callable(*self.args, **kwargs)
  File "/Users/rickwierenga/workspace/test/env/lib/python3.9/site-packages/jupyterlite/addons/base.py", line 223, in merge_one_jupyterlite
    out_path.write_text(json.dumps(config, **JSON_FMT), **UTF8)
  File "/opt/homebrew/Cellar/python@3.9/3.9.13_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 1285, in write_text
    with self.open(mode='w', encoding=encoding, errors=errors) as f:
[...]

PR #1817, opened 2025-12-15
Fix handling of `jupyter-lite.json` in sub-directories

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1095
Fixes https://github.com/jupyterlite/jupyterlite/issues/1037
Fixes https://github.com/jupyterlite/jupyterlite/issues/718
Fixes https://github.com/jupyterlite/jupyterlite/issues/1565
Fixes https://github.com/jupyterlite/jupyterlite-sphinx/issues/214

## Code changes

- [x] Fix handling of `jupyter-lite.json` in sub-directories
- [x] Tests

## User-facing changes

Less issues when running `jupyter lite build`.

None

### DIFF
--- a/py/jupyterlite-core/jupyterlite_core/config.py
+++ b/py/jupyterlite-core/jupyterlite_core/config.py
@@ -108,14 +108,14 @@ class LiteBuildConfig(LoggingConfigurable):
     ignore_lite_config: tuple[str] = Tuple(
         help=(
             "Path regular expressions for directories to ignore "
-            "when searching for jupyter-lite.json"
+            "when searching for jupyter-lite.json and jupyter-lite.ipynb"
         )
     ).tag(config=True)
 
     extra_ignore_lite_config: tuple[str] = Tuple(
         help=(
             "Additional path regular expressions for directories to ignore "
-            "when searching for jupyter-lite.json"
+            "when searching for jupyter-lite.json and jupyter-lite.ipynb"
         )
     ).tag(config=True)
 


jtpio:
cc @trungleduc testing locally with `specta`, it does seem to be helping with https://github.com/trungleduc/specta/issues/29. i.e the build works and logs the following:

```
[LiteBuildApp] WARNING | [lite] Skipping .venv/lib/python3.14/site-packages/specta/app_static/specta/jupyter-lite.json: parent directory does not exist in output. Config files should be in the root or in app directories (lab, repl, etc.)
```

jtpio:
cc @bollwyvl as the original author of the handling of these nested `jupyter-lite.json` files.

Does this look like a viable solution to fix all the issues reported by users (linked in the top comment)? Where they mostly run the `jupyter lite build` command without specifying `--lite-dir`.

bollwyvl:
Hm, we know which `apps` have been selected, but not necessarily what paths they will populate.

I guess in the case of a custom app, if this proposed change doesn't fire _after_ the custom app deploys its baseline `jupyter-lite.json` (if it has one, see `voici`), it will get clobbered if those are naively copied.  I think by convention, the first party apps are the same as their name (`tree/`, `notebooks/`), but this probably doesn't hold for all things.

`.rglob`bing `.venv` (or `.pixi` or `.venv`) is brutal in the first place, and we probably need to avoid it.

Perhaps we could limit what it's trying to find, potentially with or or more of:
- a depth limit
  - no built-in way to do this with `rglob`, but not too terrible to add one with iterators
- like contents, a set of ignore patterns which includes the "usual suspects", e.g. `.venv`, `.pixi`, but can be extended by the site owner

Semi-related, I've been thinking about an alternate approach more in line with upstream config with a `jupyter-lite.d/` that _avoids_ merging until the last possible moment (potentially runtime) :

```
_output/
  jupyter-lite.json
  jupyterlite.d/
    __all__.json  # well-known, generated path which will never be copied from downstream
    00-user-put-this-here.json
    01-some-other-addon-put-this-here.json
  lab/
    jupyterlite.d/
      __all__.json  
```

This would be a fairly breaking change, and might have to be opt-in first.

jtpio:
> like contents, a set of ignore patterns which includes the "usual suspects", e.g. .venv, .pixi, but can be extended by the site owner

This sounds like a reasonable approach 👍 

I pushed a commit to implement this, in case you would like to have a look.

bollwyvl:
yeah, change looks good (could mention `jupyter-lite.ipynb`, but reckon that's not used so much).

I could _still_ see folk surprised about how long it takes to `rglob` a huge `.venv` to (generally) not find anything, but i don't have a handy util hanging around as neither `pathlib`, `glob` or `fnmatch` supports this directly...

trungleduc (approved):

jtpio:
Thanks all.

> I could _still_ see folk surprised about how long it takes to `rglob` a huge `.venv` to (generally) not find anything, but i don't have a handy util hanging around as neither `pathlib`, `glob` or `fnmatch` supports this directly...

Yeah let's see if this is an issue in practice, and if so we can always improve that part.

Let's get this in and released then.

PR #1820, opened 2025-12-16
use uri-reference format in configuration schema, use available format checkers during check

bollwyvl:
## References

- n/a

## Code changes

- [x] use the more permissive `uri-reference` in the schema
- [x] use checkers if available during `jupyter lite check`

## User-facing changes

- editing `jupyter-lite.json` with a schema-aware IDE will show fewer squiggles
- more potentially-incorrect fields should work

- the optional `check` step may find more issues that were "working" previously


bollwyvl:
This is fine to merge. uri-ref is a superset of uri, and these are the only
formats we use.

PR #1829, opened 2025-12-23
Fix a broken link to Pyodide documentation

wata727:
Fix a broken link to Pyodide documentation in "Use additional Python packages".
The current valid URL is https://pyodide.org/en/stable/development/building-packages.html


jtpio (approved):
Thanks!

PR #1843, opened 2026-01-20
Drop `lerna`

jtpio:
## References

Removes the Lerna dependency in favor of yarn workspaces and npm (for versioning).

## Code changes

- [x] Switch all inter-package @jupyterlite/* dependencies to use `workspace:^` protocol instead of explicit
  versions
- [x] Replace lerna commands with `jlpm workspaces foreach` in package.json scripts:
- [x] Use `npm pkg set version` to bump JS packages
- [x] Add `@yarnpkg/plugin-workspace-tools`
- [x] Drop `lerna` dev dep
- [x] Remove `update:dependency` script, replaced with `jlpm up`
- [x] Check dev workflow

## User-facing changes

None

None

### DIFF
--- a/scripts/bump-version.py
+++ b/scripts/bump-version.py
@@ -7,6 +7,7 @@
 
 import argparse
 import json
+import sys
 from pathlib import Path
 
 from jupyter_releaser.util import bump_version as bump_py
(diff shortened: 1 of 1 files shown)


bollwyvl:
`yarn.lock Lines changed: 223 additions & 3406 deletions` is a lovely thing to see :joy_cat:

jtpio:
> `yarn.lock Lines changed: 223 additions & 3406 deletions` is a lovely thing to see 😹

 +384 −5,922 now since we can also do without `@jupyterlab/buildutils`.

jtpio:
Thanks for the review 👍 

Added some more logs to the bump version script.

Looks like we should be able to get this in now and tweak a few more things if needed later.

PR #1854, opened 2026-02-05
Add minimal working example of addon

rgaiacs:
## References

Related to https://github.com/jupyterlite/jupyterlite/discussions/1853.

## Code changes

Improve the documentation by providing a minimal working example as it helps me a lot to understand things.

## User-facing changes

Only in the documentation as a hit box was added.

Not applied.

### DIFF
--- a/docs/howto/extensions/cli-addons.md
+++ b/docs/howto/extensions/cli-addons.md
@@ -50,6 +50,10 @@ of sources, the CLI uses a number of layers.
 
 ## Structure of an Addon
 
+```{tip}
+A minimal working example addon is available as part of this documentation on GitHub.
+```
+
 At its very simplest, an Addon is initialized with a signature like:
 
 ```python
--- a/docs/howto/extensions/jupyterlite-mwe-addon/README.md
+++ b/docs/howto/extensions/jupyterlite-mwe-addon/README.md
@@ -1,0 +1,32 @@
+# Minimal working example of addon for JupyterLite
+
+A addon is a Python package.
+
+## Usage with `pixi`
+
+```bash
+pixi run jupyter lite status --debug
+```
+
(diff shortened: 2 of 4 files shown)


jtpio:
Thanks @rgaiacs!

Wondering if it would make sense to have a dedicated repo, similar to https://github.com/jupyterlab/extension-template, where this minimal example addon could be put?

It doesn't have to be a copier template, a simple GitHub template repo could also work.

jtpio:
> Wondering if it would make sense to have a dedicated repo, similar to https://github.com/jupyterlab/extension-template, where this minimal example addon could be put?
> 

Although it may be a bit overkill for the amount of code involved.

rgaiacs:
> Wondering if it would make sense to have a dedicated repo

I can transfer https://github.com/rgaiacs/my-jupyterlite-mwe-addon to the jupyterlite organisation or I can push it to https://github.com/jupyterlite/addon-template if I got push/merge access.

bollwyvl:
I see value in treating this like a real package in the `py/` folder, included in testing/linting, but not published, then hoisted by include to the docs, akin to how `jupyterlab` manages its `examples/`. This would more directly catch API changes.

Also the `-mwe-` acronym, though could could be puzzled out, might be more explainable as `-example-`.

For a truly minimal example, I guess plain-old-`setuptools` would be more minimal.

We'd also want to provide minimal examples of _testing_ an addon. At present, `jupyterlite_core.tests` is a thing, used on some downstream packages, but perhaps that approach should be revisited. 

One alternative would be moving some things from `conftest.py` to `jupyterlite_core.testing`. This would probably only include assertions and fixtures (or fixture precursors), with the the actual `test_*.py` moved up to `tests/` and only included in the `sdist`.

PR #1884, opened 2026-03-04
File provenance UI in the file browser

### LINKED ISSUE #638: Accessing updated version of notebook
This is a question rather than a bug report.

I set up an instance of jupyterlite using GitHub pages and published a notebook. I opened it in my browser and ran a few cells and perhaps made a few edits. But then I realised the published version needed some changes. So I made some changes, committed them to my git repo, and republished it.

However, my browser still shows the local copy it made when I edited the old version - the changes made to the git repo don't show up. How can I delete my local changes and load the updated file?

I tried `File -> Reload Notebook from Disk` and `File -> Revert Notebook to Checkpoint`, but  neither seemed to have the desired effect.

### LINKED ISSUE #694: JupyterLite Notebook versioning
### Problem

Let's say I have a notebook called `example_notebook.ipynb` in a  JupyterLite environment.
The following questions are not easy to answer :

* Did I already edit the notebook, or is it the unedited notebook from the repo?
* Is it the latest version of the unedited notebook, or is it still the outdated version from a few weeks back?
* Is the notebook version provided by the repo where I clicked "try lite now", or is it still present from another repo with a notebook that was also called `example_notebook.ipynb` by coincidence?

Currently, the best way to find this out is to delete `example_notebook.ipynb` and see if it re-appears with a new "Last Modified" entry.  
Two questions regarding this: 
Is there already a way to a separation of JupyterLite workspaces for separate repos? 
Is there an option to reset all notebooks when re-loading the JupyterLite website?

### Proposed Solution
Additional information about the notebook version, e.g like this:

When a notebook is not modified and its the latest version:

When a notebook is modified:

When a new notebook version from the repo is available (here, it does not matter if the older notebook version from the repo was modified or not modified):

### LINKED ISSUE #1468: Delete current notebook option from notebook File menu
I am embedding single Jupyter notebook inside inside instructional materials in a virtual learning environment.

image

One of the attractive fearures of JupyterLite is that whilst edits to notebooks can be save to browser storage, deleting a notebook shipped as part of the browser distribution will reset it to its original distributed value.

Deleting a notebook can be achieved from JupyterLab or notebook tables of contents, but not from the single notebook view.

Naively, it would be useful to have a "File > Close and delete notebook" option to support this behaviour.

More elaborately, it might be noght to have that option listed as  "File > Close and reset notebook" if the notebook is part of the original distribution.

### LINKED ISSUE #1526: UX

jtpio:
*Still a work in progress for now, mostly at a proof-of-concept level still*

## References

Looking into addressing common feedback from users not knowing if a file is a server provided file, or stored in the browser.

Depends on https://github.com/jupyterlab/jupyterlab/pull/18600

Fixes #1526 
Fixes #694 
Fixes #638
Fixes #1468

## Code changes

TBD

## User-facing changes

Final UX still to be defined. Currently looking like the following:

https://github.com/user-attachments/assets/2773996e-f0e8-4c5f-95d8-5a4942fbf0ed

None


agriyakhetarpal:
I like this; it's indeed a common problem. I am aware that this way of displaying file provenance is not final, so some food for thought: one better way, IMO, would be to add another "Location" column, in addition to "Name" and "Modified", with entries of either "Server" or "User".

Another way could be to mark the server-side files and user-defined files with different font colours: for example, similar to how predefined/immutable numbers and user-added numbers look on a sudoku grid. In that case, upon modifying a file, the filename itself would display a tooltip when hovered on.

krassowski:
I think that short of an extra column the other ideas would conflict with the git use case (e.g. colors: red for deleted, grey for untracked, blue for modified; colorful dots: same).

I wonder if this is a good use case for drives. If we had tree-based file browser, could we separate files by provenance by showing them under different roots? Or would that be going against the idea that they form part of the same virtual filesystem hierarchy?

jtpio:
> I wonder if this is a good use case for drives. If we had tree-based file browser, could we separate files by provenance by showing them under different roots? Or would that be going against the idea that they form part of the same virtual filesystem hierarchy?

Yes, this was discussed in some previous issues with one of the ideas being that the browser files would be in one drive and the server files would be in another drive. And if a user wants to start from a server file and have it editable and stored in a browser, they would have to copy it first.

But currently the UX for moving and copying files between drives in JupyterLab is not great which is why this layering system is probably still better for now? But if the files were rendered with a different UI, like a tree view, maybe that would indeed be a bit nicer to work with as a user.

jtpio:
Maybe this could work better:

- using icons to indicate the file browser instead of just colors
- click on the icon to "revert" to the server version

https://github.com/user-attachments/assets/2773996e-f0e8-4c5f-95d8-5a4942fbf0ed

PR #1898, opened 2026-03-18
Drive: Multiple bug fixes

martinRenou:
## Code changes

- Fix saving of files with no extension.
  The PR changes `save` so that it does not call `newUntitled`. This was a mistake since `newUntitled` cannot create a file given a chosen filename, which `save` needs! `newUntitled`'s only purpose is to create a new untitled file in a specific directory, it cannot create a titled file 😏 
- Improve handling of files uploading so that it respects the requested format instead of trying to guess the format from the file extension
- Fix `get` so that it respects the requested format instead of returning the model with the same format as saved on the storage

## User-facing changes

The following code now executes correctly:

image

NONE


ianthomas23 (approved):
Thanks @martinRenou, I've confirmed that this fixes the problem in both kernels and terminals.

jtpio:
Thanks for working on this.

Could we add a small regression UI test to cover this case?

packages/services/src/contents/drive.ts
```diff
@@ -574,43 +558,78 @@ export class BrowserStorageDrive implements Contents.IDrive {
...
-      item = await this.newUntitled({ path, ext, type: 'file' });
-    }
+    const now = new Date().toISOString();
 
-    if (!item) {
-      throw Error(`Could not find file with path ${path}`);
+    let type = options.type || 'file';
+
+    // Yeah sure, we all know notebooks are not files
+    // This is some contents API nonsense
```
jtpio: Maybe worth using a more appropriate wording here, if possible?

jtpio:
Trying it on the RTD preview, should this also work with non-text files?

image

martinRenou:
Good catch, indeed the `write_bytes` example you have seem to not behave correctly. I'll have a look.

martinRenou:
@jtpio thanks for the review and testing! The multipart upload logic was wrongly used in that case. I added a test for it.

packages/services/src/contents/drive.ts
```diff
@@ -36,6 +36,104 @@ const N_CHECKPOINTS = 5;
 const encoder = new TextEncoder();
 const decoder = new TextDecoder('utf-8');
 
+/**
+ * Converts a contents model into JSON
+ *
+ * @param model the model to convert
+ * @returns the converted model
+ */
+export function convertToJSON(model: Contents.IModel): Contents.IModel {
```
jtpio: Do these functions need to be exported? They do not seem to be used outside of this module?

martinRenou: Nope you're right, I will change that. Thanks!

jtpio (approved):
Thanks!

I did a bit of extra manual testing on the RTD preview and it appears to be working fine.

Thanks for adding new tests btw, this will help catch future regressions.

PR #1908, opened 2026-03-26
Drive: Match JupyterLab behavior in `rename` +  force renaming when the request comes from the kernel or terminal

martinRenou:
Context in https://github.com/jupyterlab/jupyterlab/pull/18681

This PR makes the drive behave like JupyterLab when trying to rename a file into a filename that already exists. Meaning `rename` can now fail. For kernels and terminals, renames should be forced (`mv` does not fail when the target exists, nor `os.rename` in Python)

https://github.com/user-attachments/assets/747f5a81-da70-4ee8-8390-82e2400f39b2

TODO in this PR:
- [x] wait for https://github.com/jupyterlab/jupyterlab/pull/18681 merged and released
- [x] then update jupyterlab version
- [x] use `overwrite` when processing emscripten drive `mv` request
- [x] `IDrive`'s rename should fail properly if the target exists, to match jupyterlab's behavior

### DIFF
--- a/packages/services/src/contents/drive.ts
+++ b/packages/services/src/contents/drive.ts
@@ -618,7 +618,7 @@ export class BrowserStorageDrive implements Contents.IDrive {
     const storage = await this.storage;
     const alreadyExists =
       (await storage.getItem(newLocalPath)) !== null ||
-      (await this._getServerContents(path)) !== null;
+      (await this._getServerContents(newLocalPath)) !== null;
 
     if (alreadyExists) {
       // Hacky but JupyterLab assumes an HTTP error
--- a/packages/services/src/contents/drive.ts
+++ b/packages/services/src/contents/drive.ts
@@ -616,7 +616,9 @@ export class BrowserStorageDrive implements Contents.IDrive {
     const name = PathExt.basename(newLocalPath);
 
     const storage = await this.storage;
-    const alreadyExists = (await storage.getItem(newLocalPath)) !== null;
+    const alreadyExists =
+      (await storage.getItem(newLocalPath)) !== null ||
+      (await this._getServerContents(path)) !== null;
 
     if (alreadyExists) {
       // Hacky but JupyterLab assumes an HTTP error
--- a/packages/services/src/contents/drive.ts
+++ b/packages/services/src/contents/drive.ts
@@ -614,13 +614,23 @@ export class BrowserStorageDrive implements Contents.IDrive {
     await this._ensureParentDirectoryExists(newLocalPath);
     const modified = new Date().toISOString();
     const name = PathExt.basename(newLocalPath);
+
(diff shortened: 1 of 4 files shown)


jtpio:
Thanks for looking into this 👍

If there is a way to add a small test to help catch potential future regressions, that would be great, thanks!

jtpio:
It seems this change introduces a new behavior when trying to rename server files, which looks like an issue.

On `main`:

https://github.com/user-attachments/assets/0c6e7f61-ea55-49d4-816d-8cad3007abc6

With this PR:


https://github.com/user-attachments/assets/e5224fb1-519f-4ef0-90a0-6c51248d1db5

martinRenou:
Ah good catch, I passed the wrong parameter to the `_getServerContents` method 😬 https://github.com/jupyterlite/jupyterlite/pull/1908/commits/7c05ea624d44bf43f0da3d77925750ca3c0c63b2 fixes it

martinRenou:
Renaming a server file seems to not delete the old name, I can see that on the current deployment too. Is that the expected behavior from the beginning?

jtpio:
> Renaming a server file seems to not delete the old name, I can see that on the current deployment too. Is that the expected behavior from the beginning?

It's somewhat "expected" mostly because it's been the behavior since the beginning. Maybe this should be changed at some point (https://github.com/jupyterlite/jupyterlite/issues/1955 and related), but likely later in a future release to have more time for testing.

packages/services/src/contents/drive.ts
```diff
@@ -614,13 +614,25 @@ export class BrowserStorageDrive implements Contents.IDrive {
     await this._ensureParentDirectoryExists(newLocalPath);
     const modified = new Date().toISOString();
     const name = PathExt.basename(newLocalPath);
+
+    const storage = await this.storage;
+    const alreadyExists =
+      (await storage.getItem(newLocalPath)) !== null ||
+      (await this._getServerContents(newLocalPath)) !== null;
+
```
jtpio, line 622: Looks like we need drop one of the conditions if we want to keep the current behavior when renaming server files (at least for now)?

```suggestion
    const alreadyExists = (await storage.getItem(newLocalPath)) !== null;

```

jtpio: Maybe this behavior could be changed in the next 0.9.0 release (based on what users expect for instance).

martinRenou: I'm not sure what you mean by this suggestion. I kept the current behavior when renaming server files (which is to keep the original) only that when the target exists, you get warned "Do you want to overwrite it?". If I apply your suggestion, you don't get warned for overwrite.

> this behavior could be changed in the next 0.9.0 release

Agreed, I'm a bit confused by keeping the original name around.

PR #1909, opened 2026-03-26
Test opening audio and video files

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1899

## Code changes

- [x] Add UI tests for opening audio and video files
- [x] Add workaround for now

[video.webm](https://github.com/user-attachments/assets/cd0c7946-2181-493b-9f6c-0dd4c3300101)

[audio.webm](https://github.com/user-attachments/assets/2aacc1e0-3345-474e-afe4-ec7dc5f697fd)

## User-facing changes

None

Potentially avoid future regressions.

None

### DIFF
--- a/packages/services/src/contents/drive.ts
+++ b/packages/services/src/contents/drive.ts
@@ -1,6 +1,6 @@
 import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 
-import type { Contents, Drive } from '@jupyterlab/services';
+import type { Contents, Drive, IContentProvider } from '@jupyterlab/services';
 import { ContentProviderRegistry, ServerConnection } from '@jupyterlab/services';
 
 import type { INotebookContent } from '@jupyterlab/nbformat';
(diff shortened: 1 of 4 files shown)


packages/services/src/contents/drive.ts
```diff
@@ -48,13 +48,15 @@ export class BrowserStorageDrive implements Contents.IDrive {
...
-      });
-    } else {
-      this.contentProviderRegistry = new ContentProviderRegistry();
-    }
+    const contentProviderRegistry = options.defaultContentProvider
+      ? new ContentProviderRegistry({
+          defaultProvider: options.defaultContentProvider,
+        })
+      : new ContentProviderRegistry();
+    const register = contentProviderRegistry.register.bind(contentProviderRegistry);
```
jtpio: cc @martinRenou this should help fix support for audio and videos until we update to the latest jupyterlab to include the fix from https://github.com/jupyterlab/jupyterlab/pull/18652

Maybe we can keep it around for now and open an issue to track its removal.

martinRenou: We could do what you suggested yes. But let's do it only for the "audio" and "video" content providers that we know for sure are not working right now?

I think the current changes from the PR will break the collaboration content provider, and will break JupyterGIS and JupyterCAD lite Python APIs which rely on that.

jtpio: Right. Since we are just a release away from the fix maybe we can actually just keep it as a draft like this for now and wait until the next lab update.

PR #1916, opened 2026-04-01
Update to pyodide-kernel 0.8.0b0

bollwyvl:
## References

- previous attempt https://github.com/jupyterlite/jupyterlite/pull/1893 
- pulls hot wheels from https://github.com/jupyterlite/pyodide-kernel/pull/275
  - forward-port of https://github.com/jupyterlite/pyodide-kernel/pull/269

## Code changes

- [x] `pyproject.toml`
  - [x] update to `pyodide-kernel 0.8.0b0` in a new `dependency-group`
  - [x] update to `jupyter-builder 1.0.0`
- [x] `*/package.json`
  - [x] update to `@jupyter/builder 1.0.0`
- [x] examples
  - [x] use `pyodide-lock`
  - [x] remove a bunch of `%pip`
  - [x] update the `pyhon-packages.ipynb` a bit, just for demo
    - > not on this PR, but might be fun to do a little interactive of the lockfile with some of the demo widget libs

## User-facing changes

- site users will see less `%pip`
- labextension packages should stay better synced with in-kernel python packages 

- n/a


bollwyvl:
Updated to `ipython 9.13.0` with `"excludes": ["psutil"]` ... also proposed an upstream fix on https://github.com/ipython/ipython/pull/15190, but this still demonstrates what having more full control over the env would do.

Also, I ran into a strict `uv` thing, and have put `jupyterlab` back into the `dev` to avoid prerelease mismatches (or having to use the icky global allow).

bollwyvl:
I guess we're ready to start taking a look at this, at least from a functional/build perspective; probably need to do a pass on the demo notebooks and their deps (e.g. `ipycytoscape` no longer works with the version of `networkx` shipped by `pyodide`), as well as some docs work.

jtpio (approved):
Nice, thanks!

PR #1917, opened 2026-04-02
Include `@jupyter/eslint-plugin`

Darshan808:
### Description

Adds `@jupyter/eslint-plugin` as dev dependency and adds eslint rules. I've set all rules to `error`.

### Lint Summary

Total: **33 errors**

- **26 errors** from `jupyter/plugin-description`
- **7 errors** from `jupyter/command-described-by`

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -457,7 +457,7 @@ const notifyCommands: JupyterFrontEndPlugin<void> = {
  */
 const opener: JupyterFrontEndPlugin<void> = {
   id: '@jupyterlite/application-extension:opener',
-  description: 'Opens documents from URL path query parameters.',
+  description: 'Opens docs from URL params.',
   autoStart: true,
   requires: [IRouter, IDocumentManager],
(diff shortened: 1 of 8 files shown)


packages/application-extension/src/index.tsx
```diff
@@ -446,6 +472,15 @@ const opener: JupyterFrontEndPlugin<void> = {
...
     commands.addCommand(command, {
+      describedBy: {
+        args: {
+          type: 'object',
+          properties: {
+            request: { type: 'string' },
+            search: { type: 'string' },
+          },
+        },
+      },
```
Darshan808, line 491: I'm not sure out the properties here!

jtpio: Something like this maybe for their description?

- request: 'The routed URL request with the router base omitted, including the query string and hash if present.'
- search: 'The URL search string containing query parameters, including any `path` values to open.'

Darshan808: Sounds good. Added.

jtpio (approved):
Thanks!

PR #1924, opened 2026-04-10
Do not hard fail upon duplicated plugins

martinRenou:
Duplicated plugins would make the page go blank and never load.

## References

This issue was noticed in Notebook.link, where it's easy to install a package that is already provided by our custom lite build.

This was discussed in [#Ask anything > Opening a link hangs after «Finalizing setup»](https://notebook.zulipchat.com/#narrow/channel/545365-Ask-anything/topic/Opening.20a.20link.20hangs.20after.20.C2.ABFinalizing.20setup.C2.BB/with/584687689)

## Code changes

Log the error instead of showing a blank page.

### DIFF
--- a/app/index.template.js
+++ b/app/index.template.js
@@ -156,7 +156,13 @@ export async function main() {
   const pluginRegistry = new PluginRegistry();
 
   // 2. Register the plugins
-  pluginRegistry.registerPlugins(pluginsToRegister);
+  for (const plugin of pluginsToRegister) {
+    try {
+      pluginRegistry.registerPlugin(plugin);
+    } catch(e) {
+      console.error(e);
+    }
+  }
 
   // 3. Get and resolve the service manager and connection status plugins
   const IServiceManager = require('@jupyterlab/services').IServiceManager;


martinRenou:
@jtpio do you think we could get this in?

jtpio:
> where it's easy to install a package that is already provided by our custom lite build.

Because they are installed on the fly and do not explicitly disable the upstream one when installed?

martinRenou:
> Because they are installed on the fly and do not explicitly disable the upstream one when installed?

Not really, well the issue is more that if you end up with a duplicate of the same extension `@foo/bar`, the page never loads (white screen on stock jupyterlite, spinner on notebook.link), so it's not like `@foo/bar` could explicitly disable `@foo/bar` :D

martinRenou:
> Because they are installed on the fly and do not explicitly disable the upstream one when installed?

Not really, the problem is when you have multiple sources of labextensions, with duplicates in them.

In that case you don't really want to disable any extension. 

The result of duplicated extensions is a blank screen with stock JupyterLite, and an infinite spinner with Notebook.link.

martinRenou:
Oh sorry for the duplicated response, I forgot I had responded and had a tab open with this PR that was not up-to-date :D

jtpio:
Still wondering whether a hard failure may actually be useful to diagnose the issue, as otherwise it may not be clear which of the duplicated plugin ends up being loaded?

martinRenou:
> may actually be useful to diagnose the issue

For devs yeah, for end-users not at all. They are just stuck at a blank screen.

The `console.error` is still useful to diagnose the issue though.

> it may not be clear which of the duplicated plugin ends up being loaded

Yep indeed, first in wins but the order is unclear.

krassowski:
Maybe an alert/error dialog is somewhere in between - user can continue working even if admin/operator messed up with extension versions, and the error is visible enough that they will get notified before user runs into subtle bugs due to version mismatch?

martinRenou:
Makes sense 👍🏽 Unfortunately at this point I guess we can't use `showDialog` from JupyterLab? I will try

PR #1940, opened 2026-05-06
Fix service worker heartbeat bind so that it is called repeatedly

ianthomas23:
## References

Bug found whilst investigating service worker registration (#1939). Service worker heartbeat is only ever sent once whereas is should be sent repeatedly.

## Code changes

The callback `this._pingServiceWorker` is now bound correctly using `this._pingServiceWorker.bind(this)`. Also moved the time between heartbeat pings to be specified once in new `const HEARTBEAT_MS`. These are the same fixes as in jupyterlite/cockle#316.

Also fixed a typo.

## User-facing changes

None. Those who watch the browser network tab will now see more than one heartbeat request, as 20 second intervals.

None.

### DIFF
--- a/packages/apputils/src/service-worker-manager.ts
+++ b/packages/apputils/src/service-worker-manager.ts
@@ -29,6 +29,11 @@ const VERSION = PageConfig.getOption('appVersion');
 const SW_PING_ENDPOINT = '/api/service-worker-heartbeat';
 
 /**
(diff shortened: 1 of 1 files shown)


trungleduc (approved):

martinRenou:
@meeseeksdev please backport to 0.7.x

PR #1942, opened 2026-05-06
Fix getattr's size

### LINKED ISSUE #1941: File `stat` request no longer returns the file size
## Description

If you `stat` a file, JupyterLite used to return the file size via `st_size` as shown here for JupyterLite 0.7.1 (from `try-jupyter`):

Image

With the latest JupyterLite 0.7.5 (from https://jupyterlite.readthedocs.io/) the `st_size` is zero:

Image

This is a big problem for the JupyterLite terminal as some wasm commands such as `git2cpp` have the workflow of `stat`-ing the file size before allocating a buffer of the correct size to read file contents into, hence `git2cpp` is currently unusable in JupyterLite 0.7.5.

## Reproduce

See above for simple reproducer in a lite python kernel.

## Expected behavior

A `stat` request needs to return the file size.

## Context

The file size is returned here:

https://github.com/jupyterlite/jupyterlite/blob/f6c5c470fb3aa243626ee5ae41a0820c4496c48f/packages/services/src/contents/drivecontents.ts#L206

and the `model` is obtained a few lines above:
```ts
const model = await this.contentsManager.get(request.path);
```

I remember something about about a recent-ish change where the file contents used to be returned automatically but aren't any more, so if this is the situation presumably it could explain why the file size is no longer available. I have not tracked down the PR causing this though.

martinRenou:
## References

Fix #1941 `getattr` (stat) requests make use of `contentManager.get(file)`, not `contentManager.get(file, {content: true})`, because we don't need the content of the file. In that case the drive was returning `0` even though we do know the size of the file in the database.

## Code changes

Return the size

### DIFF
--- a/packages/services/src/contents/drive.ts
+++ b/packages/services/src/contents/drive.ts
@@ -551,7 +551,7 @@ export class BrowserStorageDrive implements Contents.IDrive {
     } else {
       return {
         ...model,
-        size: 0,
+        size: model.size ?? 0,
         format: options?.format ?? model.format,
         content: null,
       };
--- a/ui-tests/contents/file-attr-read.ipynb
+++ b/ui-tests/contents/file-attr-read.ipynb
@@ -1,0 +1,41 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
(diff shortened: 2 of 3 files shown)


ianthomas23 (approved):
Thanks @martinRenou

PR #1949, opened 2026-05-09
Document piplite wheel flag examples

### LINKED ISSUE #1563: Document how to pass wheels to --piplite-wheels
Passing containers to trailets is not obvious and in particular passing a single item is failing.

```
$ jupyter lite build  --piplite-wheels ~/dev/there/dist/there-0.0.12-py2.py3-none-any.whl
[...]
traitlets.traitlets.TraitError: The 'piplite_urls' trait of a PipliteAddon instance expected a tuple, not the str '/Users/bussonniermatthias/dev/there/dist/there-0.0.12-py2.py3-none-any.whl'.
[...]
```

This is not a problem when passing the arg multiple time:

```
$ jupyter lite build  --piplite-wheels 1.whl --piplite-wheels 2.whl
```

But that's because I know how traitlets works. 

Maybe it is a bug on traitlets.

Give concrete examples of a adding 1, and many wheels.

musaabhasan:
## Summary

- Add examples for passing a single wheel with `--piplite-wheels`
- Clarify that multiple wheels should be passed by repeating the flag
- Show the equivalent `PipliteAddon.piplite_urls` JSON configuration
- Add a short troubleshooting note for the string-vs-list/tuple error described in #1563

Closes #1563.

## Validation

- `git diff --check`
- Checked fenced code blocks are balanced in `docs/howto/pyodide/wheels.md`
- `python -m sphinx.cmd.build -b html docs docs/_build/html -W --keep-going` reached the changed page but did not complete in this local checkout because generated docs assets/package imports are missing (`../dist`, `../build/docs-app`, generated API docs, and `jupyterlite_core` import setup). No warning was reported for the changed page.

### DIFF
--- a/docs/howto/pyodide/wheels.md
+++ b/docs/howto/pyodide/wheels.md
@@ -16,6 +16,41 @@ Extra wheels that can be installed via `piplite` in a running kernel can be adde
 the `--piplite-wheels` CLI flag or `PipliteAddon/piplite_urls` config value, or simply
 left in-place in `lite_dir/pypi`.
 
+Pass one `--piplite-wheels` flag for each wheel you want to ship. For example, to add
+a single local wheel:
+
+```bash
+jupyter lite build \
+  --piplite-wheels dist/example_package-0.1.0-py3-none-any.whl
+```
+
+To add more than one wheel, repeat the flag:
+
+```bash
+jupyter lite build \
+  --piplite-wheels dist/example_package-0.1.0-py3-none-any.whl \
+  --piplite-wheels dist/example_dependency-0.2.0-py3-none-any.whl
+```
+
+The same values can be supplied in `jupyter_lite_config.json` with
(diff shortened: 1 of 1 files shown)


posita:
Thanks for doing this!

Can a shorter version of how to use this option be included with the output of `--help-all`? That documentation is also currently misleading:

```sh
% jupyter lite --help-all | grep --after-context=3 piplite-wheels
--piplite-wheels=<list-item-1>...
    Local paths or URLs of piplite-compatible wheels to copy and index
    Default: []
    Equivalent to: [--PipliteAddon.piplite_urls]
```

This is a question for current maintainers, but in addition these long-needed documentation updates, should `--piplite-wheels` can be renamed to `--piplite-wheel` (with `--piplite-wheels` being kept as a deprecated synonym)? The plural version suggests it takes multiple arguments (which is the source of a lot of confusion).

PR #1963, opened 2026-05-27
Replace loading indicator with JupyterLab splash-style spinner

MMesch:
## Summary
- Replaces the old yellow circle spinner in the JupyterLite loading indicator with the Jupyter planet logo and orbiting moons, matching the JupyterLab splash screen design
- The SVG icon, orbit dimensions, and planet colors are adapted from the upstream JupyterLab splash so the two look identical and the transition between them is invisible
- The loading indicator now fills the full viewport (like the JupyterLab splash) and retains the "Loading JupyterLite..." label

[Screencast From 2026-05-27 16-21-21.webm](https://github.com/user-attachments/assets/1fa435b5-d512-499b-9353-56ae89a2a6ef)

### DIFF
--- a/app/index.template.html
+++ b/app/index.template.html
@@ -49,73 +49,175 @@
         color: #fff;
       }
 
+      /*
+       * Loading indicator that mirrors JupyterLab's splash screen visual
+       * design for a clean transition. The SVG, orbit dimensions, planet
+       * sizes, and planet colors below are adapted from:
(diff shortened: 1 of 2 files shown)


jtpio (approved):
Thanks!

Having the same spinner make sense and it helps not feel like two separate apps are being loaded 👍 

We can still iterate on it before the final 0.8.0 release if needed.

PR #1967, opened 2026-06-02
Reject contents writes to missing directories

MUFFANUJ:
## References
Refs: 
https://github.com/jupyter-ai-contrib/jupyterlab-ai-commands/issues/23
https://github.com/jupyterlab/plugin-playground/issues/189#issuecomment-4313463320

## Code changes
This fixes JupyterLite content operations so files are not silently created under non-existent parent directories. newUntitled, save, copy, and rename. Now, validate that the target directory exists before writing, which prevents hidden browser-storage entries that the file browser cannot display.

## User-facing changes
None

None


jtpio (approved):
Thanks!

PR #1968, opened 2026-06-08
Remove explicit `workspaces foreach` plugin

Darshan808:
### Description
After the merge and release of https://github.com/jupyterlab/jupyter-builder/pull/109, we no longer need to explicitly provide `workspaces foreach` plugin as the newer `jlpm` already contains it by default.


jtpio (approved):
Thanks!

PR #1972, opened 2026-06-12
Toolbar button to easily share a REPL with predefined state

### LINKED ISSUE #1656: New toolbar button to easily share a REPL with predefined state
### Problem

As a follow-up to https://github.com/jupyterlite/jupyterlite/issues/148, it would be useful to have an additional toolbar button to update the URL with the current state of the REPL, so it includes:

- the code currently in the prompt cell
- the prompt position
- the kernel
- maybe other or all options available via URL parameters?

So folks could:

- open a REPL
- edit the code
- change the layout
- then click on the button to generate a link with the query parameters
- share the link

### Proposed Solution

Maybe the toolbar button could be placed next to the JupyterLite icon:

Image

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1656

## Code changes

- [x] New button to the share the REPL state
- [x] Notify the link has been copied
- [x] UI test

## User-facing changes

- Documentation: https://jupyterlite--1972.org.readthedocs.build/en/1972/quickstart/embed-repl.html#enable-the-toolbar
- Try on RTD: https://jupyterlite--1972.org.readthedocs.build/en/1972/_static/repl/?toolbar=1&kernel=python&theme=JupyterLab+Light

https://github.com/user-attachments/assets/1d9eaea0-595e-4c23-9e5c-dbc09deb5d1f

None


jtpio:
cc @agriyakhetarpal who may be interested in this feature

trungleduc (approved):

agriyakhetarpal (approved):
Neat! I love this!

jtpio:
Thanks for the reviews. Let's get it in then so it's available in the next pre-release.

packages/repl-extension/src/index.ts
```diff
@@ -62,6 +104,135 @@ const buttons: JupyterFrontEndPlugin<void> = {
...
+        if (!config.showBanner) {
+          url.searchParams.set('showBanner', '0');
+        }
+
+        if (promptCode !== '') {
+          url.searchParams.set('execute', '0');
+          promptCode.split('\n').forEach((line) => {
+            url.searchParams.append('code', line);
+          });
+        }
```
krassowski, line 182: In plugin playground we learned that code is best shared via `#` URL fragment rather than query parameters due to RDT and browser length limits, see https://github.com/jupyterlab/plugin-playground/pull/193

jtpio: Thanks! Opened https://github.com/jupyterlite/jupyterlite/issues/1983 to track that. Will try to address before the final 0.8.0 release (unless someone would like to before)

PR #1987, opened 2026-06-18
Extension manager

jtpio:
## References

In JupyterLab the extension manager can be used to browse the installed extensions.

In JupyterLite, it can be tricky to know what extensions are available in a deployment. So this change tries to reuse the existing `@jupyterlab/extension-manager` to serve as a read-only display of installed extensions.

Opening early to explore the approach, which may change if there is something that can be done upstream in JupyterLab to make the implementation more minimal here.

## Code changes

- [x] Enable the extension manager
- [x] Handle `disabledExtensions`?

## User-facing changes

image

None


fleming79:
Somewhat related to this; I've been wondering if it might be possible to dynamically load plugins at runtime by sending the plugin data from the running kernel to the frontend.

The kernel can zip the plugin data from "/share/jupyter/labextensions/" and send the zip file as binary data in a message for the frontend. The frontend could then unpack and load the plugin. 

For comm based plugins, the frontend could request the kernel to send the plugin if the plugin isn't installed (in `_handleCommOpen`). But to support other types of plugins I think the kernel would need to publish that a new plugin has been installed. Perhaps it could be as part of a kernel info message.

jtpio:
Closing in favor of https://github.com/jupyterlite/jupyterlite/pull/2056

PR #1996, opened 2026-07-09
Fix version bump script

jtpio:
## References

Looking into fixing the version bump script which is causing the publish workflow to fail: https://github.com/jupyterlite/jupyterlite/actions/runs/28938314188/job/85854096693

## Code changes

- [x] Fix the version bump script
- [x] Test the script on CI

## User-facing changes

None

None

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,6 +73,10 @@ jobs:
           git diff --exit-code yarn.lock
       - name: Check app resolutions
         run: jlpm integrity:check
+      - name: Check release version bump script
+        run: |
+          python -m pip install --group release
+          python scripts/check-bump-version.py
 
   typedoc:
     runs-on: ubuntu-latest
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -183,6 +183,7 @@ ignore = [
 "docs/conf.py" = ["A001", "S603", "S607"]
 "scripts/dev-build.py" = ["S603"]
 "scripts/bump-version.py" = ["S603"]
+"scripts/check-bump-version.py" = ["S603", "S607"]
 "scripts/upgrade-dependencies.py" = ["S310"]
 "**/conftest.py" = ["S603", "S607"]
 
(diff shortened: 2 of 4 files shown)


bollwyvl:
Just watching the work a bit here (no particular notes!) but am wondering how to make this reusable, potentially in `jupyter_builder`.

I know there are _n_ "version bumpers," but most of them are concerned with a single version-of-interest, and generally don't handle inter/intra repo dependencies very well. Unsurprisingly, I'm imagining something declarative that could live in a one packaging file (e.g. `package.json`, `pyproject.toml`) that captured the rules around both internal and external dependencies, but was a bit more aware of the different package manager conventions.

Ideally, a big repo like Lab, Lite, or Notebook would end up with some kind of set of "policies" which described the patterns between multiple things, and could, with a very minimal CLI (see lab for the _n_ different `update-version` scripts in different languages and myriad switches to all of them):

- check if the current versions are up-to-date across all the files without changing anything
- actually update them, without any implicit magic VCS integration

My original take on this was in TS, but for Jupyter, Python is indeed probably a better place to stand. I guess the challenge is describing the business logic between things, accounting for crazy exceptions.

jtpio:
Right it would be great if all these projects would share the same kind of small library for bumping versions across the monorepos.

`@lerna-lite/version` sounded like what we would need, but it still seems to be pulling many dependencies, as mentioned in https://github.com/jupyterlab/jupyterlab/issues/18437#issuecomment-3881126402

bollwyvl:
Yeah, `lerna-lite` probably wouldn't do what we need, and is still fraught due to the limitations of the node stdlib.

However, we certainly _would_ want the canonical `semver` JS package, which blessedly has no further deps. 

As I don't think we'd want to do _any_ regex-based file stuff, the worst thing would be structurally updating `pyproject.toml`: on the python side we could haul in `tomlkit`, as e.g. `tomli-w` doesn't do the thing. If we wanted to do things for e.g. `conda-forge`, `helm`, GHA, etc. we'd also need a YAML parser/writer, so probably `ruamel.yaml`.

With those in mind, I feel like `jupyter_builder` but under a `[versioning]` extra might be the right place.

I don't yet have a clear picture of what a policy would look like yet... lab is the long pole for intra-repo deps, with rules that are "true for everything except one test package." Lite is probably the worst total mix of external and internal issues.

jtpio:
OK, let's look into this in https://github.com/jupyterlab/jupyter-builder/issues/146 then.

In the meantime I'll try the current approach with this PR and see if it helps get the first 0.9.0 pre-release out.

PR #2000, opened 2026-07-16
Add an optional site drive

### LINKED ISSUE #1231: Add UI test to check accessing files from a custom drive

jtpio:
## References

Fixes https://github.com/jupyterlite/jupyterlite/issues/1231

This will also be useful to validate https://github.com/jupyterlite/jupyterlite/pull/1998

## Code changes

- [x] Add a new custom drive to show some of the site config files for now
- [x] Keep disabled on the demo site because of the upstream issue: https://github.com/jupyterlab/jupyterlab/issues/15629
- [x] UI test

## User-facing changes

When enabled:

image

None

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
 
   # Check ruff version is aligned in pyproject.toml
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.10
+    rev: v0.16.0
     hooks:
       - id: ruff
         args: ['--fix']
--- a/docs/howto/configure/simple_extensions.md
+++ b/docs/howto/configure/simple_extensions.md
@@ -111,6 +111,7 @@ which translates to:
 
 ```py
 import piplite
+
 await piplite.install(["ipywidgets", "bqplot", "plotly"])
 ```
 
@@ -142,6 +143,7 @@ which translates to:
 
 ```py
 import piplite
+
 await piplite.install(["ipywidgets==7.7.0", "bqplot==0.12.30", "plotly==5.8.0"])
 ```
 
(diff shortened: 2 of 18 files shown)


jtpio:
To validate the new drive related test, https://github.com/jupyterlite/jupyterlite/pull/2000/commits/b0376a50af82d29118a0b6e9feaf7013bb0cbb25 does the same as https://github.com/jupyterlite/jupyterlite/pull/1998 and the new test fails as expected:

image

The next commit (https://github.com/jupyterlite/jupyterlite/pull/2000/commits/7977e63a6116fd4e70d491b48e7702100afec058) reverts that change and CI passes again.

jtpio:
cc @bollwyvl do you think this (opt-in) default drive could already be useful? Since you had some ideas https://github.com/jupyterlite/jupyterlite/issues/1231#issuecomment-1781036828

We can always add more things to it over time. But at least it seems to be helping with the original motivation of https://github.com/jupyterlite/jupyterlite/issues/1231 and being able to validate the file system access for custom drive.

jtpio:
OK I'll go ahead to make progress with https://github.com/jupyterlite/jupyterlite/pull/1998 and the upstream patch: https://github.com/jupyterlab/jupyterlab/pull/19157

It's still early in the 0.9.0 pre-release cycle so happy to iterate on that debug drive more.

PR #2026, opened 2026-08-14
Fix schema paths on Windows

agoudbg:
## References

None.

## Code changes

Use a platform-aware relative path when generating schema plugin IDs, preventing Windows absolute paths from being written to `all.json`.

## User-facing changes

Settings load correctly in JupyterLite apps built on Windows.

None.


jtpio (approved):
Thanks!

PR #2034, opened 2026-08-24
Fix Notebooks download for drives that give Notebooks in base64 format

martinRenou:
## References

The Notebook.link drive gives Notebooks as base64 unless it's specifically requested to provide JSON format.

This PR makes the download feature take that into `format` into account instead of assuming json format

### DIFF
--- a/packages/application-extension/src/index.tsx
+++ b/packages/application-extension/src/index.tsx
@@ -362,8 +362,12 @@ const downloadPlugin: JupyterFrontEndPlugin<void> = {
         model.format === 'json' ||
         model.mimetype === 'text/json'
       ) {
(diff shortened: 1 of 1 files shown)


packages/application-extension/src/index.tsx
```diff
@@ -362,8 +362,12 @@ const downloadPlugin: JupyterFrontEndPlugin<void> = {
         model.format === 'json' ||
         model.mimetype === 'text/json'
       ) {
+        let filecontent = model.content;
+        if (model.format === 'base64') {
+          filecontent = JSON.parse(atob(model.content));
```
jtpio, line 367: Does this work when the notebook contains non-ASCII characters?

PR #2036, opened 2026-09-02
Fix directory recreation bug

Yahiewi:
## References

This fixes an issue where if you create a directory from the terminal (using cockle/jupyterlite-terminal), then delete it through the file browser UI, trying to recreate it in the terminal throws an error and causes the shell to hang indefinitely.
A very similar bug happens when using a Python notebook instead of the terminal to create directories.
This PR has two separate fixes for both cases.

## Code changes

- **`DriveFS._hookMayCreate`**: Wrapped Emscripten's internal `FS.mayCreate` to intercept `EEXIST` errors on `DriveFS` mounts. If Emscripten's in-memory node cache (`FS.nameTable`) reports `EEXIST` even though the path no longer exists on the drive, this stale node is purged via `FS.destroyNode` and creation is retried.
- **`DriveContentsProcessor.getattr`**: Wrapped the logic in a `try...catch` block to catch missing path rejections from `contentsManager.get` to return `null` instead of an unhandled Promise rejection (which is what previously caused the terminal to hang indefinitely). 
  Also fixed an operator precedence bug for the `blocks` calculation:
  - Before: `blocks: Math.ceil(model.size || 0 / BLOCK_SIZE)` (evaluated as `model.size || 0` since `/` has higher precedence than `||`)
  - After: `blocks: Math.ceil((model.size || 0) / BLOCK_SIZE)`
- **`ContentsAPI.getattr`**: Updated to throw POSIX `ENOENT` when the drive responds with `null`.
- **`emscripten.ts`**: Added typed definitions for Emscripten's internal `destroyNode`, `lookupNode`, and `mayCreate` methods on `FS`.

## User-facing changes

Users can now recreate directories in kernels (for example using Pyodide and `os.mkdir()` in a Python notebook) or cockle terminal sessions without running into `FileExistsError: [Errno 20] File exists` after having deleted the directory from the JupyterLite UI file browser.

None.

## Screencast

[Screencast From 2026-09-01 18-11-27.webm](https://github.com/user-attachments/assets/0a21dc3d-c58d-4d10-97fb-7fda12bdc071)

### DIFF
--- a/packages/services/src/contents/drivecontents.ts
+++ b/packages/services/src/contents/drivecontents.ts
@@ -192,7 +192,13 @@ export class DriveContentsProcessor implements IDriveContentsProcessor {
   }
 
   async getattr(request: TDriveRequest<'getattr'>): Promise<TDriveResponse<'getattr'>> {
-    const model = await this.contentsManager.get(request.path);
+    let model: Contents.IModel;
+    try {
+      model = await this.contentsManager.get(request.path);
+    } catch {
+      return null;
+    }
+
     // create a default date for drives that send incomplete information
     // for nested foldes and files
     const defaultDate = new Date(0).toISOString();
@@ -205,7 +211,7 @@ export class DriveContentsProcessor implements IDriveContentsProcessor {
       rdev: 0,
       size: model.size || 0,
       blksize: BLOCK_SIZE,
-      blocks: Math.ceil(model.size || 0 / BLOCK_SIZE),
+      blocks: Math.ceil((model.size || 0) / BLOCK_SIZE),
       atime: model.last_modified || defaultDate, // TODO Get the proper atime?
       mtime: model.last_modified || defaultDate,
       ctime: model.created || defaultDate,
--- a/packages/services/src/contents/drivefs.ts
+++ b/packages/services/src/contents/drivefs.ts
@@ -101,7 +101,7 @@ type TDriveResponses = {
   getmode: number;
   lookup: DriveFS.ILookup;
   mknod: null;
-  getattr: IStats;
+  getattr: IStats | null;
   get: {
     /**
      * The returned file content
@@ -567,6 +567,10 @@ export abstract class ContentsAPI {
       path: this.normalizePath(path),
     });
 
+    if (!stats) {
+      throw new this.FS.ErrnoError(this.ERRNO_CODES['ENOENT']);
+    }
+
     // Emscripten 4.0.9+ (used by Pyodide 0.28+) requires all three timestamps
     // to be valid Date objects with .getTime() method (see https://github.com/emscripten-core/emscripten/pull/22998).
     // Fallback to epoch if any timestamp is missing/null/undefined.
@@ -677,6 +681,8 @@ export class DriveFS {
 
     this.node_ops = new DriveFSEmscriptenNodeOps(this);
     this.stream_ops = new DriveFSEmscriptenStreamOps(this);
+
+    this._hookMayCreate();
   }
 
   node_ops: IEmscriptenNodeOps;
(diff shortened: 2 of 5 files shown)


packages/services/src/contents/drivecontents.ts
```diff
@@ -205,7 +211,7 @@ export class DriveContentsProcessor implements IDriveContentsProcessor {
       rdev: 0,
       size: model.size || 0,
       blksize: BLOCK_SIZE,
-      blocks: Math.ceil(model.size || 0 / BLOCK_SIZE),
+      blocks: Math.ceil((model.size || 0) / BLOCK_SIZE),
```
martinRenou, line 214: 👀

martinRenou (commented):
Thanks! Do you think we could add a test for it?

packages/services/src/contents/drivefs.ts
```diff
@@ -734,6 +740,35 @@ export class DriveFS {
...
+   */
+  private _hookMayCreate(): void {
+    const FS = this.FS;
+    const { destroyNode, lookupNode, mayCreate } = FS;
+
+    if (!destroyNode || !lookupNode || !mayCreate) {
+      return;
+    }
+
+    FS.mayCreate = (dir: IEmscriptenFSNode, name: string): number => {
```
martinRenou, line 756: Is this called for each filesystem operation?

It worries me that we're calling `API.lookup` there as it would add a lot of overhead on the filesystem if we do it for each operation (basically multiplying by two all API calls over the service-worker)

Yahiewi: So `FS.mayCreate` as defined in [Emscripten](https://github.com/emscripten-core/emscripten/blob/7afdae44d67ca8169d4c0f30a0d62092a54e23db/src/lib/libfs.js) is only called when creating new nodes (`mkdir`, `mknod`, `symlink`, etc.) and if you look at the `if` block:
```
      if (
        errCode !== this.ERRNO_CODES['EEXIST'] ||
        dir.node_ops !== this.node_ops ||
        this.API.lookup(this.PATH.join2(this.realPath(dir), name)).ok
      ) {
        return errCode;
      }
```
You can see that thanks to short-circuit evaluation in JS, the first true condition in the `OR` chain ignores the remaining conditions. So the `API.lookup` is only executed when a creation fails with `EEXIST` on a DriveFS directory (which is the bug we're getting) to verify whether or not the node is actually present.

Yahiewi:
> Thanks! Do you think we could add a test for it?

I don't think we can really add a test for the terminal bug since that requires cockle+jupyterlite-terminal but I guess I can add the notebook test shown in the screencast to ui-tests. I'll do that in a sec

packages/services/src/contents/drivefs.ts
martinRenou: Makes sense, thanks for the clarification

martinRenou (approved):
Thanks for working on this!

Yahiewi:
Thanks for the review Martin. I added a playwright test (similar to the one in the screencast) and it took me a while mainly because it's such a pain to test this because of the pyodide-kernel which consumes `@jupyterlite/services `(which is what we're modifying). So for the CI to pass, jupyterlite-pyodide-kernel would need the latest `@jupyterlite/services` (after this is merged and released).
To test this, I patched the locally built pyodide-kernel to use the updated jupyterlite/services and the test passes fine:
image

martinRenou:
Oh yes I've had this pain too...

So if we want the CI to pass here, you'd need to take a similar approach as what I did in this PR: https://github.com/jupyterlite/jupyterlite/pull/1715

Basically make a pyodide-kernel PR where you patch jupyterlite's code [like here](https://github.com/jupyterlite/pyodide-kernel/pull/209), that PR should trigger a build of the package on readthedocs, then point to that built package in the jupyterlite repo as I [did here](https://github.com/jupyterlite/jupyterlite/pull/1715/changes/1f829301f9697d9c539c472216231875511426c2).

Yahiewi:
Thanks @martinRenou for the help, I opened this [PR](https://github.com/jupyterlite/pyodide-kernel/pull/338) and modified the pyproject.toml (since we're no longer using requirements.txt as in your example PR) to use the readthedocs build. Can you please approve the workflows?

martinRenou (approved):
That LGTM! Thanks :)

cc. @jtpio if you'd like to have a look

jtpio:
Thanks I'll have a look!

Right the workflow when changes are also required in kernel repositories is not great, hopefully we can find a way to make that easier at some point.

pyproject.toml
```diff
@@ -77,7 +77,7 @@ lint = [
 
 # An example downstream kernel
 pyodide-kernel = [
-    "jupyterlite-pyodide-kernel[lock] >=0.9.0a1; sys_platform != 'emscripten'",
+    "jupyterlite-pyodide-kernel[lock] @ https://jupyterlite-pyodide-kernel--338.org.readthedocs.build/en/338/_static/jupyterlite_pyodide_kernel-0.9.0a1-py3-none-any.whl ; sys_platform != 'emscripten'",
```
jtpio, line 80: We can keep this link as is for now until `jupyterlite-pyodide-kernel` updates to a newer JupyterLite that will include this change, and release 0.9.0a2.

jtpio (approved):
Thanks!

PR #2043, opened 2026-09-07
docs: contents added after the build need contentsAllJsonFile set

### LINKED ISSUE #2041: Adding notebooks to a prebuilt site silently stopped working in 0.8
## Description

In 0.7 a site builder can build JupyterLite then add notebooks freely later, by drop them into `files/` and write `api/contents/all.json` programmatically (as the "Hard Way" page describes), and the file browser can list them correctly. This is how a static site generator with no Python at build time includes notebooks in JupyterLite: build the app once, add each site's notebooks when the site is generated.

In 0.8 this broke silently. The same layout shows an empty file browser, and opening a notebook fails with "Could not find content with path". No request is made, no 404 is logged, and the file browser is silent. Two site generators hit this independently, Chickadee (Swift) in June and moss (Rust) in July.

## Reproduce

1. `pip install jupyterlite-core==0.8.3 && jupyter lite build` in an empty directory
2. Copy a notebook to `_output/files/x.ipynb` and write `_output/api/contents/all.json` listing it
3. Serve `_output` and open `notebooks/?path=x.ipynb`
4. See "Could not find content with path x.ipynb"; the file browser under `tree/` is empty

Repeat with `jupyterlite-core==0.7.x` and the notebook opens.

Diagnosis:
[`_getServerDirectory`](https://github.com/jupyterlite/jupyterlite/blob/v0.8.3/packages/services/src/contents/drive.ts#L1096-L1106) fetches `all.json` only when `jupyter-config-data.contentsAllJsonFile` is set. [`ContentsAddon.post_build`](https://github.com/jupyterlite/jupyterlite/blob/v0.8.3/py/jupyterlite-core/jupyterlite_core/addons/contents.py#L68-L71) sets it, but returns first when `files/` does not exist, so a build without contents never writes the key.

[`migration.md`](https://jupyterlite.readthedocs.io/en/stable/migration.html#contents-and-workspaces-indexing) documents the change and names the key. It does not say that a build with no contents never writes it, so anyone adding contents afterward must set it themselves. The key is absent from [`jupyterlite.schema.v0.json`](https://github.com/jupyterlite/jupyterlite/blob/v0.8.3/app/jupyterlite.schema.v0.json), from [`hard.md`](https://jupyterlite.readthedocs.io/en/stable/howto/configure/advanced/hard.html) (which still says to write `all.json` by hand), and from [`files.md`](https://jupyterlite.readthedocs.io/en/stable/howto/content/files.html) (which still says the 404 on `all.json` "can be safely ignored").

## Expected behavior

Either the 0.7 behaviour, or the new key in the schema and on the three pages, plus a console warning when it is unset.

I am planning to open a PR for the latter two unless instructed otherwise.

## Context

- JupyterLite version: 0.8.3 (0.8.0 onward; 0.7 unaffected)

guoliu:
## References

Fixes [#2041.](https://github.com/jupyterlite/jupyterlite/issues/2041)

## Code changes

Since 0.8, the frontend reads `api/contents/{dir}/all.json` only when `contentsAllJsonFile` is set in `jupyter-lite.json`, and `jupyter lite build` sets it only when `files/` exists at build time. A site that builds the app once and adds its notebooks afterwards now silently breaks; the key is not in the schema and neither how-to that explain writing `all.json` mentions it.

- `app/jupyterlite.schema.v0.json`: add `contentsAllJsonFile` and `workspacesAllJsonFile`. The frontend reads both and the build addons write both, but the schema did not list them.
- `docs/howto/configure/advanced/hard.md`: the hand-written `all.json` recipe now sets the key, without which that recipe has not worked since 0.8.
- `docs/howto/content/files.md`: say the build writes the key, and drop the note about a 404 that "can be safely ignored". 0.8 removed that request, so the 404 no longer occurs.

## User-facing changes

Documentation and schema only. The two keys now validate and autocomplete in `jupyter-lite.json`.

None.

### DIFF
--- a/app/jupyterlite.schema.v0.json
+++ b/app/jupyterlite.schema.v0.json
@@ -110,6 +110,14 @@
           "type": "string",
           "default": "JupyterLite Storage"
         },
+        "contentsAllJsonFile": {
+          "description": "name of the index file, relative to `api/contents/{dir}/`, listing the contents served from `files/`. Written by `jupyter lite build` when `files/` exists; set it (usually to `all.json`) when contents are added after the build, or no index is read",
+          "type": "string"
+        },
+        "workspacesAllJsonFile": {
+          "description": "name of the index file under `api/workspaces/` listing the workspaces served from the site. Written by `jupyter lite build` when workspaces exist; set it when workspaces are added after the build",
+          "type": "string"
+        },
         "workspacesStorageDrivers": {
           "description": "names of the localforage driver for workspaces, or `null` for the best available",
           "$ref": "#/definitions/localforage-driver-set"
--- a/docs/howto/configure/advanced/hard.md
+++ b/docs/howto/configure/advanced/hard.md
@@ -46,6 +46,8 @@ Open a browser:
 ```
 
 - Paste this JSON in `$YOUR_JUPYTERLITE/api/contents/all.json`
+- Set `"contentsAllJsonFile": "all.json"` under `jupyter-config-data` in
+  `$YOUR_JUPYTERLITE/jupyter-lite.json`; without it the index is never read
 - Copy your files in `$YOUR_JUPYTERLITE/files`
 - Repeat this for every subfolder `:(`
 
--- a/docs/howto/content/files.md
+++ b/docs/howto/content/files.md
@@ -20,13 +20,13 @@ Will be:
 - copied to the built site under `{output-dir}/files/`
   - may have timestamps changed if `--source-date-epoch` is provided.
 - indexed to provide `{output-dir}/api/contents/{subdir?}/all.json`
+- named in `jupyter-lite.json` as `contentsAllJsonFile`, which is what makes the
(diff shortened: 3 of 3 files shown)


jtpio (approved):
Thanks, good catch!

This was indeed changed in 0.8.0 as a way to avoid 404 requests when no content was indexed during the build process.

guoliu:
Thanks for the quick response @jtpio!

I maintain a static site generator that ships Jupyter notebooks with JupyterLite; I have users already using this feature, and I intend to further improve the Jupyter notebooks support.

I'd love to add the test and reference page this issue implies, so it becomes easier and more reliable for other static site generators as well; I submitted a Jupyter Foundation proposal to fund that effort: https://github.com/jupyter-governance/funding-proposals/issues/92

If you think that's a good direction to go, a comment on that issue would be greatly appreciated!

PR #2047, opened 2026-09-10
Fix "Save As"

martinRenou:
## References

Workaround https://github.com/jupyterlab/jupyterlab/issues/19723

### DIFF
--- a/packages/services/src/contents/drive.ts
+++ b/packages/services/src/contents/drive.ts
@@ -517,7 +517,10 @@ export class BrowserStorageDrive implements Contents.IDrive {
     if (path === '') {
       const folder = await this._getFolder(path);
       if (folder === null) {
-        throw Error(`Could not find file with path ${path}`);
+        // Hacky but JupyterLab assumes an HTTP error
+        const err = new Error(`Path ${path} does not exist.`);
+        (err as any).response = { status: 404 };
+        throw err;
       }
       return folder;
     }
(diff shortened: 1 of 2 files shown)


packages/services/src/contents/drive.ts
```diff
@@ -517,7 +517,10 @@ export class BrowserStorageDrive implements Contents.IDrive {
     if (path === '') {
       const folder = await this._getFolder(path);
       if (folder === null) {
-        throw Error(`Could not find file with path ${path}`);
+        // Hacky but JupyterLab assumes an HTTP error
```
jtpio: Wondering if we could create a small helper similar to the one used in JupyterLab here:

https://github.com/jupyterlab/jupyterlab/blob/7241239e2a393ff6fc94a5d7078be0aa5b80929c/packages/services/src/testutils.ts#L717-L720

Something like the following:

```ts
function notFoundError(path: string): ServerConnection.ResponseError {
  const response = new Response(null, { status: 404, statusText: 'Not Found' });
  return new ServerConnection.ResponseError(response, `Path ${path} does not exist.`);
}
```

And then throw the error with:

```ts
throw notFoundError(path);
```

monca-dev (approved):

packages/services/src/contents/drive.ts
martinRenou: Nice idea! I'll apply that

jtpio (approved):
Thanks!

PR #2051, opened 2026-09-11
Fix the plugin IDs of `@jupyterlite/apputils-extension`

jtpio:
## References

As noticed in https://github.com/jupyterlab/jupyterlab/pull/19627#pullrequestreview-5178222861

## Code changes

- [x] Fix naming of some plugins
- [x] Mention in the 0.9.0 migration guide

## User-facing changes

Normally none, except if some of these plugins were explicitly referenced for example in `disabledExtensions`.

None


MUFFANUJ (approved):
LGTM!

PR #2056, opened 2026-09-23
Add the extension manager

jtpio:
## References

Reboot https://github.com/jupyterlite/jupyterlite/pull/1987 leveraging the new JupyterLab extension point.

## Code changes

- [x] Enable a read-only extension manager
- [x] Tests

## User-facing changes

image

None


jtpio:
It's probably in a good enough state for now to be included in the next pre-release. It would be interesting to follow up and look into being able to enable and disable plugins and extensions in the browser directly.

